PR #20, opened 2016-10-29
Fix and improve robot failure example notebook.

jkleint:
The robot failure example notebook did not contain the output, rendering it
not so informative an example.

The notebook did not download the data before using it.

It did not use the data imputation settings for `extract_relevant_features()`, leading to a crash.

Added timings for the extraction functions, as it's important to know that the current
implementation is surprisingly time consuming.

Consolidated the example plots and put them first, as that is more meaningful for someone
viewing the example with no background.

Added a bit of narrative text to give readers a basic idea what's going on.


MaxBenChrist:
Thank you for your contribution @jkleint .

Maintaining ipython notebooks in a git repository is always messy. We decided to clear the output of the notebook's cells as it gives you slightly better diffs. 

In general, you can clear the output of all the cells with the following button:
bildschirmfoto 2016-10-30 um 12 37 08

Further, in the notebook I added some words that the filtered extraction is essentially equal to an extraction plus filtering step

PR #27, opened 2016-10-31
I25 fix quickstart documentation

### LINKED ISSUE #25: Fix code on quick start documentation
Reported by user WickedWicky on r/machinelearning

> Following this page http://tsfresh.readthedocs.io/en/latest/text/quick_start.html and I don't know if that is supposed to be up to date but the code blocks don't fit together.
"from tsfresh import select features" should be "from tsfresh import select_features"
and the last block you use 'df' as an argument where you defined it as 'timeseries' earlier.

MaxBenChrist:
fixes #25

### DIFF
--- a/docs/text/quick_start.rst
+++ b/docs/text/quick_start.rst
@@ -99,7 +99,7 @@ We will now remove all ``NaN`` values and select only the relevant features
 
 .. code-block:: python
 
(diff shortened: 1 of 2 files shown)


MaxBenChrist:
closes #25

PR #37, opened 2016-11-04
additional tsfresh example

mynameisvinn:
include another tsfresh example, using the popular uci accelerometer dataset for activity recognition. 

this is a good comparison against traditional time series techniques (eg dtw + knn): https://github.com/markdregan/K-Nearest-Neighbors-with-Dynamic-Time-Warping


MaxBenChrist:
Great contribution @mynameisvinn.

We try to put the notebooks under git without the output. I will clear it.
Further I will add some code to download the dataset

MaxBenChrist:
Sorry @mynameisvinn, I can not merge this PR as you included the data. In the README.txt of the data set it says that 

> This dataset is distributed AS-IS and no responsibility implied or explicit can be addressed to the authors or their institutions for its use or misuse. Any commercial use is prohibited. 

If I merge this PR, we have commits that include the activity data set. That would undermines tsfreshs MIT License which allows a commercial use :-(

MaxBenChrist:
Could you please create a new pull request just containing the cleared notebook without output?

Afterwards we have to add a routine to download the acitivity data set on the that pr. We can use the ` download_robot_execution_failures` method from [robot_execution_failures.py](https://github.com/blue-yonder/tsfresh/blob/d10cdd1398b187f556e7ad98572cef9d8f949123/tsfresh/examples/robot_execution_failures.py).

PR #39, opened 2016-11-06
Standalone script to run on csv

mynameisvinn:
@jneuff you mentioned a standalone script that can be used to run tsfresh directly on csv (and subsequently output results to a csv). 

is this what you had in mind? lmk


scripts/run_tsfresh.py
```diff
@@ -0,0 +1,54 @@
...
+	'''
+	
+	# transpose since tsfresh reads times series data column-wise, not row-wise
+	df_t = df.transpose()
+	
+	# grab first 50 readings to save time
+	master_df = pd.DataFrame(df_t[0])
+	master_df['id'] = 0
+
+	for i in range(1, 50):
```
MaxBenChrist: And what happens if we have more than 50 different time series? ;-)

scripts/run_tsfresh.py
```diff
@@ -0,0 +1,54 @@
...
+saved as features_path_to_your_csv.csv
+'''
+
+import sys
+import numpy as np
+import pandas as pd
+from tsfresh import extract_features, extract_relevant_features, select_features
+
+
+def _preprocess(df):
```
MaxBenChrist: this is essentially a transpose?

In this method you build a stacked data frame. You could take our dictionary time series container format, this would should fit better here

scripts/run_tsfresh.py
```diff
@@ -0,0 +1,54 @@
...
+	'''
+	given a dataframe where records are stored row-wise, rearrange it
```
MaxBenChrist: You could specify the format a little bit more. I think you wrote this method for dfs where every row has one kind of time series

MaxBenChrist:
What is the scripts/data.txt?

You could add your script to the setup.cfg to get a standalone script like this 

```
[entry_points]
# Add here console scripts like:
console_scripts =
    run_my_project = my_project.cli:run
# as well as other entry_points.

```

PR #44, opened 2016-11-07
Implement approximate entropy algorithm

CYHSM:
Information about this algorithm : [Approximate Entropy](https://en.wikipedia.org/wiki/Approximate_entropy) 

Also see [this](https://www.ncbi.nlm.nih.gov/pubmed/10843903?dopt=Abstract) paper for some limtations and similar algorithms which could also be implemented as a new feature in tsfresh.

### DIFF
--- a/tests/feature_extraction/test_feature_calculations.py
+++ b/tests/feature_extraction/test_feature_calculations.py
@@ -429,3 +429,10 @@ class FeatureCalculationTestCase(TestCase):
         self.assertEqualPandasSeriesWrapper(range_count, list(range(10)), 10, min=0, max=10)
         self.assertEqualPandasSeriesWrapper(range_count, list(range(0, -10, -1)), 9, min=-10, max=0)
         self.assertEqualPandasSeriesWrapper(range_count, [np.NaN, np.PINF, np.NINF] + list(range(10)), 10, min=0, max=10)
+
+    def test_approximate_entropy(self):
+        self.assertEqualOnAllArrayTypes(approximate_entropy, [1], 0, m=2, r=0.5)
+        self.assertEqualOnAllArrayTypes(approximate_entropy, [1, 2], 0, m=2, r=0.5)
+        self.assertEqualOnAllArrayTypes(approximate_entropy, [1, 2, 3], 0, m=2, r=0.5)
+        self.assertEqualOnAllArrayTypes(approximate_entropy, [1, 2, 3], 0, m=2, r=0.5)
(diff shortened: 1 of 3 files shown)


MaxBenChrist:
Nice work! Unfortunately, our unit tests are failing. Can you look into that?

Do you want to implement the features from the "Physiological time-series analysis using approximate entropy and sample entropy" paper as well?

MaxBenChrist:
[Here](http://tsfresh.readthedocs.io/en/latest/text/how_to_contribute.html) you can find an instruction how to run the tests locally.

CYHSM:
Ah i see, sorry! I will look into it!

MaxBenChrist:
Perfect! And while you are on it, it would be great if you could also add some unit tests for your new feature calculator.

If you need some help, let us know

CYHSM:
Ok i looked into it. Took me some time to figure this out. (is there an easier way to debug pytest instead of using stdout?) 

So maybe for others it should be made more clear that some tests pass sliced panda.Series to the functions so the keys do not always start at 0 (this can be resolved with `pd_series.reset_index(drop=True)` or `x = np.asarray(x)`)

Also i'm curious, why not make numpy array the standard input?

MaxBenChrist:
Hi @CYHSM great addition. 👍  I will merge this pr during the next days. Before merging, I will think about this _phi function. Maybe I can find a way to vectorize it

CYHSM:
I tried vectorizing the rest. It works with: 
```
x_re = np.array(np.squeeze(np.split(np.tile(x, [1, N-m+1]),
                         np.insert(np.arange(m, (N+1)*(N-m+1)+m, step=N+1),
                         np.arange(N-m+1),
                         np.arange(0, (N+1)*(N-m+1), step=N+1)), axis=1)[1::2]))
```
but its slower in the end because you have to create a NxN matrix or an array with size N^2.  

I removed one of the loops and replaced it with indexing: 
`np.array([x[i:i+m] for i in range(N - m + 1)])`
This gives another speed boost!

MaxBenChrist:
Awesome, the first feature calculator that was contributed.

You want to solve that merge conflict? ( you just have to move your feature calculator settings to the init method of the FeatureExtractionSettings Object)

CYHSM:
Yay. Finally (:

MaxBenChrist:
I think you reversed the last commit on master bf9e830

MaxBenChrist:
So close! 😂 I can fix this later

CYHSM:
Oh sorry. What should i do? :grimacing:

MaxBenChrist:
fixed it in 8048ef4

PR #57, opened 2016-11-13
More informative docstring

earthgecko:
To highlight key points of the script and handle saving the output file to any user specified path.


MaxBenChrist:
Nice!

PR #64, opened 2016-11-16
run_tsfresh.py - remove sample first 50 only

earthgecko:
This removes the only sample the first 50 logic and adds the required iloc referenced in https://github.com/blue-yonder/tsfresh/pull/63

### DIFF
--- a/scripts/run_tsfresh.py
+++ b/scripts/run_tsfresh.py
@@ -3,7 +3,6 @@ Run the script with:
 ```
 python run_tsfresh.py path_to_your_csv.csv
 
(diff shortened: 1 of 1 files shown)


MaxBenChrist:
👍

MaxBenChrist:
@earthgecko 

the
 
``` python
df_range = (len(list(df)) - 1)
for i in range(0, df_range):
```
will not extract the last element (range(n) includes the numbers from 0 to n-1)
Further, it can be easier expressed as
``` python
for i in range(0, len(df)):
```

Issue #21, opened 2016-10-30
Add MinimalFeatureExtractionSettings


MaxBenChrist:
Right now our FeatureExtractionSettings contains many features.

It would be nice to have a smaller FeatureExtractionSettings object that only calculates basic properties such as mean, median, min, max. This would allow to fastly  tinker with data and then later calculate the big list of features

nils-braun:
Do you think it would be better to delete the other calculations from the beginning in the name_to_param array or just do not add them to the kind_to_calculation_settings_mapping in the set_default_parameters function?

And how to deal with it in the from_columns method? I think here the functions should still be added if the user requests them - this is why I would not delete them from name_to_param but from kind_to_calculation_settings_mapping.

MaxBenChrist:
@jneuff looked into that. Maybe he can add some remarks as well. We moved the setup of `name_to_param` from the `set_default_parameters` to the constructor. See 8048ef4.

The current logic is *"extract all the features from `name_to_param` for all types of time series"*. 

If we follow that for a minimal MinimalFeatureExtractionSettings Object, then it just have a smaller  `name_to_param` object.

jneuff:
I'd offer methods to instantiate a minimal and a full Settings object. A simple solution would be to initialize `name_to_param` as empty dict in the __init__ and provide two class methods that will instantiate and configure a minimal and a full Settings object respectively. Similar to what we're doing right now with from_columns.

nils-braun:
I am actually against doing this with another member function - because it would be something everyone has to call everytime and it would be easily forgotten. I would rather create a child class MinimalFeatureExtractionSettings and overload the constructor.

PR #70, opened 2016-11-20
Minimal feature extractor settings

nils-braun:
This is a first draft for #21 open for discussion.

### DIFF
--- a/tests/feature_extraction/test_settings.py
+++ b/tests/feature_extraction/test_settings.py
@@ -6,14 +6,15 @@ from __future__ import absolute_import, division
 
 from unittest import TestCase
 import numpy as np
-from tsfresh.feature_extraction.settings import FeatureExtractionSettings
+import pandas as pd
+from tsfresh.feature_extraction.extraction import extract_features
+from tsfresh.feature_extraction.settings import FeatureExtractionSettings, MinimalFeatureExtractionSettings
 import six
 from tsfresh.feature_extraction import feature_calculators
 
+
 class TestSettingsObject(TestCase):
-
     def test_from_columns(self):
-
         tsn = "TEST_TIME_SERIES"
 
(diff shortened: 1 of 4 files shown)


MaxBenChrist:
I really like the usage of the @set_property decorator to mark feature calculator as basic.

We could also use this property to mark a feature as computational expensive. (like the cwt and fourier stuff)

MaxBenChrist:
I vote for merging this in

MaxBenChrist:
We could probably reference this MinimalFeatureExtractionSettings object somewhere in the documentation

jneuff:
Really good idea setting the properties on the calculators. We could come up with a threshold to denote all of them with either expensive or cheap. However, I'd say 'minimal' is not the best choice for a property denoting cheap calculators.

We should come up with the threshold (n milliseconds on some reference timeseries on average hardware) and a naming scheme. The question is, do we want to merge it in now as it is and change later or merge it after we have come up with a criteria and naming scheme?

As the minimal FeatureSettings object is quite useful, I'm for a merge now. Breaking the property naming scheme later should be OK, maybe we can add a comment, that the naming is due to change.

MaxBenChrist:
I would merge this now and open an issue for the naming scheme. Or do you already have something in mind to replace "minimal" ? Maybe "complexity_low"?

nils-braun:
I also do not like the name and your proposal @jneuff to have a continuous number (in this case the runtime on a standardized machine) is very good. The problem is only, that we have to define this standardized machine and this  will be hard (because not all our contributors have access to the same machine). Another approach may be to have tree levels of complexity without the need to "measure".

This would then be probably another pull-request, but if you have a better name immideately, I can edit this PR now.

MaxBenChrist:
I think it is fine as it is. I merge it and open a new issue regarding the naming

PR #76, opened 2016-11-24
Parallel p values

nils-braun:
Related to #35

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,8 @@ install:
 - pip install .
 - pip install -r requirements.txt
 - pip install -r test-requirements.txt
+env:
+- NUMBER_OF_CPUS=1
 script:
 - python setup.py test
 deploy:
--- a/tests/feature_selection/test_feature_selectors.py
+++ b/tests/feature_selection/test_feature_selectors.py
@@ -184,6 +184,17 @@ class FeatureSelection(TestCase):
         self.assertLess(result_series.p_value, self.maximal_p_value_for_significant_features)
         self.assertEqual(result_series.type, "real")
 
+    def test_const_feature(self):
+        """Constant features should not be used in the p-value calculation."""
+        x = pd.Series([1.111]*250, name="TEST")
+        y = x + pd.Series(np.random.normal(0, 1, 250))
+
+        result_series = tsfresh.feature_selection.feature_selector._calculate_p_value("TEST", pd.DataFrame(x), y,
+                                                                                      self.settings, False)
+        self.assertEqual(result_series.name, "TEST")
+        self.assertEqual(result_series.type, "const")
+        self.assertEqual(result_series.rejected, False)
+
 
 # noinspection PyUnresolvedReferences
 class FeatureSelectionConfigTestCase(TestCase):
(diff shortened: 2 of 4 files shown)


MaxBenChrist:
hey nils, have you seen the changes I made in 7a21b5e05fff2917aad8d7bcd753eff976b1c6df ? 

I think they are not yet incorporated in this pr?

nils-braun:
They are, I have merged master into the PR and your new unittest changes are also in (and are still running :-) ). However, it seems I should include some more unittests to increase the coverage again.

MaxBenChrist:
We can use our existing unit tests and let them point to `_calculate_p_value` or not?

MaxBenChrist:
I am also working on increasing the coverage, see for example #79

nils-braun:
Actually I have seen that coverage is a bit wrong here, as I do not decrease the coverage, but because I had to wrap the function _calcuate_p_values into a partial function, coverage does not seems to recognize the unittests actually call this function (at least twice). How should we fix this?

MaxBenChrist:
cant we let the unit test target the _calcuate_p_values function directly?

MaxBenChrist:
perfect, can I merge?

nils-braun:
I do think so :-)

MaxBenChrist:
what do you think of taking half the cores as number of processes?

MaxBenChrist:
or all

nils-braun:
Do you think we should make multiprocessing the default or better give hints in cases the calculation is too slow in the documentation? Using more than one core by default is - lets say - maybe a bit unexpected?

nils-braun:
But I may be a bit too pessimistic - I think going the way you said (setting it to number of cores by default) is good. I was just thinking on batch-systems, where you may be not allowed to use more than one core... But I think this is not our main application field :-)

MaxBenChrist:
You already set the default to 2 jobs in the feature selection settings, don't you?

I think number of cores/2 should be fine :)

MaxBenChrist:
or all :D

nils-braun:
Yes, my comment was more of a general reason :-) But I think just using all CPUs is fine ;-)

MaxBenChrist:
travis stalled again. 

Just a wild guess, but maybe cpu_count() returns the number of physical cpus, even though the travis vm has only access to 1-2?

MaxBenChrist:
after you solved that cpu issue and merged it with the current master, we can have a versioning jump to 0.1.3 :)

nils-braun:
Perfect, I will solve this when my network connection is a bit more stable again and I can access travis a bit faster...

nils-braun:
I have now added an override shell environment variable which is set to 1 for travis. I think this should work well.

MaxBenChrist:
Yes, we could also directly use the TRAVIS enviroment variable. 

see https://docs.travis-ci.com/user/environment-variables/

nils-braun:
Yes, this is possible, but I think is is more "general" to have a setting for the number of used CPUs. For example, when the number of each used CPU costs (e.g. on batch farms), you probably want to have a setting for this...

Issue #11, opened 2016-10-27
Improve unit tests with data sets


jneuff:
Currently our test cases only use integer data sets. We should add tests using data sets containing floats, NaNs and +/-Infs as well as data sets of length 1.

mynameisvinn:
just submitted PR

nils-braun:
We do have many tests no which also contain non-integer values.

PR #77, opened 2016-11-24
include a more diverse set of datasets for unit testing

mynameisvinn:
for https://github.com/blue-yonder/tsfresh/issues/11

@jneuff ive included a few other datasets (w nans, floats, single value, etc.) for unit tests wrt feature calculation. for now, im using sum_values as an example - is this what you had in mind?


MaxBenChrist:
i close this pr because it is already contained in #78

Issue #5, opened 2016-10-26
Add new feature: p-value of the Augmented Dickey-Fuller test


jneuff:
Currently we calculate the Augmented Dickey-Fuller test statistic as [feature](https://github.com/blue-yonder/tsfresh/blob/master/tsfresh/feature_extraction/feature_calculators.py#L253). The p-value of the Augmented Dickey-Fuller test might also be a relevant feature.

As the statsmodels function `adfuller` returns the test statistic as well as the p-value, it would make sense to memoize or buffer the results and add the new feature as a separate function.

A tutorial on how to add new features to tsfresh can be found in our documentation: http://tsfresh.readthedocs.io/en/latest/text/how_to_add_custom_feature.html

PR #80, opened 2016-11-26
include new feature calculator for dickey fuller p-value

mynameisvinn:
@jneuff in reference to https://github.com/blue-yonder/tsfresh/issues/5

returning p-value from dickey fuller calculation

### DIFF
--- a/tsfresh/feature_extraction/feature_calculators.py
+++ b/tsfresh/feature_extraction/feature_calculators.py
@@ -276,6 +276,27 @@ def augmented_dickey_fuller(x):
 
 @set_property("fctype", "aggregate")
 @not_apply_to_raw_numbers
+def augmented_dickey_fuller_p_val(x):
+    """
+    The Augmented Dickey-Fuller is a hypothesis test which checks whether a unit root is present in a time
+    series sample. This feature calculator returns the value of the corresponding p-value.
+
+    :param x: the time series to calculate the feature of
+    :type x: pandas.Series
+    :return: the value of this feature
+    :return type: float
+    """
+    maxlag = 1
+    try:
+        return adfuller(x, maxlag)[1]
+    except LinAlgError:
+        return np.NaN
+    except ValueError:  # occurs if sample size is too small
+        return np.NaN
+
+
+@set_property("fctype", "aggregate")
+@not_apply_to_raw_numbers
 def abs_energy(x):
     """
     Returns the absolute energy of the time series which is the sum over the squared values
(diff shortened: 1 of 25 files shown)


MaxBenChrist:
Nice, but we should wrap the call to the ad fuller test to not calculate I several times. We can use a decorator for that. See http://stackoverflow.com/questions/815110/is-there-a-decorator-to-simply-cache-function-return-values

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -276,6 +276,27 @@ def augmented_dickey_fuller(x):
...
+    """
+    The Augmented Dickey-Fuller is a hypothesis test which checks whether a unit root is present in a time
+    series sample. This feature calculator returns the value of the corresponding p-value.
+
+    :param x: the time series to calculate the feature of
+    :type x: pandas.Series
+    :return: the value of this feature
+    :return type: float
+    """
+    maxlag = 1
```
MaxBenChrist: Line 290 to 295 should be moved into its own method and wrapped with a cache decorator

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -251,6 +251,26 @@ def mean_autocorrelation(x):
         return np.nanmean(r / var)
 
 
+def memoize(f):
+    cache = {}
```
MaxBenChrist: I dont like that the cache is not size limited. have you looked at some caching libraries?

https://pythonhosted.org/cachetools/

mynameisvinn: makes sense. i have not but will take a look.

mynameisvinn:
in all sorts of pain right now - cant figure out why caching decorator is failing tests... will review next weekend

mynameisvinn:
using decorator to cache results from augmented dickey fuller calculation for (1) test statistic and (2) p-value. i think this is what u had in mind but not 100% certain...

latest pr is failing all sorts of tests but not anything related to dickey fuller calculations.

MaxBenChrist:
the failing unit tests are related to the caching mechanism. do you want to look into that?

MaxBenChrist:
@mynameisvinn are you still interested in finishing this or should we take over?

mynameisvinn:
@MaxBenChrist sorry for slow response - busy year end work stuff. i wont have much free time until the new year, so go ahead without me!

see u in the new year!

MaxBenChrist:
thanks for the response, see you!

PR #87, opened 2016-12-01
Provide p-values of selected features

kempa-liehr:
* FeatureSelection().relevant_features is a DataFrame which contains the p-values of the selected features. 
* FeatureSelection().relevant_features.index holds the column names of the selected features.

### DIFF
--- a/tsfresh/transformers/feature_selector.py
+++ b/tsfresh/transformers/feature_selector.py
@@ -39,7 +39,7 @@ class FeatureSelector(BaseEstimator, TransformerMixin):
     >>> []
 
     The same holds for the feature importance:
-    >>> selector.feature_importance_
+    >>> selector.feature_importances_
     >>> array([], dtype=float64)
 
     The estimator keeps track on those features, that were relevant in the training step. If you
@@ -91,7 +91,7 @@ class FeatureSelector(BaseEstimator, TransformerMixin):
         :type chunksize: int
         """
         self.relevant_features = None
-        self.feature_importance_ = None
+        self.feature_importances_ = None
         self.p_values = None
 
         self.test_for_binary_target_binary_feature = test_for_binary_target_binary_feature
@@ -132,7 +132,7 @@ class FeatureSelector(BaseEstimator, TransformerMixin):
                                 self.fdr_level, self.hypotheses_independent,
                                 self.test_for_binary_target_real_feature)
         self.relevant_features = df_bh.loc[df_bh.rejected].Feature.tolist()
-        self.feature_importance_ = 1.0 - df_bh.p_value.values
+        self.feature_importances_ = 1.0 - df_bh.p_value.values
         self.p_values = df_bh.p_value.values
 
         return self
--- a/tsfresh/transformers/feature_selector.py
+++ b/tsfresh/transformers/feature_selector.py
@@ -116,9 +116,10 @@ class FeatureSelector(BaseEstimator, TransformerMixin):
         if not isinstance(y, pd.Series):
             y = pd.Series(y.copy())
 
-        df_bh = check_fs_sig_bh(X, y, self.n_processes, self.chunksize, self.fdr_level, self.hypotheses_independent,
+        df_bh = check_fs_sig_bh(X, y, self.n_processes, self.chunksize,
+                                self.fdr_level, self.hypotheses_independent,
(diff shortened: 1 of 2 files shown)


.coveragerc
```diff
@@ -1,31 +0,0 @@
-# .coveragerc to control coverage.py
```
MaxBenChrist: why are you deleting the .coveragerc?

MaxBenChrist:
this is just a change of the internal api right? what is the motivation behind the change?

kempa-liehr:
In this particular case, I had applied the FeatureSelection transformer to my feature set, because I wanted to check the individual p-values of the selected features. This will be important for teaching purposes.

And yes, it’s just a slight change to the internal API, which simultaneously provides an external interface for retrieving the p-values of the selected features.

kempa-liehr:
.coveragerc is generated automatically by running 

python setup.py test

And because there are no modifications to it, it’s redundant, right?

MaxBenChrist:
On my system it does not get generated after calling `python setup.py test`

kempa-liehr:
So, there is .coveragerc again.

MaxBenChrist:
I think it would be a nice idea to save the df_bh DataFrame in the object, similar to `feature_importances` in for example sklearns random forest

kempa-liehr:
Good point.

MaxBenChrist:
You are right!

jneuff:
This is a very good idea @kempa-liehr ! Maybe we could have an option to return the p-values or feature importances from the select_features function.

tsfresh/transformers/feature_selector.py
```diff
@@ -82,7 +84,7 @@ def fit(self, X, y):
             y = pd.Series(y.copy())
 
         df_bh = check_fs_sig_bh(X, y, self.settings)
-        self.relevant_features = df_bh.loc[df_bh.rejected].Feature
+        self.relevant_features = df_bh.loc[df_bh.rejected].p_value
```
MaxBenChrist: For me, that is a public API change. (It will break some of my own scripts...)

Lets undo this changes and add the line 

`self.feature_importances_  = 1- df_bh.loc[df_bh.rejected].p_value`

MaxBenChrist: Also `relevant_features` sound like a name of a feature and not a numerical measurement of their importance

kempa-liehr: Okay, I'm going to put `self.feature_importance_` in.

kempa-liehr: But the change of the API does make sense, because `self.relevant_features` contains the relevant features both in its index and its values.

We could document this change for version 0.5 thoroughly and maybe even think about combining it with a diagram explaining the Benjamini-Yekutieli (cf. Fig 2 in ACML paper).

>

MaxBenChrist: Okay so to make it clear lets have three atrributes:

- feature_importance_ = 1-p_values, no index 
- p_values =   p_value, no index 
- relevant_features = list of relevant feature names, no index
- features = list of feature names, no index

can you change the target of this pr to the feature_extraction_settings_refactoring branch?

kempa-liehr:
Is there a reason for dismissing the relation between feature names and p_values/significance?

tsfresh/transformers/feature_selector.py
kempa-liehr: Is there a reason for dismissing the relation between feature names and p_values/significance?

MaxBenChrist: I guess it is more similar to the attributes of sklearn transformers. E.g. http://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomForestRegressor.html, the feture_importance_ is just an array with the feature importances

MaxBenChrist:
I would merge this pr. However, you are also sneaking in changes to the dritbif files. :D 

Can you undo them, so this pr will only tackle the following files:

tests/transformers/test_feature_selector.py
tsfresh/feature_extraction/feature_calculators.py
tsfresh/transformers/feature_selector.py

Thanks

kempa-liehr:
I also skipped the changes on feature_calculators.py, because they were only minor wording corrections not related to this PR.

PR #98, opened 2016-12-09
Improve parallelization

jneuff:
The PR is not ready for a merge. I opened it to facilitate the discussion of the introduced changes. Please comment!

Here's what I did: I extracted the parallel calculation of the features and added two functions for this purpose. One works just as before, parallelising the calculation per time series kind. The other additionally parallelizes per sample.
As the parallelization per sample introduces significant overhead on large data sets with several time series per sample, we should offer both options. Therefore I introduced a new optional argument to the extract_features function. A rule-of-thumb is used to determine the default strategy.

TODO:
- ~~Ensure test coverage for both parallel extract functions~~
- ~~Add documentation discussing the tradeoffs involved in the two options~~
- ~~Add test cases asserting the resulting feature matrix on given input is identical for both options~~
- Optional: Add test case to validate rule-of-thumb
- ~~solve merge conflict~~

### DIFF
--- a/docs/text/parallelization.rst
+++ b/docs/text/parallelization.rst
@@ -37,8 +37,8 @@ Additionally there are two options for how the parallelization is done:
 To enforce an option, either pass ``'per_kind'`` or ``'per_sample'`` as the ``parallelization=`` parameter of the
 :func:`tsfresh.extract_features` function. By default the option is chosen with a rule of thumb:
 
-If the number of different time series (kinds) is less than half the available worker processes (``n_processes``)
-then ``'per_sample'`` is chosen, otherwise ``'per_kind'``.
+If the number of different time series (kinds) is less than half of the number of available worker
+processes (``n_processes``) then ``'per_sample'`` is chosen, otherwise ``'per_kind'``.
 
 Generally, there is no perfect setting for all cases. On the one hand more parallelization can speed up the calculation
 as the work is better distributed among the computers resources. On the other hand parallelization
--- a/docs/text/parallelization.rst
+++ b/docs/text/parallelization.rst
@@ -31,13 +31,13 @@ for the feature selection.
 
 Additionally there are two options for how the parallelization is done:
 
-1.  ``'per_kind'`` parallelizes the feature calculation per kind of timeseries.
+1.  ``'per_kind'`` parallelizes the feature calculation per kind of time series.
 2.  ``'per_sample'`` parallelizes per kind and per sample.
 
 To enforce an option, either pass ``'per_kind'`` or ``'per_sample'`` as the ``parallelization=`` parameter of the
 :func:`tsfresh.extract_features` function. By default the option is chosen with a rule of thumb:
 
-If the number of different timeseries (kinds) is less than half the available worker processes (``n_processes``)
+If the number of different time series (kinds) is less than half the available worker processes (``n_processes``)
 then ``'per_sample'`` is chosen, otherwise ``'per_kind'``.
 
 Generally, there is no perfect setting for all cases. On the one hand more parallelization can speed up the calculation
@@ -50,7 +50,7 @@ Implementing the parallelization we observed the following points:
 -   For small data sets the difference between parallelization per kind or per sample should be negligible.
 -   For data sets with one kind of time series parallelization per sample results in a decent speed up that grows
     with the number of samples.
--   The more kinds of timeseries the data set contains, the more samples are necessary to make parallelization
+-   The more kinds of time series the data set contains, the more samples are necessary to make parallelization
     per sample worthwhile.
 -   If the data set contains more kinds of time series than available cpu cores, parallelization per kind is
     the way to go.
--- a/docs/text/parallelization.rst
+++ b/docs/text/parallelization.rst
@@ -1,0 +1,56 @@
+.. _parallelization-label:
+
+Parallelization
+===============
+
+The feature extraction as well as the feature selection offer the possibility of parallelization.
+Out of the box both tasks are parallelized by tsfresh. However, the overhead introduced with the
+parallelization should not be underestimated. Here we discuss the different settings to control
+the parallelization. To achieve best results for your use-case you should experiment with the parameters.
+
+Parallelization of Feature Selection
+------------------------------------
+
+We use a :class:`multiprocessing.Pool` to parallelize the calculation of the p-values for each feature. On
+instantiation we set the Pool's number of worker processes to
+:attr:`tsfresh.feature_selection.FeatureSignificanceTestsSettings.n_processes`. This field defaults to
+the number of processors on the current system. We recommend setting it to the maximum number of available (and
+otherwise idle) processors.
+
+The chunksize of the Pool's map function is another important parameter to consider. It can be set via the
+:attr:`tsfresh.feature_selection.FeatureSignificanceTestsSettings.chunksize` field. By default it is up to
+:class:`multiprocessing.Pool` to decide on the chunksize.
+
+Parallelization of Feature Extraction
+-------------------------------------
+
+For the feature extraction tsfresh exposes the parameters
+:attr:`tsfresh.feature_extraction.FeatureExtractionSettings.n_processes` and
+:attr:`tsfresh.feature_extraction.FeatureExtractionSettings.chunksize`. Both behave anlogue to the parameters
+for the feature selection.
+
+Additionally there are two options for how the parallelization is done:
+
+1.  ``'per_kind'`` parallelizes the feature calculation per kind of timeseries.
+2.  ``'per_sample'`` parallelizes per kind and per sample.
+
+To enforce an option, either pass ``'per_kind'`` or ``'per_sample'`` as the ``parallelization=`` parameter of the
+:func:`tsfresh.extract_features` function. By default the option is chosen with a rule of thumb:
+
+If the number of different timeseries (kinds) is less than half the available worker processes (``n_processes``)
+then ``'per_sample'`` is chosen, otherwise ``'per_kind'``.
+
+Generally, there is no perfect setting for all cases. On the one hand more parallelization can speed up the calculation
+as the work is better distributed among the computers resources. On the other hand parallelization
+introduces overheads such as copying data to the worker processes, splitting the data to enable the distribution and
+combining the results.
+
+Implementing the parallelization we observed the following points:
+
+-   For small data sets the difference between parallelization per kind or per sample should be negligible.
+-   For data sets with one kind of time series parallelization per sample results in a decent speed up that grows
+    with the number of samples.
+-   The more kinds of timeseries the data set contains, the more samples are necessary to make parallelization
+    per sample worthwhile.
+-   If the data set contains more kinds of time series than available cpu cores, parallelization per kind is
+    the way to go.
--- a/docs/text/parallelization.rst
+++ b/docs/text/parallelization.rst
@@ -8,6 +8,9 @@ Out of the box both tasks are parallelized by tsfresh. However, the overhead int
 parallelization should not be underestimated. Here we discuss the different settings to control
 the parallelization. To achieve best results for your use-case you should experiment with the parameters.
 
+Please let us know about your results tuning the below mentioned parameters! It will help improve this document as
+well as the default settings.
+
 Parallelization of Feature Selection
 ------------------------------------
 
--- a/tsfresh/feature_extraction/extraction.py
+++ b/tsfresh/feature_extraction/extraction.py
@@ -124,9 +124,8 @@ def _extract_features_parallel_per_kind(kind_to_df_map, settings, column_id, col
     """
     Parallelize the feature extraction per kind.
 
-    :param timeseries_container: The pandas.DataFrame with the time series to compute the features for, or a
-               dictionary of pandas.DataFrames.
-    :type timeseries_container: pandas.DataFrame or dict
+    :param kind_to_df_map: The time series to compute the features for in our internal format
+    :type kind_to_df_map: dict of pandas.DataFrame
 
     :param column_id: The name of the id column to group by.
     :type column_id: str
--- a/tsfresh/feature_extraction/extraction.py
+++ b/tsfresh/feature_extraction/extraction.py
@@ -198,7 +198,7 @@ def _extract_features_parallel_per_sample(kind_to_df_map, settings, column_id, c
         results_fifo.put(
             pool.map_async(
                 partial_extract_features_for_one_time_series,
-                [(kind, df_group) for _, df_group in df_grouped_by_id],
+                [(kind, df_group.astype(np.float64)) for _, df_group in df_grouped_by_id],
                 chunksize=settings.chunksize
             )
         )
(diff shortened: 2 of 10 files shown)


MaxBenChrist:
We investigated the performance of tsfresh under different parallelizations and data sets. One has to differentiate two cases:

1.) one has different kind of time series with low to medium number of samples
2.) one has only one to two kind of time series with medium to high number of samples

for 2.) the parallelization by id sample chunks seems to be faster (new). For 1.) the already implemented parallelization by kind seem to be faster (already implemented) and the additional parallelization by sample id is introducing so much overhead that it takes longer.

tsfresh/feature_extraction/extraction.py
```diff
@@ -103,9 +114,103 @@ def extract_features(timeseries_container, feature_extraction_settings=None,
...
+    :param column_value: The name for the column keeping the value itself.
+    :type column_value: str
+
+    :param settings: settings object that controls which features are calculated
+    :type settings: tsfresh.feature_extraction.settings.FeatureExtractionSettings
+
+    :return: The (maybe imputed) DataFrame containing extracted features.
+    :rtype: pandas.DataFrame
+    """
+    all_possible_unique_id_values = set(id_value for kind, df in kind_to_df_map.items()
```
MaxBenChrist: it is unfortunate that we have to iterate through all kinds here

MaxBenChrist: maybe something like

```
lst =  [set(df[column_id]) for df in kind_to_df_map.items()]
all_possible_unique_id_values = set.union(*lst)
```

ist faster?

tsfresh/feature_extraction/extraction.py
```diff
@@ -103,9 +114,103 @@ def extract_features(timeseries_container, feature_extraction_settings=None,
...
+    :param column_value: The name for the column keeping the value itself.
+    :type column_value: str
+
+    :param settings: settings object that controls which features are calculated
+    :type settings: tsfresh.feature_extraction.settings.FeatureExtractionSettings
 
+    :return: The (maybe imputed) DataFrame containing extracted features.
+    :rtype: pandas.DataFrame
+    """
+    all_possible_unique_id_values = set(id_value for kind, df in kind_to_df_map.items()
```
MaxBenChrist: same as above. I think the second for loop with `for id_value in df[column_id]` is not as fast as it could be.

tsfresh/feature_extraction/extraction.py
```diff
@@ -48,22 +52,27 @@ def extract_features(timeseries_container, feature_extraction_settings=None,
...
+
     :param column_kind: The name of the column keeping record on the kind of the value.
     :type column_kind: str
+
     :param column_value: The name for the column keeping the value itself.
     :type column_value: str
 
-    :param feature_extraction_settings: settings object that controls which features are calculated
-    :type feature_extraction_settings: tsfresh.feature_extraction.settings.FeatureExtractionSettings
+    :param parallelization: Either 'per_sample' or 'per_kind'
```
nils-braun: Here you could maybe add a bit more documentation or refer to the other doc.

tsfresh/feature_extraction/extraction.py
```diff
@@ -103,9 +114,103 @@ def extract_features(timeseries_container, feature_extraction_settings=None,
...
+                                                           column_value=column_value,
+                                                           settings=settings)
+    pool = Pool(settings.n_processes)
+
+    # Submit map jobs per kind per sample
+    results_fifo = Queue()
+    for kind, df_kind in kind_to_df_map.items():
+        gb = df_kind.groupby(column_id)
+        results_fifo.put(
+            pool.map_async(
```
MaxBenChrist: why do we use a Queue here and in the `_extract_features_parallel_per_sample` function not?

tsfresh/feature_selection/feature_selector.py
```diff
@@ -115,8 +115,9 @@ def check_fs_sig_bh(X, y, settings=None):
     pool = Pool(settings.n_processes)
 
     # Helper function which wrapps the _calculate_p_value with many arguments already set
-    f = partial(_calculate_p_value, X=X, y=y, settings=settings, target_is_binary=target_is_binary)
-    p_values_of_features = pd.DataFrame(pool.map(f, df_features['Feature']))
+    f = partial(_calculate_p_value, y=y, settings=settings, target_is_binary=target_is_binary)
+    results = pool.map(f, [X[column] for column in set(X.columns)], chunksize=settings.chunksize)
```
nils-braun: Is there a reason, why you have changes df_feature["Features"] to set(X.columns)? I know that both are the same, but I think df_features["Features"] is a bit more readable.

tsfresh/feature_extraction/extraction.py
```diff
@@ -103,9 +114,103 @@ def extract_features(timeseries_container, feature_extraction_settings=None,
...
                                 sorting=feature_extraction_settings.PROFILING_SORTING)
 
+    return result
+
+
+def _extract_features_parallel_per_kind(kind_to_df_map, settings, column_id, column_value):
+    """
+    Parallelize the feature extraction per kind.
+
+    :param timeseries_container: The pandas.DataFrame with the time series to compute the features for, or a
```
nils-braun: kind_to_df_map or timeseries container?

tsfresh/feature_extraction/extraction.py
```diff
@@ -103,9 +114,103 @@ def extract_features(timeseries_container, feature_extraction_settings=None,
...
+
     pool.join()
+    return result
+
+
+def _extract_features_parallel_per_sample(kind_to_df_map, settings, column_id, column_value):
+    """
+    Parallelize the feature extraction per kind and per sample.
+
+    :param kind_to_df_map: The time series to compute the features for in our internal format
```
nils-braun: Above, you have another comment to this parameter.

tsfresh/feature_extraction/extraction.py
```diff
@@ -103,9 +114,103 @@ def extract_features(timeseries_container, feature_extraction_settings=None,
...
+            pool.map_async(
+                partial_extract_features_for_one_time_series,
+                [(kind, group) for _, group in gb],
+                chunksize=settings.chunksize
+            )
+        )
+
+    pool.close()
+
+    # Wait for the jobs to complete and conatenate the partial results
```
nils-braun: typo :-)

tsfresh/feature_extraction/extraction.py
```diff
@@ -103,9 +114,103 @@ def extract_features(timeseries_container, feature_extraction_settings=None,
...
+    """
+    all_possible_unique_id_values = set(id_value for kind, df in kind_to_df_map.items()
+                                        for id_value in df[column_id])
+    df_with_ids = pd.DataFrame(index=all_possible_unique_id_values)
+
+    partial_extract_features_for_one_time_series = partial(_extract_features_for_one_time_series,
+                                                           column_id=column_id,
+                                                           column_value=column_value,
+                                                           settings=settings)
+    pool = Pool(settings.n_processes)
```
nils-braun: you have quite some code duplication here. Actually I do not know if it is good to have two separated functions at all....

tsfresh/feature_extraction/extraction.py
```diff
@@ -103,9 +114,103 @@ def extract_features(timeseries_container, feature_extraction_settings=None,
...
+    for kind, df_kind in kind_to_df_map.items():
+        gb = df_kind.groupby(column_id)
+        results_fifo.put(
+            pool.map_async(
+                partial_extract_features_for_one_time_series,
+                [(kind, group) for _, group in gb],
+                chunksize=settings.chunksize
+            )
+        )
+
```
nils-braun: ok, I have understood your code and it is correct - but I think it is a but unreadable. Some questions:
1. Why are you using a Queue and an async map here? Is it to have "something" in case things break?
2. Maybe you can extract the code in `[(kind, group) for _, group in gb], into something more readable.

docs/text/parallelization.rst
```diff
@@ -0,0 +1,56 @@
...
+
+Implementing the parallelization we observed the following points:
+
+-   For small data sets the difference between parallelization per kind or per sample should be negligible.
+-   For data sets with one kind of time series parallelization per sample results in a decent speed up that grows
+    with the number of samples.
+-   The more kinds of timeseries the data set contains, the more samples are necessary to make parallelization
+    per sample worthwhile.
+-   If the data set contains more kinds of time series than available cpu cores, parallelization per kind is
+    the way to go.
```
jneuff: Yes, good idea.

tsfresh/feature_extraction/extraction.py
```diff
@@ -76,23 +85,25 @@ def extract_features(timeseries_container, feature_extraction_settings=None,
         for key in kind_to_df_map:
             feature_extraction_settings.set_default_parameters(key)
 
+    # Choose the parallelization according to a rule-of-thumb
+    if parallelization is None and (feature_extraction_settings.n_processes / 2) > len(kind_to_df_map):
```
jneuff: The idea behind the rule is exactly this: If I have more cores it might be worth to parallelize by sample. But only if the cores wouldn't be busy anyway by parallelizing by kind.
A better rule should also incorporate the number of samples and the length of the time series.

The rule as it is now maps with my experiments while developing this branch. However, we could also just use parallelization per kind as default and leave it up to the user to enable paralleliziation per sample.

Logging the resulting parallelization is a good idea. I'll add it.

tsfresh/feature_extraction/extraction.py
jneuff: I left this part untouched and just copied it from the pevious function. From what I observed, there is not much speed to gain here. We should settle on the most readable solution.

Besides, I'm not entirely sure we actually need this part. It is there to ensure that the feature matrix contains entries for all samples. Shouldn't the feature calculators return at least NaN in any case?

tsfresh/feature_extraction/extraction.py
jneuff: The queue here serves a very distinct purpose. See my reply to Nils' comment below.

tsfresh/feature_extraction/extraction.py
jneuff: Yes, I'll refer to the respective functions.

tsfresh/feature_extraction/extraction.py
jneuff: I'll fix it.

tsfresh/feature_extraction/extraction.py
jneuff: First, I think two distinct functions mean less complexity. Second, I thought a lot about what to put in the functions and what to leave out. By having all code related to the parallelization (including the setup of the Pool) it is easier to individually change it later. The idea is to have it completely unrelated.

tsfresh/feature_extraction/extraction.py
jneuff: 1. My first draft did not use an async map. However, splitting the dataframe is quite costly. The idea here is to interleave the splitting, calculation and combining. This results in an overall speed improvement. I should add a comment explaining that.
2. Yes it reads awful. However I'm not sure how to really improve it. I'll try to come up with better  naming of the variables. But it seems to me that for this part you just need to understand the relevant pandas fu.

tsfresh/feature_extraction/extraction.py
jneuff: I'll fix it.

tsfresh/feature_selection/feature_selector.py
jneuff: You are right. I'll change it.

tsfresh/feature_extraction/extraction.py
MaxBenChrist: What about: using per Kind as default if number of kind bigger than 1 otherwise by sample?

tsfresh/feature_extraction/extraction.py
MaxBenChrist: Yes, I think the concatenation in line 155 merges everything

MaxBenChrist: Line 163

MaxBenChrist: And yes, the feature calculator will at least return a nan.

tsfresh/feature_extraction/extraction.py
MaxBenChrist: +1 for two functions

tsfresh/feature_extraction/extraction.py
MaxBenChrist: let's remove it

tsfresh/feature_extraction/extraction.py
```diff
@@ -103,9 +123,106 @@ def extract_features(timeseries_container, feature_extraction_settings=None,
...
+                                                           settings=settings)
+    pool = Pool(settings.n_processes)
+
+    extracted_features = pool.map(partial_extract_features_for_one_time_series, kind_to_df_map.items(),
+                                  chunksize=settings.chunksize)
+
     pool.close()
+
+    # Concatenate all partial results
+    result = pd.concat([df_with_ids] + extracted_features, axis=1, join='outer', join_axes=[df_with_ids.index]) \
```
MaxBenChrist: lets remove the join_axes=[df_with_ids.index] part, concat with `join=outer`should be sufficient

tsfresh/feature_extraction/extraction.py
nils-braun: Ok, its 2 agains 1 now, but still I think that at least the first part of the function - which is perfectly the same in both - should be extracted then into a common function.

jneuff: If we can get rid of the ``df_with_ids`` only the creation of the partial and the setup of the Pool are common.
I would definitely leave the pool inside the functions as it enables us to later switch to e.g. joblib for per sample and keep the Pool for per kind. Also I wouldn't seperate the setup and closing of the pool into seperate functions.
For the partial you might be right, but it is one line and we might want to set it up differently in the future.

tsfresh/feature_extraction/extraction.py
jneuff: The question is if we can get rid of the `[df_with_ids] + extracted_features` part. I know it works in the per sample function as I only added it later. If I understand it correctly its only purpose is to ensure, the result contains every sample id. But if the calculators at least return ``NaN``, this is already covered.

jneuff:
I updated the PR, see the first comment.

However, we now have a failing unit test, I'll investigate this tomorrow.

tsfresh/feature_extraction/extraction.py
nils-braun: Ok, after the refactoring I see you point - now most of the functions are different. +1 from me.

jneuff:
The problem was that if a DataFrame is created from integers only, subsequent calculations are carried out as integer calculations. Parallelizing per sample increases the probability for an all integer df. That is why this problem surfaced now. Thanks @MaxBenChrist for the insights.

jneuff:
I think we can merge, @MaxBenChrist and @nils-braun

MaxBenChrist:
Huge effort and great work @jneuff. I will have a look at the changes later

nils-braun:
@jneuff Very good work. From my side, we can merge now.

MaxBenChrist:
Great work 👍

Issue #89, opened 2016-12-02
Remove Warnings from the user experience


ClimbsRocks:
I understand that printing all the UserWarnings and RuntimeWarnings is useful while developing on the source code of tsfresh, but as a user, it makes me wonder whether I screwed something up, or if there are bugs in the calculations themselves. 

I'm assuming it's all intentional, but that's an assumption I'd rather not have to make, and a not-ideal user experience. 

If you're ok with blocking warnings by default, I'm happy to implement that (I do something similar for [auto_ml](https://github.com/ClimbsRocks/auto_ml), since I assume it's my responsibility as the project's author to handle these warnings, and that my end users should not have to be bothered by my design choices).

Kiv:
I agree with this. It makes a bad first impression to try and run the example code in the Quick Start, and get a full page of warnings like:

```
/home/cmacleod/anaconda3/lib/python3.5/site-packages/pandas/computation/expressions.py:182: UserWarning: evaluating in Python space because the '*' operator is not supported by numexpr for the bool dtype, use '&' instead
  unsupported[op_str]))
/home/cmacleod/anaconda3/lib/python3.5/site-packages/pandas/computation/expressions.py:182: UserWarning: evaluating in Python space because the '*' operator is not supported by numexpr for the bool dtype, use '&' instead
  unsupported[op_str]))
/home/cmacleod/anaconda3/lib/python3.5/site-packages/pandas/computation/expressions.py:182: UserWarning: evaluating in Python space because the '*' operator is not supported by numexpr for the bool dtype, use '&' instead
  unsupported[op_str]))
/home/cmacleod/anaconda3/lib/python3.5/site-packages/pandas/computation/expressions.py:182: UserWarning: evaluating in Python space because the '*' operator is not supported by numexpr for the bool dtype, use '&' instead
  unsupported[op_str]))
/home/cmacleod/anaconda3/lib/python3.5/site-packages/statsmodels/regression/linear_model.py:1148: RuntimeWarning: divide by zero encountered in double_scalars
  return np.dot(wresid, wresid) / self.df_resid
/home/cmacleod/anaconda3/lib/python3.5/site-packages/statsmodels/regression/linear_model.py:1148: RuntimeWarning: divide by zero encountered in double_scalars
  return np.dot(wresid, wresid) / self.df_resid
[...]

PR #99, opened 2016-12-09
Introduced a new settings parameter ... 

nils-braun:
... to dismiss all warnings during extraction.

Fixing #89

### DIFF
--- a/tsfresh/feature_extraction/extraction.py
+++ b/tsfresh/feature_extraction/extraction.py
@@ -6,6 +6,9 @@ This module contains the main function to interact with tsfresh: extract feature
 """
 
 from __future__ import absolute_import, division
+
+import warnings
+
 from builtins import str
 from multiprocessing import Pool
 from functools import partial
@@ -13,6 +16,7 @@ import pandas as pd
 import numpy as np
 from tsfresh.utilities import dataframe_functions, profiling
 from tsfresh.feature_extraction.settings import FeatureExtractionSettings
+
 
 def extract_features(timeseries_container, feature_extraction_settings=None,
                      column_id=None, column_sort=None, column_kind=None, column_value=None):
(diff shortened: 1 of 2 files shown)


nils-braun:
I have not included this during feature selection, because there I think it is good to have the warnings displayed.

nils-braun:
Do we really need a test for this? Because coverage drops for this of course...

ClimbsRocks:
Nice, setting it as a user-selectable parameter is an even better implementation than I had in mind! Otherwise, that's very much in line with the code I've got for my own projects. 

Also great job having an editor that switches tabs to spaces :)

tsfresh/feature_extraction/extraction.py
```diff
@@ -172,29 +176,35 @@ def _extract_features_for_one_time_series(prefix_and_dataframe, column_id, colum
...
+            warnings.simplefilter("ignore")
+        else:
+            warnings.simplefilter("default")
+
+        if settings.set_default and column_prefix not in settings.kind_to_calculation_settings_mapping:
+            settings.set_default_parameters(column_prefix)
 
-    # Calculate the aggregation functions
-    column_name_to_aggregate_function = settings.get_aggregate_functions(column_prefix)
+        # Calculate the aggregation functions
```
MaxBenChrist: can we just stick with whitespaces?

MaxBenChrist: the diff gets bigger than it has to be ;)

MaxBenChrist:
@nils-braun I would prefer to keep the coverage high. here it does not seem essential, but maybe we could use a time series that causes a warning and compare if the warning get reported if we set the variable

tsfresh/feature_extraction/extraction.py
nils-braun: Ohm, these are whitespaces, but I have to increase the indention, because this is all in a context wrapper... or am I getting something wrong here? :-)

MaxBenChrist: aaah you are right

PR #100, opened 2016-12-09
Fix an error with division by zero

nils-braun:


nils-braun:
Sorry, this was wrong... I will update it

MaxBenChrist:
I opened a pr as well. you are too fast @nils-braun 

our calculation allowed negative mass, so the value of the feature calculator was not unique. i corrected this

PR #101, opened 2016-12-09
there is no such thing as negative mass

MaxBenChrist:


MaxBenChrist:
why does the coverage decrease ? :-/

good idea with the unit tests nils

MaxBenChrist:
do you think something is still missing @nils-braun? otherwise lets merge it

nils-braun:
I still don't understand why the coverage goes down, but yes, lets merge.

Issue #97, opened 2016-12-08
calculated_cwt encounter out of bounds


netivs:
As my data to abstracting features is a little too big (more than 8G),  I abstracting the relevant features on a sampled much more small data (1% ) using  `extract_relevant_features` and create a extractingsetting using `FeatureExtractionSettings.from_columns` and then to abstracting only the relevant features on the full data. Relative code is shown as followng:
```
df_filter_features = extract_relevant_features(x_filter_train,y_filter_train['label'], column_id="rank_id", column_sort="day_index")
feature_setting = FeatureExtractionSettings.from_columns(df_filter_features.columns)
df_features = extract_features(ts_ec, column_id="rank_id", column_sort="day_index",feature_extraction_settings=feature_setting)
```
The `extract_relevant_features` run all fine but after hours waiting, the `extract_features` on the full data encounter an error:
`
multiprocessing.pool.RemoteTraceback: 
"""
Traceback (most recent call last):
  File "D:\ouyang\Anaconda3\lib\multiprocessing\pool.py", line 119, in worker
    result = (True, func(*args, **kwds))
  File "D:\ouyang\Anaconda3\lib\multiprocessing\pool.py", line 44, in mapstar
    return list(map(*args))
  File "D:\ouyang\Anaconda3\lib\site-packages\tsfresh\feature_extraction\extraction.py", line 189, in _extract_features_for_one_time_series
    current_result = dataframe.groupby(column_id)[column_value].apply(apply_function, **kwargs).unstack()
  File "D:\ouyang\Anaconda3\lib\site-packages\pandas\core\groupby.py", line 694, in apply
    return self._python_apply_general(f)
  File "D:\ouyang\Anaconda3\lib\site-packages\pandas\core\groupby.py", line 698, in _python_apply_general
    self.axis)
  File "D:\ouyang\Anaconda3\lib\site-packages\pandas\core\groupby.py", line 1601, in apply
    res = f(group)
  File "D:\ouyang\Anaconda3\lib\site-packages\pandas\core\groupby.py", line 685, in f
    return func(g, *args, **kwargs)
  File "D:\ouyang\Anaconda3\lib\site-packages\tsfresh\feature_extraction\feature_calculators.py", line 77, in func_on_nonNumberObject
    return func(x, *arg, **args)
  File "D:\ouyang\Anaconda3\lib\site-packages\tsfresh\feature_extraction\feature_calculators.py", line 832, in cwt_coefficients
[...]

PR #108, opened 2016-12-13
check if coeff are out of cwt range

MaxBenChrist:
fixes issue #97


nils-braun:
Do you think it would be possible to implement a unittest for this? Especially for the case, that there is a NaN to see of the shape etc. stays the same.

MaxBenChrist:
sure, added one

Issue #107, opened 2016-12-12
IOError: bad message length


MaxBenChrist:
While working with a larger data set, I got a `IOError: bad message length` from the multiprocessing pool. I will try to reproduce the bug with random data.

I was trying to extract the MinimalProperties.

Maybe this will be solved by the new parallelizing by samples.

nils-braun:
Yes, this could maybe be solved but still we need to find out, why this is happening. 

This is a quite well known problem, and there are plenty of solutions out there (e.g. http://stackoverflow.com/questions/10721915/shared-memory-objects-in-python-multiprocessing). We just have to find one that suits our needs best :-)

nils-braun:
@MaxBenChrist Any news on that? Still problems?

MaxBenChrist:
I did not occur again and I did not had the time to reproduce it

MaxBenChrist:
Should be fine with our parallel extraction

PR #112, opened 2016-12-14
Tried to reduce the number of temporary copies 

nils-braun:
by using generator objects wherever possible.

This may solve #107 
@MaxBenChrist can you check with your sample?

For this I had to change the _extract function to return the kind and the dataframe.

### DIFF
--- a/tsfresh/feature_extraction/extraction.py
+++ b/tsfresh/feature_extraction/extraction.py
@@ -19,6 +19,7 @@ import numpy as np
 from tsfresh.utilities import dataframe_functions, profiling
 from tsfresh.feature_extraction.settings import FeatureExtractionSettings
 
+from itertools import groupby
 
 _logger = logging.getLogger(__name__)
 
@@ -157,7 +158,7 @@ def _extract_features_parallel_per_kind(kind_to_df_map, settings, column_id, col
     pool.close()
 
     # Concatenate all partial results
-    result = pd.concat(extracted_features, axis=1, join='outer').astype(np.float64)
+    result = pd.concat((df for _, df in extracted_features), axis=1, join='outer').astype(np.float64)
 
     pool.join()
     return result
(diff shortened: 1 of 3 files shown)


nils-braun:
Sorry, wrong button pressed...

tsfresh/feature_extraction/extraction.py
```diff
@@ -185,36 +186,39 @@ def _extract_features_parallel_per_sample(kind_to_df_map, settings, column_id, c
...
     pool.close()
+    pool.join()
 
-    # Wait for the jobs to complete and concatenate the partial results
-    dfs_per_kind = []
-    while not results_fifo.empty():
-        map_result = results_fifo.get()
-        dfs = map_result.get()
-        dfs_per_kind.append(pd.concat(dfs, axis=0).astype(np.float64))
+    # dfs_per_kind is now a list of (kind, df). We want to extract all dfs with the same kind and concatenate them
```
MaxBenChrist: then we should refactor `dfs_per_kind`, maybe `kind_and_df`

tsfresh/feature_extraction/extraction.py
```diff
@@ -185,36 +186,39 @@ def _extract_features_parallel_per_sample(kind_to_df_map, settings, column_id, c
...
+    # We have to make sure to not create unnecessary copies here.
+
+    # We will use groupby to group the list by their kind entry. The keyfunction extracts this kind for us.
+    keyfunction = lambda kind_and_df: kind_and_df[0]
```
MaxBenChrist: lets refactor `keyfunction` to maybe `first_i` or use anynymous functions `lambda x: x[0]`. keyfunction kind of implies that it is generating keys

tsfresh/feature_extraction/extraction.py
```diff
@@ -185,36 +186,39 @@ def _extract_features_parallel_per_sample(kind_to_df_map, settings, column_id, c
...
-        df_grouped_by_id = df_kind.groupby(column_id)
-        results_fifo.put(
-            pool.map_async(
-                partial_extract_features_for_one_time_series,
-                [(kind, df_group) for _, df_group in df_grouped_by_id],
-                chunksize=settings.chunksize
-            )
-        )
+    # Submit map jobs per kind per sample in a generator object
+    kind_and_df_group_list = ((kind, df_group)
```
MaxBenChrist: The choice of a tuple of tuples feels unnatural. Maybe use a [OrderedDict](https://docs.python.org/2/library/collections.html#collections.OrderedDict) or do you need the immutability?

MaxBenChrist:
Nice @nils-braun, 

I saw that you refactored the Queue() object. Can you give elaborate which effects this will have?

jneuff:
Good idea using generators! 👍 

Did you check the performance? I'll run some tests to compare it to master.

I was planning to move the imputing step (if enabled) into the while loop waiting for the async results. With the current approach this would not be possible. But I'm not sure how much performance this would gain us.

tsfresh/feature_extraction/extraction.py
jneuff: That decision was made when we introduced the multiprocessing Pool:
`extracted_features = pool.map(partial_extract_features_for_one_time_series, kind_to_df_map.items(),
                                  chunksize=settings.chunksize)`

MaxBenChrist: 🙈

jneuff:
As I said, I ran some performance tests:

- **1 time series, 16085 samples** PR: 416.93 s vs. master: 416.61 s
- **2 time series, 10624 samples** PR: 551.62 s vs. master: 541.91 s
- **3 time series, 10624 samples** PR: 832.12 s vs. master: 817.03 s

Disclaimer: I ran each test only once.

I also tried to steal your idea of using generators and applied it to the map_async:

```python
df_grouped_by_id = df_kind.groupby(
kind_df_groups = ((kind, df_group) for _, df_group in df_grouped_by_id)
results_fifo.put(
   pool.map_async(
       partial_extract_features_for_one_time_series,
       kind_df_groups,
       chunksize=settings.chunksize
    )
)
```

- **1 time series, 16085 samples** 419.54 s
- **2 time series, 10624 samples** 540.84 s
- **3 time series, 10624 samples** 819.68 s

I have yet to test how much we would gain on imputing per kind as I proposed above.

Performance is not everything, but I don't think the maintainability and clarity of using the async map with a queue is so much worse compared to the PR.

@nils-braun and @MaxBenChrist : What do you think?

nils-braun:
The idea behind this is more or less only to decrease the amount of memory used - and not that much for performance reasons Not using the queue object had only the reason to not use another additional concept.

nils-braun:
@jneuff Very nice study. I would definitely go for a solution using generators (because of the memory footprint) - but actually I do not care if yours or mine solution. If we go for mine, I will include Max' changes :-)

MaxBenChrist:
So, if I see that right, then the existing queue + pool.map_async + generator and Nils's pool.map nearly have the same computational costs / speed.

There are only differences in readability, right?

jneuff:
I would expect the speed to further diverge on larger data sets. And with the readability, I'd say they are on par. As Nils said, the async + queue solution is a bit more complicated as it contains more concepts.

I vote for my solution, as I'd like to introduce yet another change:
```python
    dfs_per_kind = []
    while not results_fifo.empty():
        map_result = results_fifo.get()
        dfs = map_result.get()
        df_this_kind = pd.concat(dfs, axis=0).astype(np.float64)
        if settings.IMPUTE is not None:
            df_this_kind = settings.IMPUTE(df_this_kind)
        dfs_per_kind.append(df_this_kind)

    result = pd.concat(dfs_per_kind, axis=1).astype(np.float64)
```

Here we might gain even more speed by already imputing some kinds while for others the features are still being calculated.

But the argument for a simpler (less concepts) solution is good and I am fine either way :-)

nils-braun:
Well, I think @jneuff already mentioned the important points here. I also do not have a real opinion on that - especially if @jneuff wannts to add some more features in his schema, I am totally ok with that. I just wanted to make sure, that we use generators here. By the way, could you @MaxBenChrist test already if one of those solutions above helps with larger data samples?

MaxBenChrist:
I am testing it on a big data frame at the moment. So, we will not merge this pr at the moment?

Issue #113, opened 2016-12-16
Travis has problems with HAT test


nils-braun:
The HAT unittest is failing in travis (see e.g. here https://travis-ci.org/blue-yonder/tsfresh/jobs/184653867).

Can someone maybe have a look, why this is happening? It seems the data is downloaded, but can not be found...

MaxBenChrist:
Indeed should be solved by pr 114

PR #114, opened 2016-12-16
Check for the UCI dataset

nils-braun:
... and not for the top folder to download the files.
Should solve #113


tsfresh/examples/har_dataset.py
```diff
@@ -47,7 +47,7 @@ def download_har_dataset():
     
     zipurl = 'https://github.com/MaxBenChrist/human-activity-dataset/blob/master/UCI%20HAR%20Dataset.zip?raw=true'
 
-    if os.path.exists(data_file_name):
+    if os.path.exists(data_file_name_dataset):
```
MaxBenChrist: can you change this to if os.path.exists(data_file_name_dataset) and if os.path.exists(data_file_name_classes) to make sure that both datasets and classes files are downloaded?

PR #116, opened 2016-12-18
fixed out of index error

MaxBenChrist:


MaxBenChrist:
It tried to think of a pr to catch or reproduce this. we can only catch future errors of this type by increasing the index numbers in our data fixture

Issue #88, opened 2016-12-02
Provide the user with some sense of progress while extract_features is calculating


ClimbsRocks:
Right now the only progress indicators are the Warning messages. 

I know that doing the computation inside an async map call makes this trickier. But even just knowing how many different things it's calculating, and reporting back every once in a while on how many total have been completed, would be really useful. I'm not sure whether to read a quick xkcd comic while waiting, or go home and let this run overnight.

MaxBenChrist:
Good point. If you extract features for different kind of time series this should be easily implemented.

earthgecko:
Can it at least be a log friendly progress indicator and not a some spinnings forward slashes, pipes and backslashes or some incrementally filling up progress bar which are visually pleasing but can be hell to logs.

Although we must think about the humans sitting there wondering how long it will take and if they have time to read a comic, we must also think of the machines that are conducting their work and calculating programmatically as well.  A progress indicator may not be of interest to them as they are not wondering about comics, but just trying to do the job assigned to them and calculate some features.

Hopefully it is possible to find a happy middle ground here that provides appropriate info to all executing parties, whether they be people or machines.

earthgecko:
Just try clarify, this point.

Universe Sandbox2 just presented me with an example of a progress indicator implementation that could be classified and classed as person_friendly and also not_great_for_a_machine_or_log.   It seemed relevant to share and I ironically choose Universe Sandbox2 over having to further contemplate the 50 / 3 Python paradox/dilemma (with which I have recently been confronted with) after further discussing and explaining it in detail to with my wife and dog on our evening walk.  She got it and agrees, our dog Dude has not weighed in his opinion yet, but I am not waiting on it.

### an example of a not so machine friendly progress implementation

by Steam (although it is not meant to be consumed programmatically), this is in no way a criticism of Steam's progress indicator, I find it quite informative and visually detailed and I like that, but I am person.
 
![an example of a not so machine friendly progress implementation](https://cloud.githubusercontent.com/assets/96679/20846883/9fb34b0c-b8c3-11e6-8230-f14db772e762.png)

ClimbsRocks:
@earthgecko : Yeah, I've had my logs wrecked by some of those terrible progress indicators you're talking about. They also seem like unnecessary complexity at this stage. I'm just hoping for something simple, that's also reassuring for first-time users as they sit wondering whether things are working on their dataset or not :)

nils-braun:
[...]

PR #117, opened 2016-12-18
Added a progress bar to the feature extraction.

nils-braun:
For this, I had to change quite some things:
* I added tqdm to the requirements, as we use this for drawing the progress bar
* I had to change each parallel map to an imap, as I need to know immediately, if and when a new result is there (otherwise the function will only continue, when all results are there)
* imap needs a proper chunksize - so I copied the calculation from the multiprocessing.map function
* The result getting had to be wrapped with a tqdm. In case of per_kind this is easy - in case of per_sample this needed some more code...

This should solve #88

### DIFF
--- a/tsfresh/feature_extraction/extraction.py
+++ b/tsfresh/feature_extraction/extraction.py
@@ -250,6 +250,8 @@ def _calculate_best_chunksize(iterable_list, settings):
     :param iterable_list: A list which defines how many calculations there need to be.
     :param settings: The settings object where the chunksize may already be given (or not).
     :return: The chunksize which should be used.
+
+    TODO: Investigate which is the best chunk size for different settings.
     """
     if not settings.chunksize:
         chunksize, extra = divmod(len(iterable_list), settings.n_processes * 5)
--- a/tsfresh/feature_extraction/extraction.py
+++ b/tsfresh/feature_extraction/extraction.py
@@ -245,13 +245,14 @@ def _calculate_best_chunksize(iterable_list, settings):
     """
     Helper function to calculate the best chunksize for a given number of elements to calculate,
     or use the one in the settings object.
-    The chunksize calculation taken from map_async of multiprocessing module.
+
+    The formula is more or less an empirical result.
     :param iterable_list: A list which defines how many calculations there need to be.
     :param settings: The settings object where the chunksize may already be given (or not).
     :return: The chunksize which should be used.
     """
     if not settings.chunksize:
-        chunksize, extra = divmod(len(iterable_list), settings.n_processes * 4)
+        chunksize, extra = divmod(len(iterable_list), settings.n_processes * 5)
         if extra:
             chunksize += 1
     else:
--- a/tsfresh/feature_extraction/extraction.py
+++ b/tsfresh/feature_extraction/extraction.py
@@ -16,6 +16,8 @@ import logging
 import pandas as pd
 import numpy as np
 
+from tqdm import tqdm
+
 from tsfresh.utilities import dataframe_functions, profiling
 from tsfresh.feature_extraction.settings import FeatureExtractionSettings
 
@@ -151,8 +153,11 @@ def _extract_features_parallel_per_kind(kind_to_df_map, settings, column_id, col
                                                            settings=settings)
     pool = Pool(settings.n_processes)
 
-    extracted_features = pool.map(partial_extract_features_for_one_time_series, kind_to_df_map.items(),
-                                  chunksize=settings.chunksize)
+    chunksize = _calculate_best_chunksize(kind_to_df_map, settings)
+
+    total_number_of_expected_results = len(kind_to_df_map)
+    extracted_features = tqdm(pool.imap(partial_extract_features_for_one_time_series, kind_to_df_map.items(),
+                                        chunksize=chunksize), total=total_number_of_expected_results)
 
     pool.close()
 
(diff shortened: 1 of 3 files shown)


nils-braun:
If you want to test it - best use an IPython notebook and do the quickstart.

tsfresh/feature_extraction/extraction.py
```diff
@@ -190,32 +195,68 @@ def _extract_features_parallel_per_sample(kind_to_df_map, settings, column_id, c
...
     for kind, df_kind in kind_to_df_map.items():
         df_grouped_by_id = df_kind.groupby(column_id)
+
+        total_number_of_expected_results += len(df_grouped_by_id)
+
+        chunksize = _calculate_best_chunksize(df_grouped_by_id, settings)
+
         results_fifo.put(
-            pool.map_async(
+            pool.imap(
```
MaxBenChrist: implications of imap versus map_async?

MaxBenChrist:
I had a telefon call with nils. we discussed the changes, the only thing that is unclear at this moments are the implications of imap versus map_async.

can you have a look at this pr as well, @jneuff ?

tsfresh/feature_extraction/extraction.py
```diff
@@ -151,8 +153,11 @@ def _extract_features_parallel_per_kind(kind_to_df_map, settings, column_id, col
                                                            settings=settings)
     pool = Pool(settings.n_processes)
 
-    extracted_features = pool.map(partial_extract_features_for_one_time_series, kind_to_df_map.items(),
-                                  chunksize=settings.chunksize)
+    chunksize = _calculate_best_chunksize(kind_to_df_map, settings)
+
+    total_number_of_expected_results = len(kind_to_df_map)
+    extracted_features = tqdm(pool.imap(partial_extract_features_for_one_time_series, kind_to_df_map.items(),
```
earthgecko: Would this not be better suited as also being optional, even if default is on, but can be disabled?  If tsfresh runs constantly (in app), all the tqdm outputs are not going to be or any use and are just going to cost additional tiny ticks and fill app logs unnecessarily.  Perhaps a future pull request waiting to happen :)

MaxBenChrist: We can just add a flag to the almighty settings object to steer if tqdm is used or not, right?

tsfresh/feature_extraction/extraction.py
jneuff: I have no idea. @nils-braun Wouldn't it better to use an imap_unordered? The ordering shouldn't matter to us and I'd suspect that unordered is faster.

I'll measure the performance of using the imap vs the map_async. If it is noticeably slower, we could revert to map_async and provide a less granular progress.

tsfresh/feature_extraction/extraction.py
```diff
@@ -190,32 +195,68 @@ def _extract_features_parallel_per_sample(kind_to_df_map, settings, column_id, c
...
+            dfs_per_kind.append(pd.concat(dfs, axis=0).astype(np.float64))
+
+        result = pd.concat(dfs_per_kind, axis=1).astype(np.float64)
+
+        pool.join()
+        return result
+
+
+def _calculate_best_chunksize(iterable_list, settings):
+    """
```
jneuff: Isn't this the same as just letting Pool calculate the chunk size?

tsfresh/feature_extraction/extraction.py
```diff
@@ -190,32 +195,68 @@ def _extract_features_parallel_per_sample(kind_to_df_map, settings, column_id, c
...
+    Helper function to calculate the best chunksize for a given number of elements to calculate,
+    or use the one in the settings object.
+    The chunksize calculation taken from map_async of multiprocessing module.
```
jneuff: Is this enough in terms of the Python license?

```
#
# Module providing the `Pool` class for managing a process pool
#
# multiprocessing/pool.py
#
# Copyright (c) 2006-2008, R Oudkerk
# All rights reserved.
#
# Redistribution and use in source and binary forms, with or without
# modification, are permitted provided that the following conditions
# are met:
#
# 1. Redistributions of source code must retain the above copyright
#    notice, this list of conditions and the following disclaimer.
# 2. Redistributions in binary form must reproduce the above copyright
#    notice, this list of conditions and the following disclaimer in the
#    documentation and/or other materials provided with the distribution.
# 3. Neither the name of author nor the names of any contributors may be
#    used to endorse or promote products derived from this software
#    without specific prior written permission.
#
# THIS SOFTWARE IS PROVIDED BY THE AUTHOR AND CONTRIBUTORS "AS IS" AND
# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
# ARE DISCLAIMED.  IN NO EVENT SHALL THE AUTHOR OR CONTRIBUTORS BE LIABLE
# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
# OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
# HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY
# OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
# SUCH DAMAGE.
#
```

I think we would have to reproduce the complete notice.

MaxBenChrist:
Good job @earthgecko !

tsfresh/feature_extraction/extraction.py
```diff
@@ -190,32 +195,69 @@ def _extract_features_parallel_per_sample(kind_to_df_map, settings, column_id, c
...
+    Helper function to calculate the best chunksize for a given number of elements to calculate,
+    or use the one in the settings object.
+
+    The formula is more or less an empirical result.
+    :param iterable_list: A list which defines how many calculations there need to be.
+    :param settings: The settings object where the chunksize may already be given (or not).
+    :return: The chunksize which should be used.
+    """
+    if not settings.chunksize:
+        chunksize, extra = divmod(len(iterable_list), settings.n_processes * 5)
```
MaxBenChrist: maybe add a todo here that we investigate the optimal chunksize this in the future

MaxBenChrist:
This is so awesome, @nils-braun. The progress bar improves the user experience so much.

Issue #74, opened 2016-11-23
Classifying the feature calculators.


MaxBenChrist:
We can use our `@set_property(` decorator to give the feature calculators objects properties. 

This can be for example used to classify `fast` calculators such as `@set_property("minimal", True)`. We could use such properties to classify our feature calculators even further. For example we could denote calculators that are *"time independent"* or that are *"medium costly"* or highly *"highly costly"*. 

This issue should be a starting point for a discussion.

nils-braun:
Some hints on how to find those functions (and already many fixes) are in #186

nils-braun:
We have a "high_comp_cost" setting which does exactly this.

PR #121, opened 2016-12-19
Improve performance

MaxBenChrist:
I started comparing the runtime of individual feature calculator as in issue #74 

turns out that "sample_entropy" and "approximate_entropy" take forever to calculate. I marked them with the attribute "high_comp_cost". 

Further, now the constructor of the FeatureExtractionSettings object will not calculate those "high_comp_cost" features.


MaxBenChrist:
I am not sure why the coverage decreases

MaxBenChrist:
Ah because of the driftbif_datasets.py, I added it to be omited

Issue #122, opened 2016-12-19
Add reasonable feature extraction settings


MaxBenChrist:
The current build is failing because earthgeckko's snapshot can not find the entropy features whose calculation was removed in 8a586ee1d11ffa7524f3b0d223552dbf197c0e74

to mitigate this we could offer three extraction objects

1.    FeatureExtractionSettings => all features
2.   ReasonableFeatureExtractionSettings => faster, calculates all but highly complex features. all features without attribute "high_comp_cost"
3.   MinimalFeatureExtractionSettings => fastest, only the basic features

nils-braun:
This was implemented and will be refined in the next major version.

PR #124, opened 2016-12-19
added ReasonableFeatureExtractionSettings

MaxBenChrist:
please do not merge this yet. there are still some unit tests missing for the ReasonableFeatureExtractionSettings object.

this will fix the failing unit tests, the build error and issue #122

### DIFF
--- a/tsfresh/feature_extraction/settings.py
+++ b/tsfresh/feature_extraction/settings.py
@@ -103,11 +103,6 @@ class FeatureExtractionSettings(object):
                 "approximate_entropy": [{"m": 2, "r": r} for r in [.1, .3, .5, .7, .9]]
             })
 
-        # drop all features with high computational costs
-        for fname, f in six.iteritems(feature_calculators.__dict__):
-            if hasattr(f, "high_comp_cost"):
-                del self.name_to_param[fname]
-
         # default None means one procesqs per cpu
         n_cores = int(os.getenv("NUMBER_OF_CPUS") or cpu_count())
         self.n_processes = max(1, n_cores//2)
@@ -358,3 +353,26 @@ class MinimalFeatureExtractionSettings(FeatureExtractionSettings):
                 name_to_param_copy[feature_calculator] = self.name_to_param[feature_calculator]
 
         self.name_to_param = name_to_param_copy
+
+
+class ReasonableFeatureExtractionSettings(FeatureExtractionSettings):
+    """
+    This class is a parent class of the FeatureExtractionSettings class
+    and has the same functionality as its base class.
+
+    The only difference is, that the features with high computational costs are not calculated. Those are denoted by
+    the attribute "high_comp_cost"
+
+    You should use this object when calling the extract function, like so:
+
+    >>> from tsfresh.feature_extraction import extract_features, ReasonableFeatureExtractionSettings
+    >>> extract_features(df, feature_extraction_settings=ReasonableFeatureExtractionSettings)
+    """
+
+    def __init__(self):
+        FeatureExtractionSettings.__init__(self, True)
+
+        # drop all features with high computational costs
+        for fname, f in six.iteritems(feature_calculators.__dict__):
+            if hasattr(f, "high_comp_cost"):
+                del self.name_to_param[fname]
--- a/tsfresh/feature_extraction/settings.py
+++ b/tsfresh/feature_extraction/settings.py
@@ -328,7 +328,7 @@ class FeatureExtractionSettings(object):
 
 class MinimalFeatureExtractionSettings(FeatureExtractionSettings):
     """
-    This class is a parent class of the FeatureExtractionSettings class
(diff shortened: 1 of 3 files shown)


MaxBenChrist:
added unit tests. this will repair the broken build

tsfresh/feature_extraction/settings.py
```diff
@@ -357,4 +352,27 @@ def __init__(self):
...
                 name_to_param_copy[feature_calculator] = self.name_to_param[feature_calculator]
 
-        self.name_to_param = name_to_param_copy
\ No newline at end of file
+        self.name_to_param = name_to_param_copy
+
+
+class ReasonableFeatureExtractionSettings(FeatureExtractionSettings):
+    """
+    This class is a parent class of the FeatureExtractionSettings class
```
jneuff: typo: child class

nils-braun:
Should we actually make this the default?

earthgecko:
@nils-braun :+1: 

I spent a few hours debugging why timeseries with very little variability take so long.
The example timeseries below (all 15.0, except the last datapoint) is taking ~578 seconds
to extract features from, where as mostly all other timeseries of the same length
are taking > 7 seconds.

```
timestamps = []
values = []
count = 0
ts = 1482644940
while count < 10075:
    count += 1
    timestamps.append(ts)
    values.append(15.0)
    ts += 60

timeseries = map(list, zip('t' * 10075, timestamps, values))
timeseries[-1][2] = 16.0

df = pd.DataFrame(timeseries, columns=['metric', 'timestamp', 'value'])
df_features = extract_features(df, column_id='metric', column_sort='timestamp', column_kind=None, column_value=None)
```

Using the new ReasonableFeatureExtractionSettings solved these edge cases, ~1.78 seconds with the loss of 6 features, so 210 features instead of 216.

```
tsf_settings = ReasonableFeatureExtractionSettings()
# Disable tqdm progress bar
tsf_settings.disable_progressbar = True
df_features = extract_features(
    df, column_id='metric', column_sort='timestamp', column_kind=None,
    column_value=None, feature_extraction_settings=tsf_settings)
```

I am very happy that @MaxBenChrist added ReasonableFeatureExtractionSettings and
I would say that it would probably be useful as default as it is going to catch
a lot of edge cases, like this type of timeseries described above.

It suggests that one or more of the tsfresh algorithms are calculating using
a sequence which is being appended for every same value in sequence and then
calculating the result and repeating the same operation over and over.  In most
timeseries this/these algorithms would probably break out of the sequence and
reset as soon as the next value was different from the last?

MaxBenChrist:
> I am very happy that @MaxBenChrist added ReasonableFeatureExtractionSettings and
I would say that it would probably be useful as default as it is going to catch
a lot of edge cases, like this type of timeseries described above.

As we have a semantic versioning scheme, this could be interpreted as a major change (a major version is making incompatible API changes). This is something I would prefer to not do at the moment.

> It suggests that one or more of the tsfresh algorithms are calculating using
a sequence which is being appended for every same value in sequence and then
calculating the result and repeating the same operation over and over. In most
timeseries this/these algorithms would probably break out of the sequence and
reset as soon as the next value was different from the last?

I am not sure what you are proposing here, @earthgecko. By the way, you can use constructions like [15.0] *1000 to get your time series lists faster.

earthgecko:
@MaxBenChrist quite sensible considering semantic versioning, perhaps maybe be
suitable for v1.x if it is considered overall beneficial.

Well on the culprit algorithm/s it is not a bug, I am sure that calculations are
correct, I was just wondering if it would be possible to optimise the running of
sample_entropy or approximate_entropy any further.  Not optimizing the algorithms
in any way, just adding some sanity checks to assess if they should run or not.
That may be non-trivial in some case and result in "skewed" im terms of repeatability,
if an algorithm is only run if not some_condition, which would create a state.
For example if a timeseries has a certain "shape" shapes defined by certain
features e.g. range, unique values, etc then do not run x high_comp_cost algorithm.

Not desirable.  Not semver but perhaps something that may be useful in the future.

However that said, I would say that the cost of running to high_comp_cost
algorithms on the edge cases is exorbitantly high, given the possible timeseries
that are being thrown at tsfresh.  There have been people that have run into
long runtimes and people are going to encounter these edge cases.

So if the API cannot change, the documentation could now promote the use of 
ReasonableFeatureExtractionSettings a bit more and specifically describe some
now known possible pitfalls that may be encountered running the default
settings on extract_features, unless long runtimes and a times very variable in
runtimes can be expected and handled.  Not all timeseries created equally :)

On the "you can use constructions" I normally do, but I did not want to have to
upload a csv timeseries file with the example and match the data and method to
reproduce injecting the metric string into the map constructor seemed problematic

```
timeseries = map(list, zip('t' * 10075, map(int, range(int(ts) - 604740, int(ts) + 60)), [15.0] * 10075))
```

Does not behave the same I tried and hammered around a bit with it and then I
just decide to get it **done** instead of trail and error how to map it :)

PR #126, opened 2016-12-20
Fix travis

MaxBenChrist:
the snapshot tests on travis are failing because it gets a dirty version tag (0.4.0+ng7e4c397.dirty).

Travis compares the calculated features to a non existing baseline file. this pr will solve it


MaxBenChrist:
@earthgecko, you refactored your code a little bit. tried to make it more pythonic ;)

PR #127, opened 2016-12-21
Driftbif simulation

kempa-liehr:
Create regression datasets from DriftBif simulation


MaxBenChrist:
You have to fix the import driftbif_simulations in examples/__init__.py

At the moment, `from tsfresh.examples import load_driftbif` fails with `No module named driftbif_datasets`

MaxBenChrist:
Does this now works with arbitrary number of dimensions? (e.g. 1 or 3 dimensions?)

kempa-liehr:
Yes, it does.

kempa-liehr:
Just pushed the bugfix.

MaxBenChrist:
Awesome, @kempa-liehr. out of curiousity: One can increase the number of dimensions over 3. does this still makes sense from a physicist point of view?

kempa-liehr:
@MaxBenChrist, for this specific model even three spatial
dimensions are kind of extrapolation. This is because, the
ODE we are solving has been derived from a phenomenological
two-dimensional PDE modelling a semiconductor gas discharge
systen. But having one more spatial dimensions, well,
always becomes interesting. But four of them. No, it does
not make sense.

MaxBenChrist:
okay, somaybe we can add a small warning that this data set is only well defined for one or two dimensions and m>2 does not make sense?

kempa-liehr:
Good point. I am going to add the interpretation for
m=1,2,3  and a "generalization warning" for m>=4.

On Thu, 22 Dec 2016 05:12:17 -0800
 MaxBenChrist <notifications@github.com> wrote:
> okay, somaybe we can add a small warning that this data
> set is only well defined for one or two dimensions and
> m>2 does not make sense?
> 
> -- a
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
>
https://github.com/blue-yonder/tsfresh/pull/127#issuecomment-268796772

==========================================================
Dr. Andreas Kempa-Liehr
Freiburger Materialforschungszentrum
Stefan-Meier-Str. 21
D-79104 Freiburg i. Br.
===========================================================

PR #132, opened 2017-01-03
Removed the non-feature-calculator settings

nils-braun:
in favour of parameters of the different functions.


nils-braun:
Will be open again soon....

Issue #61, opened 2016-11-15
Explain the FeatureExtractionSettings object in documentation


MaxBenChrist:
Especially how to control which features are calculated with `setting.name_to_param` dictionary.

MaxBenChrist:
Further we should explain how to prevent unnecessary features being calculated with the from columns method

MaxBenChrist:
@nils-braun Keen to write some documentation? :)

nils-braun:
Will be fixed with #133 - so in the next version.

MaxBenChrist:
is obsolete like the extraction settings object :D

nils-braun:
:+1: I like solving issues by deleting the code they are referring to :-D

PR #133, opened 2017-01-03
Feature extraction settings refactoring

nils-braun:
Solves also #61 

This is a breaking change, so we should probably increase the major version after that.


MaxBenChrist:
reporting is gone?

MaxBenChrist:
Background: we had a maintainer meeting today. We are about to ditch the settings objects to make tsfresh more user friendly

MaxBenChrist:
Great guys!

I had a look at the changes and would like to discuss the following points:

1. Why do we still have FeatureExtractionSettings ? lets just make them a dictionary?
2. Lets have three new modes of extracting features: Minimal, Smart und Full with smart being the defaul. Its easier for the user to select the proper calculation than with the current naming
3. I would refactor a few things (e.g. rename the `from_columns` or `select_features` methods)
4. What is the point behind the default.py? It will lay somewhere in /site-packages/ when tsfresh is installed. Not many users will change it. As a drawback, the default values do not appear in the docstrings in the current setup. 
5. Regarding your question `# TODO: Do we want to keep the flag `evaluate_only_added_features`?` I would save yes, but I would rename it to `filter_only_time_series_featues`

I am fine with merging this beast of a pr, but not into master. I would open a develop branch, merge this thing into it and then tackle the other points I mentioned above. Later, when we have everything ready we merge into master and jump to 0.5.0.

MaxBenChrist:
Also, regarding the dev branch: I want to provide 1-2 notebooks with examples of all the new and old functionalities. They should from the beginning of 0.5.0 so users can switch to the new api

jneuff:
1. The class inherits from dictionary. It seemed odd to me too at first. It's just a way to do the setup. I don't know what the most pythonic way would be. But I do think we could improve the naming. Why not call it ExtractionSettings? The prefix feature seems redundant to me. Also the named parameters could be called extraction_settings.
2. I think reasonable is a better name than smart. But maybe we can come up with something better. The name should reflect the fact that it calculates most features but is way faster than to calculate all features.
4. The module default.py is needed as the default settings are accessed from various functions and we should have them in one place. It is not intended to be changed by the user. To change the settings one can set the named parameters.
5. This TODO has been there for quite a while. We could change the parameter's name of course while we're busy breaking things. I would just leave it. But now is probably a good time to remove the TODO.
6. We should definitely have two dictionaries. It's way easier to handle if you want to have the same settings for several kinds and just alter the settings for one kind.

We could just take care of the remaining points in this branch and then merge the pr into master. I don't see why we should open another branch.

Regarding the reporting: In my opinion it wasn't flexible enough this way. It's better to get rid of it now than having to stick with it. We should rather aim to make more internal information available as proposed in PR #87 .


Remaining points we need to tackle:

- Typos in the docs (Where?)
- The notebooks have yet to be updated
- Change some of the names (Which?)
- Remove the TODO mentioned in point 5 above

MaxBenChrist:
> 1. The class inherits from dictionary. It seemed odd to me too at first. It's just a way to do the setup. I don't know what the most pythonic way would be. But I do think we could improve the naming. Why not call it ExtractionSettings? The prefix feature seems redundant to me. Also the named parameters could be called extraction_settings.

Just return a dictionary?

> 2. I think reasonable is a better name than smart. But maybe we can come up with something better. The name should reflect the fact that it calculates most features but is way faster than to calculate all features.

Minimal, Normal, Extended ? 

> 3. The module default.py is needed as the default settings are accessed from various functions and we should have them in one place. It is not intended to be changed by the user. To change the settings one can set the named parameters.

Okay so keep it. 

> 5. We should definitely have two dictionaries. It's way easier to handle if you want to have the same settings for several kinds and just alter the settings for one kind.

I still think that it is confusing...

MaxBenChrist:
Another thought regarding 5.:

> We should definitely have two dictionaries. It's way easier to handle if you want to have the same settings for several kinds and just alter the settings for one kind.

You can just duplicate the dictionary for those several kind.

MaxBenChrist:
An example: lets say you want to apply settings 
`A = {"median": None}` to time series of kind "kind1" 
and 
`B = {"max":  None, "min": None}` to the remaining time series (kind "kind2" and "kind3")

with your proposal you would have to call `extract_features(..., default_calculation_settings_mapping=B, kind_to_calculation_settings_mapping = {"kind1": A} ,...)`

with mine it becomes

`extract_features(...,  settings_mapping={"kind1": A, "kind2": B, "kind3":B} ,...)` or just 
`extract_features(...,  settings_mapping= B,...)` if you want to apply B it to all time series

MaxBenChrist:
Shouldn't we have two chunksizes? one for the extraction (=number of rows) and one for the selection (=number of tested feature columns)

MaxBenChrist:
I refactored some stuff and send in two prs for controversial decisions

Remaining points we need to tackle:
- The notebooks have yet to be updated
- Introduce two chunksizes for extraction and selection
- discuss my proposal regarding extraction settings from above 
- make feature extraction settings a dictionary

jneuff:
> Shouldn't we have two chunksizes? one for the extraction (=number of rows) and one for the selection (=number of tested feature columns)

We could have that. But if the need arises, the user could also call `extract_features` and then `select_features` each with a different chunk size.

Regarding the two parameter dictionaries discussion (point 5): If we only have one, we need a '*' affecting all timeseries. Otherwise, if you have a lot of timeseries and just want to change a few of them, you need to do a lot of setup work just to call our function.

On point 1 above:
> Just return a dictionary?

Where and from what?

I gave this some more thought. Although the settings class now behaves like a plain old dictionary, you still have to instantiate it. Better we setup the dictionaries as variables in the module.

MaxBenChrist:
> We could have that. But if the need arises, the user could also call extract_features and then select_features each with a different chunk size.

If I read chunksize, I think about rows not about columns. I would rename the variable. Maybe `chunk_row` and `chunk_col` to make it more clear which chunksize we are refering to.

> I gave this some more thought. Although the settings class now behaves like a plain old dictionary, you still have to instantiate it. Better we setup the dictionaries as variables in the module.

:+1 for the plain dictionaries as variable solution.

nils-braun:
To 5.: @jneuff gave the right comment. It is maybe easier of you have only two kinds, but you will probably be annoyed if you have to setup the same settings object for 1000 kinds because you only wanted to change the kind number 1001...
I do not think that two dictionaries is inconvenient, because in most of the times you want to keep the default settings for most of the kinds and only change one or two, this means you only have to pass one of the dicts (because the other has a sane default). Additionaly, users can still work as you proposed and just do never touch the `default_calculation_settings_mapping`.

To the dictionaries: the initialization (now in the constructor of `FeatureExtractionSettings`) needs some calculation and "work" as one has to go though all feature calculators. If you want to do this as a plain dictionary, you would have to initialize them in the global section of the module, which means it will be reevaluated everytime the module is loaded. Or you implement some nasty caching algorithm...
Keeping it as a class still makes it possible for turning it back to a "normal" class later without changing the API and people will easier identify it as a "thing".

Renaming the chunksize is a good idea.

docs/text/feature_extraction_settings.rst
```diff
@@ -8,8 +8,9 @@ function and select only the relevant features later. Then you can call the func
 `kind_to_calculation_settings_mapping` object, which means you are using the default options (which will use all
 feature calculators in this package).
 
-But if you want to dig deeper into your data and use custom settings for the feature extractors, you
-have to use a custom settings object, like shown here:
+After digging deeper into your data, you maybe want to calculate more of a certain type of feature and less of another
```
nils-braun: I do not like the "more" and "less". Maybe better "with other/more parameter combinations".

tsfresh/examples/driftbif_simulation.py
```diff
@@ -133,6 +135,10 @@ def load_driftbif(n, l, m=2, classification=True):
     # todo: add ratio of classes
     # todo: add start seed
 
+    if m > 2:
+        logging.warning("You set the dimension parameter for the dissipative soliton to m={}, however it is only"
+                        "properly defined for m=1 or m=2.".format(m))
+
```
nils-braun: this is fine, but it has nothing to do with this PR....

nils-braun:
@MaxBenChrist Yes, I like the new renaming! :-)

MaxBenChrist:
From my perspective the solution with two dictionaries is fine after renaming the settings to `fc_parameters. I like `default_fc_parameters` and `kind_to_fc_parameters`.

So, what is still to do:
- make feature extraction settings a dictionary
- Introduce two chunksizes for extraction and selection

After this is done
- The notebooks have yet to be updated and
- The documentation should be checked and be aligned with new api

MaxBenChrist:
So, what is still to do? From my perspective:

- Introduce two chunksizes for extraction and selection
- make feature extraction settings a plain dictionary

Finally

- the notebooks have yet to be updated
- do a final pass over the documentation

tsfresh/examples/driftbif_simulation.py
MaxBenChrist: Makes sense to separate the refactoring from other changes

kempa-liehr: And the warning should be triggered for m>3.

nils-braun:
@MaxBenChrist I am still against making the settings objects pure dicts. The user will never notice (because it feels like a dict and even `isinstance(settings, dict)` is `True`) and for the developer it is much cleaner and easier to really define the objects in their constructor. Plus, we still have the possibility to go back to "real" objects later.

MaxBenChrist:
I would close this opportunity, the object fc_parameters should contain the name of the feature calculator as key and the repective parameters ( not settings or configurations but really the parameters of the respective function calculator) as Values. It should not contain any other parameters as for example an attribute. This was the motivation behind the ditching of the old clumsy settings object.

Judt having a dictionary makes the  the object easy understandable and pythonic in my Opinion. So when a user inspects the fc_parameters object he immediately knows how to work with that

nils-braun:
Ok guys, we should converge to a decision here... Otherwise we will never go to the next level :-)

MaxBenChrist:
yep.

I talked with @jneuff today and as I understood he also is in favor of using plain dictionaries. So lets use plain dictionaries as the majority of developers is in favor for it.

MaxBenChrist:
Lets merge this during the next week. 

I can update this pr around Wednesday.

MaxBenChrist:
I finally was able to solve all merge conflicts and issues. Took me several hours. Lessons learned: Do not wait with merging a branch with that many changes.

things left to do:

- check doc strings
- check coverage
- check notebooks
- maybe write a short page for documentation about the changes

jneuff:
Thanks @MaxBenChrist for finishing this major PR! I'd say it is ready to be merged.

As an explanation of the changes, a short paragraph will suffice:
____________________________________________________________________________
Breaking change: Removal of FeatureExtractionSettings and FeatureSignificanceTestSettings
----------------------------------------------------------------------------------------------------------------------------

With the release of tsfresh X.Y.Z we removed the `FeatureExtractionSettings` and `FeatureSignificanceTestSettings`. Their purpose was to: 

1. Specify the features to be etracted along with their parameters
2. Add additional settings to the `extract_features` and `select_features` functions
 
For the first use case, we now provide the much simpler classes `MinimalFCParameters`, `EfficientFCParameters` and `ComprehensiveFCParameters`. These behave just like dictionaries. which is all we need, as we map features to parameters.
We recognized the second use case as a poor design choice and now the respective settings are passed directly as arguments to `extract_features` and `select_features`.

If one of these use cases has been relevant to you, please take a look at the `Feature extraction settings` chapter of the documentation and the docstrings for `extract_features` and `select_features` to get a more detailed understanding about the changes.
___________________________________________________________________________________

But I'm not quite sure where to put it. I don't actually see the need to create a chapter in the docs for it. We might put it into the README or just the changelog. What do you think?

nils-braun:
If we want to put something like this in the README, we better give more examples on how to use it (and what to use it for), instead of describing our motivations and intentions. This would better go into some kind of "blog post" (if we would have one)

nils-braun:
At least I could increase the coverage a bit again ;-)

nils-braun:
From the code side, I would also say it is ready to merge. The documentation may still need some tweaking - or has anyone taken care of this also (I mean, we have already included much of the needed docu in this PR before, but things may have diverged from master)

nils-braun:
Ok, I have had a look by myself. Docu seems fine for me. Ready to merge (if the notebooks are fine)

MaxBenChrist:
great teamwork!

Issue #123, opened 2016-12-19
Impute takes quite some time


nils-braun:
If we want to be faster, we probably also have to make sure the "impute" function is not that slow...

MaxBenChrist:
Good point!

F-A:
Hi all,

I think one of the problem is tha youare iterating through the columns of the dataframe. When you have a lot of ime series and extract a lot of features you get a huge number of columns and this iteration takes a lot of time...

For example, using the tsfresh.utilities.dataframe_functions.get_range_values_per_column function on a dataframe with 200 columns and 60 indeces I get an execution time for this function of 329ms (best of 3 execution on a total of 100 with %timeit).

I wrote an alternative function working on the numpy array directly:
```
 def get_range_values_per_column(df):
    """
    Retrieves the finite max, min and mean values per column in df and stores them in three dictionaries, each mapping
    from column name to value. If a column does not contain finite value, 0 is stored instead.
    :param df: Dataframe
    :return: Dictionaries mapping column names to max, min, mean values
    """
    column = df.get_values()
    masked = np.ma.masked_invalid(column)
    
    columns = df.columns
    
    col_to_max     = dict( zip(columns, np.max(masked,axis=0) ) )
    col_to_min     = dict( zip(columns, np.min(masked,axis=0) ) )
    col_to_median  = dict( zip(columns, np.median(masked,axis=0) ) )
        
        
    return col_to_max, col_to_min, col_to_median
```

It get an execution time of 25ms (same measure). This function still needs to be more tested (it still gets a problem with median in some cases) and maybe improved but this could give the right direction for improving the performance.

For the record, I found this out, because I wanted to detect features on 60 EEG channels. The feature extraction was still usable but the impute function took way to long so I stopped it...

jneuff:
Nice investigation! Are you planning to cover all cases with your function? Can you open a pull request with your proposal?

F-A:
To be honest I'm not sure how much time I can spend on this... I just planned to try out tsfresh to classify EEG trials but the impute function was to slow for me, so I looked at the function and just wanted to document my investigation. 
I'm currently looking at the impute function, to see if one can do something similar.
Depending on how far I come, and if my function make through the unit tests I'll try to do a PR.

MaxBenChrist:
@F-A have you tried to install the latest master version? It will impute early on, after each feature is calculated. That should speed up the calculation further.

MaxBenChrist:
[...]

PR #135, opened 2017-01-06
Improve performance of the impute function.

F-A:
Improve performance of the functions:
- utilities.dataframe_function.get_range_values_per_column(df)
- utilities.dataframe_function.impute(df_impute)
More specifically: apply the impute function directly on numpy array to improve computation time.

Now the impute function runs in 109ms (60 samples, 14256 features (or columns) ).

Note: I did not impoved the performance of impute_dataframe_range(...) since it would have been to much of a hassle to implement all the checks in that function, e.g. in case the min/max/median values of each columns are not present. In our case we call the get_range_values_per_column just before so these checks are not necessary.
So I just reimplemented the function impute_dataframe_rage directly in the impute function. This is less modular. Maybe you could pack this code in a new impute_dataframe_range function.

Solve the issue #123 .

### DIFF
--- a/tsfresh/utilities/dataframe_functions.py
+++ b/tsfresh/utilities/dataframe_functions.py
@@ -30,9 +30,11 @@ def check_for_nans_in_columns(df, columns=None):
     if columns is None:
         columns = df.columns
 
-    for key in columns:
-        if df[key].isnull().any():
-            raise ValueError("Column {} of dataframe must not contain NaN values".format(key))
+    if pd.isnull(df.loc[:, columns]).any().any():
+        if not isinstance(columns, list):
+            columns = list(columns)
+        raise ValueError("Columns {} of dataframe must not contain NaN values".format(
+            df.loc[:, columns].columns[pd.isnull(df.loc[:, columns]).sum() > 0].tolist()))
 
 
 def impute(df_impute):
--- a/tsfresh/utilities/dataframe_functions.py
+++ b/tsfresh/utilities/dataframe_functions.py
@@ -33,7 +33,7 @@ def check_for_nans_in_columns(df, columns=None):
     if pd.isnull(df.loc[:, columns]).any().any():
         if not isinstance(columns, list):
             columns = list(columns)
-        raise ValueError("Columns {} of dataframe must not contain NaN values".format(
+        raise ValueError("Columns {} of DataFrame must not contain NaN values".format(
             df.loc[:, columns].columns[pd.isnull(df.loc[:, columns]).sum() > 0].tolist()))
 
 
--- a/tsfresh/utilities/dataframe_functions.py
+++ b/tsfresh/utilities/dataframe_functions.py
@@ -78,7 +78,7 @@ def impute(df_impute):
         df_impute.values[ indices ] = newValues
 
     # Ensure a type of "np.float64"
-    df_impute.astype(np.float64)
+    df_impute.astype(np.float64,copy=False)
 
 
 def impute_dataframe_zero(df_impute):
--- a/tsfresh/utilities/dataframe_functions.py
+++ b/tsfresh/utilities/dataframe_functions.py
@@ -86,9 +86,8 @@ def impute(df_impute):
 
 def impute_dataframe_zero(df_impute):
     """
-    Replaces all ``NaNs`` and ``infs`` from the DataFrame `df_impute` with 0s.
-
-    `df_impute` will be modified in place. All its columns will be of datatype ``np.float64``.
+    Replaces all ``NaNs``, ``-infs`` and ``+infs`` from the DataFrame `df_impute` with 0s.
+    The `df_impute` will be modified in place. All its columns will be into converted into dtype ``np.float64``.
 
     :param df_impute: DataFrame to impute
     :type df_impute: pandas.DataFrame
--- a/tsfresh/utilities/dataframe_functions.py
+++ b/tsfresh/utilities/dataframe_functions.py
@@ -10,7 +10,6 @@ import numpy as np
 import pandas as pd
 import logging
 
-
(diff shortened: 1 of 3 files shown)


tsfresh/utilities/dataframe_functions.py
```diff
@@ -55,7 +55,30 @@ def impute(df_impute):
...
+
+    # NaN -> median
+    columns = df_impute.columns
+    indices = np.nonzero(np.isnan(df_impute.values))
+    if len(indices[0]) > 0:
+        newValues = [ col_to_median[ columns[i] ] for i in indices[1]]
+        df_impute.values[ indices ] = newValues
+
+    # Ensure a type of "np.float64"
+    df_impute.astype(np.float64)
```
MaxBenChrist: you have to set `copy=False` here

MaxBenChrist: otherwise it will just return a copy

MaxBenChrist:
>Note: I did not impoved the performance of impute_dataframe_range(...) since it would have been to much of a hassle to implement all the checks in that function, e.g. in case the min/max/median values of each columns are not present.

I think the only check that is missing in your version is if there are no finite values at all

``` python

            if len(finite_values_in_column) == 0:
                _logger.warning(
                    "The replacement column {} did not have any finite values. Filling with zeros.".format(column_name))
                df_impute[column_name] = [0] * len(column)
                continue
```

F-A:
The check in case there no finite values in done in the "get_range_values_per_column" function. Line 173-175. But it does not log anything.

MaxBenChrist:
Ah yes you are right, couldn't you do something like

``` python
if len(masked.mask.sum(axis=0) == masked.data.shape[0]) > 0
    logging message")
```

F-A:
The log message was previously done in the impute function directly. In this function I do not create the masked array.
I could add the warning in the "get_range_values_per_column" where I have the masked array but this does make less sense...

jneuff:
@F-A Nice work!

The impute function is provided as a default for users who don't know which impute function to choose (we currently have two: impute_dataframe_range and impute_zero). From the semantics you're proposal is an improvement of the impute_dataframe_range function. So that's the function it should replace and impute should remain just a wrapper.

jneuff:
Ah, and while we're at it: I'd prefer the impute function to return the altered dataframe. Currently we're inconsistent about this.

MaxBenChrist:
> Ah, and while we're at it: I'd prefer the impute function to return the altered dataframe. Currently we're inconsistent about this.

The problem is that you create (temporary) copys of the dataframe if you impute that way. For big dataframes this could result in out of memory problems.

F-A:
Rewriting the  impute_dataframe_range with its current behavior would take me too much time: it assumes the max/min and median dict given in parameters can be incomplete, which results in a lot of column-wise checks. The impute function directly compute these dictionaries, so we know they are complete, which make the implementation easier.

As I mentioned earlier, I do not have much time to accord to tsfresh. I just wanted to test it on my data and I fixed this issue to do so. This PR correspond to my fix.
You can accept it if you want and improve the functions "impute_dataframe_range" and "impute_zero" in the same spirit later on. But this will be too much for me right now, sorry.

MaxBenChrist:
Thank you. no worries, we will take over from here. 

May I push some changes to your master branch to update this pr?

F-A:
Yes of course. I'll grant you access to it when I am in front of a computer.

On 6 Jan 2017 23:46, "Maximilian Christ" <notifications@github.com> wrote:

Thank you. no worries, we will take over from here.

May I push some changes to your master branch to update this pr?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/blue-yonder/tsfresh/pull/135#issuecomment-271031182>,
or mute the thread

MaxBenChrist:
> Yes of course. I'll grant you access to it when I am in front of a computer.

Perfect, let me know when you arranged it.

jneuff:
> The problem is that you create (temporary) copys of the dataframe if you impute that way. For big dataframes this could result in out of memory errors. I think we should do all imputing inplace

We could do it in place and return the dataframe. That way, we won't have to change the function if we later decide to make our data flow more explicit as some kind of pipeline.

@F-A No, worries. Thanks for contributing!

I just noticed, that we currently have quite some code duplication between `impute_dataframe_range` and `get_range_values_per_column`. We should simplify that anyway. This PR seems like a good opportunity.

F-A:
Ok I added both of you as collabortors for my fork. You should be able to push then.

MaxBenChrist:
Thanks, let's use this opportunity to refactor the module. I will have a look at it during the next days.

nils-braun:
@MaxBenChrist Is there news on this branch and PR? Or do you want to tell us your plans for refactoring - so we can help out? ;-)

MaxBenChrist:
I do not have a fixed plan yet but some aspects I want to tackle

- remove code duplication between impute_dataframe_range and get_range_values_per_column
- maybe remove impute_dataframe_zero
- check if we still need get_range_values_per_column if we use the masks
- ...

Maybe I will find some time during the next days to tackle this.

MaxBenChrist:
I finally found some hours to spend on this. I finished refactoring the dataframe_functions module.

However, I found that the usage of the impute functions in the transformers is not right. I will try to fix that as well

MaxBenChrist:
Well, I finished everything. If either @nils-braun or @jneuff gives his go, we can merge this thing.

jneuff:
Thanks @F-A and @MaxBenChrist !

MaxBenChrist:
Awesome! We are on the way to make tsfresh blazing fast ;)

F-A:
Nice refactoring! I'm glad my original investigations helped you :)

PR #138, opened 2017-01-07
renaming of settings classes and arguments for extract_features

MaxBenChrist:
I renamed `calculation_settings_mapping` to `para_map`. 

I do not like the long name, which result in a java like beasts such as  `default_calculation_settings_mapping` or `kind_to_calculation_settings_mapping`.

We should have a catchy, short (<10 character) name for the extraction settings. We are breaking our public API, so now is the time to change it.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -24,7 +24,7 @@ The following chapters will explain the tsfresh package in detail:
    Data Formats <text/data_formats>
    scikit-learn Transformers <text/sklearn_transformers>
    Feature Calculation <text/feature_calculation>
-   Feature Extraction Settings <text/feature_extraction_settings>
+   Feature Calculator Settings <text/feature_extraction_settings>
    Feature Filtering <text/feature_filtering>
    How to write custom Feature Calculators <text/how_to_add_custom_feature>
    Parallelization <text/parallelization>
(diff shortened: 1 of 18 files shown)


MaxBenChrist:
@earthgecko any ideas?

jneuff:
Again, I think `parameter` is the better choice.

MaxBenChrist:
Then we should come up with a neologism, something short and catchy 

Maybe fc_params? Short for featurecalculator_parameters

nils-braun:
How is this PR related to #138? Which one do you want to apply?

MaxBenChrist:
> How is this PR related to #138? Which one do you want to apply?

I dont get the question :-/. this pr is changing the naming of some variables in the feature_extraction_settings_refactoring branch

nils-braun:
Sorry, I put it in the wrong way :-)

I just wanted to make the relation to the other PR. If we rename the calculation settings object there into parameter map (or something else), we should also rename the variables and parameters as you have done it here.
Maybe we could merge those two PR together, because they need to be applied both (or none).
Just wanted to make this clear :-)

MaxBenChrist:
This pr here targets the feature_extraction_settings_refactoring branch.

I would do all changes on the feature_extraction_settings_refactoring branch and then later merge it all api changes, variable renaming and so on into master at the same time.

MaxBenChrist:
We will rename them to minimal, efficient and comprehensive settings.

MaxBenChrist:
- renamed `calculation_settings_mapping` to `fc_parameters`
- renamed setting types to `minimal`, `efficient` and `comprehensive`

PR #139, opened 2017-01-07
renamed extraction setting objects

MaxBenChrist:
To make them more catchy and to better reflect their behavior,  I picked `ExtendedParaMap`, `RecommendedParaMap` and `MinimalParaMap`


docs/text/feature_extraction_settings.rst
```diff
@@ -12,6 +12,216 @@ After digging deeper into your data, you maybe want to calculate more of a certa
...
+>>> from tsfresh.feature_extraction import ExtendedParaMap
+>>> settings = FeatureExtractionSettings()
+>>> # Set here the options of the settings object as shown in the chapters below
+>>> # ...
+>>> from tsfresh.feature_extraction import extract_features
+>>> extract_features(df, default_calculation_settings_mapping=settings)
+
+
+The
+feature calculators in this package).
```
jneuff: Trailing ')'

docs/text/feature_extraction_settings.rst
```diff
@@ -12,6 +12,216 @@ After digging deeper into your data, you maybe want to calculate more of a certa
...
+
+After digging deeper into your data, you maybe want to calculate more of a certain type of feature and less of another
+type. So, you need to use custom settings for the feature extractors. To do that with tsfresh you will have to use a
+custom settings object, like shown now:
+
```
jneuff: You repeat this paragraph a couple of times.

jneuff:
The names `Recommended-` and `Minimal-` are good. But in my opinion `Extended-` should rather be `Full-` or `Complete-`. Also, I think it would be easier to read and understand if they were called `-ParameterMap`.

MaxBenChrist:
Complete or full implies that we calculated all possible features, which is wrong. We could calculate infinitely many features by alternating the parameters. I would stick with extended.

Von meinem iPhone gesendet

docs/text/feature_extraction_settings.rst
MaxBenChrist: Just want to make sure that people get the message ;)

MaxBenChrist:
Before we merge this, we still have to redo the documentation. What do you think about the naming @nils-braun ?

nils-braun:
@MaxBenChrist I am with @jneuff for naming it `ParameterMap` instead of the shorter but more unreadable version.
Concerning the prefix, this is fine for me. I am a bit worried about `Recommended`, because who will want to use a not-recommended parameter map (which implies the naming)...

MaxBenChrist:
> @MaxBenChrist I am with @jneuff for naming it ParameterMap instead of the shorter but more unreadable version.
Concerning the prefix, this is fine for me. I am a bit worried about Recommended, because who will want to use a not-recommended parameter map (which implies the naming)...

I think most users should use the rommended settings. The extended settings are slower and only contain 2 features more at this point.

MaxBenChrist:
In most cases I checked, it slowed down the process significantly.

nils-braun:
Yes, I am with you. It's just about the naming. I am not sure if Recommend Settings is good, because it implies the other settings should never be used (because they are not recommended)

MaxBenChrist:
Maybe _efficient_ or _rational_ are better wordings instead of recommended? :D

MaxBenChrist:
I will merge this pr into #138 so we only have one pr

PR #140, opened 2017-01-11
Added sample ratio to driftbif simulation

kempa-liehr:
Also parameters were adapted and documentation improved. More tests were added.


MaxBenChrist:
Thanks @kempa-liehr

PR #143, opened 2017-01-12
New feature calculator estimating the largest fixed point of deterministic dynamics

kempa-liehr:


MaxBenChrist:
The unit tests are failiung 


```
self = <test_settings.TestSettingsObject testMethod=test_default_calculates_all_features>
    def test_default_calculates_all_features(self):
        """
            Test that by default a FeatureExtractionSettings object should be set up to calculate all features defined
            in tsfresh.feature_extraction.feature_calculators
            """
        settings = FeatureExtractionSettings()
        all_feature_calculators = [name for name, func in feature_calculators.__dict__.items()
                                   if hasattr(func, "fctype")]
    
        for calculator in all_feature_calculators:
            self.assertIn(calculator, settings.name_to_param,
                          msg='Default FeatureExtractionSettings object does not setup calculation of {}'
>                         .format(calculator))
E           AssertionError: Default FeatureExtractionSettings object does not setup calculation of max_fixed_point
```

because there are no extraction settings inside FeatureExtractionSettings for the feature calculator max_fixed_point

Can you move your default parameters from the feature calculator into the settings object?

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -1245,3 +1245,36 @@ def _phi(m):
         return np.sum(np.log(C)) / (N - m + 1.0)
 
     return np.abs(_phi(m) - _phi(m + 1))
+
+@set_property("fctype", "aggregate_with_parameters")
+@not_apply_to_raw_numbers
+def max_fixed_point(x, r=3, m=30):
```
MaxBenChrist: we do not have default values for the fc parameter

MaxBenChrist: Can you move them into the FeatureExtractionSettings object?

kempa-liehr: Yes, I'm going to.

MaxBenChrist:
Looks technically fine. I still have to take some time to think about what this feature is actually calculating :D 

Do have some ideas about the computational complexity? Maybe compared it to some other feature calculators?

tests/feature_extraction/test_extraction.py
```diff
@@ -201,3 +202,6 @@ def test_extract_features(self):
         self.assertTrue(np.all(extracted_features.b__abs_energy == np.array([36619, 35483])))
         self.assertTrue(np.all(extracted_features.b__mean == np.array([37.85, 34.75])))
         self.assertTrue(np.all(extracted_features.b__median == np.array([39.5, 28.0])))
+
+if __name__ == '__main__':
```
nils-braun: why is this needed for your feature? I would rather go with `pytest test_extraction.py` instead of calling the files by itself, because we set additional settings for pytest only (and not for unittest), which could in principle lead to strange things like the test is working here but not there...

tests/feature_extraction/test_feature_calculations.py
```diff
@@ -505,3 +507,26 @@ def test_approximate_entropy(self):
...
+        v0 = max_fixed_point(v[:,0], **default_params)
+        self.assertTrue(abs(ds.deterministic-v0)<0.0001)
+
+        # Brownian motion
+        ds = velocity(tau=2.0/0.3-3.8, delta_t=0.05, R=3e-4, seed=0)
+        v = ds.simulate(1000000, v0=np.zeros(1))
+        v0 = max_fixed_point(v[:,0], **default_params)
+        self.assertTrue(v0<0.001)
+
+if __name__ == '__main__':
```
nils-braun: same thing

kempa-liehr:
time series with 1000000 samples:

> 10 loops, best of 3: 136 ms per loop

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -1245,3 +1245,43 @@ def _phi(m):
...
+        return x.mean()
+    quantiles = df.groupby('quantiles')
+    
+    result = pd.DataFrame({'x_mean': quantiles.signal.mean(),
+                           'y_mean': quantiles.delta.mean()
+    })
+    
+    try:
+        coeff = np.polyfit(result.x_mean, result.y_mean, deg=m)
+    except:
```
nils-braun: is it possible to name the possible exceptions here? Because this would make it possible for the user to see, if the triggered exception was "intended" or not.

tests/feature_extraction/test_feature_calculations.py
```diff
@@ -505,3 +507,26 @@ def test_approximate_entropy(self):
...
         self.assertAlmostEqualOnAllArrayTypes(approximate_entropy, [12, 13, 15, 16, 17]*10, 0.282456191, m=2, r=0.9)
         self.assertRaises(ValueError, approximate_entropy, x=[12, 13, 15, 16, 17]*10, m=2, r=-0.5)
+
+        
+    def test_max_fixed_point(self):
+        """
+        Estimating the intrinsic velocity of a dissipative soliton
+        """
+        settings = FeatureExtractionSettings()
+        default_params = settings.name_to_param['max_fixed_point'][0]
```
nils-braun: maybe better just explicitely set the parameters here. Because otherwise it is a but unrelated, that changing the settings object leads to a failing test of a feature calculator... :-)

tsfresh/feature_extraction/feature_calculators.py
kempa-liehr: Dynamic systems have stable and unstable fixed points. Without noise these systems would be driven to one of their stable fixed points and stay there for ever. So, fixed points resemble an equilibrium state. 

Noise pushes the system out of their equilibrium and the system's dynamics try to bring it back. So, by averaging the change of state we are able to model the deterministic dynamics as polynomial. Its roots are the fixed points of the dynamics. We are taking the largest fixed point. In the dissipative soliton example, this is an estimate for the equilibrium velocity of the dissipative soliton.

driftbif20170113a_active_brownian_motion
driftbif20170113a_brownian_motion

tests/feature_extraction/test_extraction.py
kempa-liehr: Good point, I was not aware of that. I am going to revert these commits.

tests/feature_extraction/test_feature_calculations.py
kempa-liehr: Copy that.

kempa-liehr:
Dynamic systems have stable and unstable fixed points. Without noise these systems would be driven to one of their stable fixed points and stay there for ever. So, fixed points resemble an equilibrium state.

Noise pushes the system out of their equilibrium and the system's dynamics try to bring it back. So, by averaging the change of state we are able to model the deterministic dynamics as polynomial. Its roots are the fixed points of the dynamics. We are taking the largest fixed point. In the dissipative soliton example, this is an estimate for the equilibrium velocity of the dissipative soliton.

driftbif20170113a_active_brownian_motion
driftbif20170113a_brownian_motion

tsfresh/feature_extraction/feature_calculators.py
kempa-liehr: Yes, see new commit.

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -1245,3 +1245,43 @@ def _phi(m):
...
+    result = pd.DataFrame({'x_mean': quantiles.signal.mean(),
+                           'y_mean': quantiles.delta.mean()
+    })
+    
+    try:
+        coeff = np.polyfit(result.x_mean, result.y_mean, deg=m)
+    except (np.linalg.LinAlgError, ValueError):
+        return 0.
+    
+    return np.max(np.real(np.roots(coeff)))
```
MaxBenChrist: Wouldnt it make more sense to return all all coefficients? You are throwing features away at this point aren't you?

We could make it an apply feature that returns multiple coefficients.

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -1245,3 +1245,43 @@ def _phi(m):
...
+@set_property("fctype", "aggregate_with_parameters")
+def max_fixed_point(x, r, m):
+    """
+    Largest fixed point of dynamics estimated from polynomial, which has been fitted to 
+    the deterministic dynamics of Langevin model as described by
+
+        Friedrich et al. (2000): Physics Letters A 271, p. 217-222
+        *Extracting model equations from experimental data*
+
+    For short time-series this method is highly dependent on the parameters.
```
MaxBenChrist: I think I unterstand the feature. It fits a polynomial over quantile-binned changes of the time series. It would be nice to have a formula of that in the docstring

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -1245,3 +1245,43 @@ def _phi(m):
         return np.sum(np.log(C)) / (N - m + 1.0)
 
     return np.abs(_phi(m) - _phi(m + 1))
+
+@set_property("fctype", "aggregate_with_parameters")
+def max_fixed_point(x, r, m):
```
MaxBenChrist: lets rename the feature calculator, what do you think about `max_langevin_fixed_point`

MaxBenChrist:
By the way, great plots!

kempa-liehr:
Thanks for the positive feedback and the suggestions!
* I'm going to implement the feature calculator as apply feature and return the coefficients of the polynomial.
* I'm putting the Langevin model into the doctoring.
* I suggest to rename the feature calculator to =langevin_coefficients=, or even better to =Friedrich_coefficients= in commemoration of Prof. Friedrich

tsfresh/feature_extraction/feature_calculators.py
earthgecko: @kempa-liehr this looks like another very useful, description feature, thanks Andreas :)

MaxBenChrist:
I am fine with merging this. 

I checked the performance, runtime is okay.

PR #152, opened 2017-02-13
Added progressbar for feature selection.

moritzgelb:


tsfresh/feature_selection/feature_selector.py
```diff
@@ -116,8 +117,12 @@ def check_fs_sig_bh(X, y, settings=None):
 
     # Helper function which wraps the _calculate_p_value with many arguments already set
     f = partial(_calculate_p_value, y=y, settings=settings, target_is_binary=target_is_binary)
-    results = pool.map(f, [X[feature] for feature in df_features['Feature']], chunksize=settings.chunksize)
-    p_values_of_features = pd.DataFrame(results)
+
+    total_number_of_features = len(X.columns)
+    results = tqdm(pool.imap_unordered(f, [X[feature] for feature in df_features['Feature']]),
```
nils-braun: What happened to the chunk size setting?

MaxBenChrist:
Hey @moritzgelb, Nice to have you on board as well. :)

tsfresh/feature_selection/feature_selector.py
moritzgelb: Can we use the same function ```_calculate_best_chunksize``` from ```extraction.py``` to calculate the chunksize? If yes, it may be worth to move it to utililties.

nils-braun: Yes, I would suggest to do so

tsfresh/feature_selection/feature_selector.py
```diff
@@ -118,8 +119,10 @@ def check_fs_sig_bh(X, y, settings=None):
     # Helper function which wraps the _calculate_p_value with many arguments already set
     f = partial(_calculate_p_value, y=y, settings=settings, target_is_binary=target_is_binary)
 
+    chunksize = helper_functions.calculate_best_chunksize(X, settings)
```
nils-braun: Is the length of X the same as the length of the features list? Anyhow, I think it is easier readable to have the features list here directly

moritzgelb: Yeah, you're right...

Issue #147, opened 2017-01-24
Warning while feature selection


bobchennan:
No handlers could be found for logger "tsfresh.feature_selection.feature_selector"

Does it matters?

MaxBenChrist:
It is fine, you have to [configure a logger](http://excid3.com/blog/no-handlers-could-be-found-for-logger)

MaxBenChrist:
Maybe we can consider to call

```
import logging
logging.basicConfig()
```

somewhere in our scripts if the user does not specify a logging handler.

bobchennan:
Yes, since it is not mentioned in the document.

MaxBenChrist:
It is in the master, thanks for reporting it @bobchennan

PR #154, opened 2017-02-15
Issue #147

moritzgelb:
Fix for issue #147 like recommended [here](http://docs.python-guide.org/en/latest/writing/logging/).

### DIFF
--- a/tsfresh/__init__.py
+++ b/tsfresh/__init__.py
@@ -18,6 +18,18 @@ except:
     __version__ = 'unknown'
 
 
+# Set default logging handler to avoid "No handler found" warnings.
+import logging
+try:  # Python 2.7+
+    from logging import NullHandler
+except ImportError:
+    class NullHandler(logging.Handler):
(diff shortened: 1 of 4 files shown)


nils-braun:
Why did you commit also the other things not related to this issue?

nils-braun:
Otherwise I am fine with this

MaxBenChrist:
Hi Moritz, 

if I got your changes right, under your changes the custom behavior is that if the user is not specifying an logging handler, tsfresh drops all logs... am I right?

nils-braun:
I do think this is reasonable for the non error logs

moritzgelb:
@nils-braun the first four commits were a mistake... I'll make a new pull request. 
@MaxBenChrist yes. The user can then specify the log level by something like: ```import logging
logging.basicConfig(level=logging.INFO)```

nils-braun:
You could just have reverted the commits (because the branch will be squanshed anyways) - but never mind.

PR #155, opened 2017-02-15
Fix for issue 147 - no handler warning

moritzgelb:
New pull request for issue #147. 
 The user can then specify the log level by something like: 
```import logging```
```logging.basicConfig(level=logging.INFO)```

### DIFF
--- a/tsfresh/__init__.py
+++ b/tsfresh/__init__.py
@@ -18,6 +18,18 @@ except:
     __version__ = 'unknown'
 
 
+# Set default logging handler to avoid "No handler found" warnings.
+import logging
+try:  # Python 2.7+
+    from logging import NullHandler
+except ImportError:
+    class NullHandler(logging.Handler):
+        def emit(self, record):
+            pass
+
+logging.getLogger(__name__).addHandler(NullHandler())
+
+
 from tsfresh.convenience.relevant_extraction import extract_relevant_features
 from tsfresh.feature_extraction import extract_features
 from tsfresh.feature_selection import select_features
--- a/tsfresh/feature_extraction/extraction.py
+++ b/tsfresh/feature_extraction/extraction.py
@@ -83,6 +83,9 @@ def extract_features(timeseries_container, feature_extraction_settings=None,
     :return: The (maybe imputed) DataFrame containing extracted features.
     :rtype: pandas.DataFrame
     """
+    import logging
+    logging.basicConfig()
+    
     # Always use the standardized way of storing the data.
     # See the function normalize_input_to_internal_representation for more information.
     kind_to_df_map, column_id, column_value = \

### CODE REVIEW

**Comment by MaxBenChrist** [comment]
What do you think @jneuff @nils-braun , I think it is dangerous that the default behavior is to drop the messages.

File: tsfresh/__init__.py [hunk 1]
```diff
@@ -18,6 +18,18 @@
     __version__ = 'unknown'
 
 
+# Set default logging handler to avoid "No handler found" warnings.
+import logging
+try:  # Python 2.7+
+    from logging import NullHandler
```
MaxBenChrist: Can we have one console print that emits a message that no handler was found? So users are aware that they should specify a logger

nils-braun:
This would then exactly be the reason, why issue #147  was started....

tsfresh/__init__.py
nils-braun: https://github.com/blue-yonder/tsfresh/issues/147

MaxBenChrist: I disagree, Nils. From my perspective, the motivaiton behind #147 is that tsfresh was, due to the multiprocessing, spamming the command lines with no logger found messages in the line of 

```
No handlers could be found for logger "tsfresh.feature_selection.feature_selector"
No handlers could be found for logger "tsfresh.feature_selection.feature_selector"
No handlers could be found for logger "tsfresh.feature_selection.feature_selector"
No handlers could be found for logger "tsfresh.feature_selection.feature_selector"
...
```

I am fine with a singular message that no logger was found. However, having 10s or 100s of these messages is quite annoying. Now, if I get the changes right, we have no message at all. I think not saying anything about a missing logging backend is dangerous.

nils-braun: ok, I understand. How would you store the information, that the logging message was already shown?

MaxBenChrist: I was thinking about setting a void logger (something that just writes all logging messages to /dev/null) somewhere in the top level in extract_features if no logging backend is specified (when setting the void logger one can print a short warning "hey we found no logger, going to drop all messages from here on"). However, my knowledge of the logging package is limited.

moritzgelb: This seem a bit heavy to me. 
Can we not just write out a short warning and make a proper documentation of how to use the logging? Also I would then add a logging config to all examples. So, if the user starts from scratch by looking at the examples he will know what to do.

jneuff: I agree with @MaxBenChrist that the problem are these multiple no-handlers-found messages.

nils-braun:
I am a bit against not doing, what the logging documentation says about logging in a library (https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library) - and checking for the list of defined handlers seems a but too "low level digging into the logging library" for me.
The "problem" is a bit, that we are library and application developer in the same time. So, after the logging documentation and various blogs I have found in the internet, we should do the following:

In the library code part (this is, e.g. in feature_selector.py), we should go with the NullHandler, as described here (https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library) and as @moritzgelb implemented it. This behaviour is needed for python versions below 3.2 (read the paragraph above the page I have quoted), but we are using all python versions, so we will need it.
In the application part (e.g. the convenience functions, that most people will use and maybe also the transformers), we should call
```
import logging
logging.basicConfig()
```
once in the beginning. This will not override the settings of the user, if he has defined one.

Also, reading through https://fangpenlin.com/posts/2012/08/26/good-logging-practice-in-python/, we should probably not get the logger on module level, but at a point we will need it.

MaxBenChrist:
Great analysis @nils-braun. I never thought about that we are an applications and a library at the same time. but makes perfect sense. 

So, I think that your proposal is a great idea: setting the basic config in the convenience function and use the null Handler in the library.

moritzgelb:
Thanks @nils-braun  for pointing this out. I added the two lines to ```extract_features```.

MaxBenChrist:
great job @moritzgelb and @nils-braun

PR #160, opened 2017-03-01
Fix issue 159

MaxBenChrist:


MaxBenChrist:
After merging, we should bump up to 0.5.1

tsfresh/transformers/relevant_feature_augmenter.py
```diff
@@ -212,19 +212,24 @@ def transform(self, X):
...
+                                                        column_kind=self.feature_extractor.column_kind,
+                                                        column_value=self.feature_extractor.column_value)
 
-        X_augmented = relevant_feature_extractor.transform(X)
+        feature_augmenter_restricted.set_timeseries_container(self.feature_extractor.timeseries_container)
 
-        return X_augmented.copy().loc[:, self.feature_selector.relevant_features]
+        if self.evaluate_only_added_features:
+            X_tsfresh = feature_augmenter_restricted.transform(X).loc[:, self.feature_selector.relevant_features]
+            return pd.concat([X_tsfresh, X], axis=1)
```
jneuff: So the bug was that we forgot to merge in `X` in case of evaluate_only_added_features?

This transformer got really complicated. Maybe it'd be easier to understand if we implemented it in terms of `extract_features` and `select_features`?

tests/transformers/test_relevant_feature_extractor.py
```diff
@@ -67,25 +68,44 @@ def test_impute_works(self):
...
+
+        y = pd.Series({10: 1, 500: 0})
+        X = pd.DataFrame(index=[10, 500])
+        X["pre_feature"] = 0
+
+        augmenter.set_timeseries_container(self.test_df)
+        augmenter.fit(X, y)
+        transformed_X = augmenter.transform(X.copy())
+
+        self.assertEqual(sum(["pre_feature" == x for x in transformed_X.columns]), 1)
```
MaxBenChrist: I want to make sure that is contained exactly one time. 

I want to prevent bugs where the non-time series feature is duplicated.

If you come up with a more elegant way to check that, I would be happy to hear it

tsfresh/transformers/relevant_feature_augmenter.py
MaxBenChrist: The functionality to add non-time series features does neither fit the `extract_features` (which just extract the time series features) nor the `select_features` method (which just filters all features, no madder if they originate from time series or are externally given).

In my opinion, the `relevant_feature_augmenter` is the right place for that functionality.

Regarding the complexity of the `relevant_feature_augmenter`, If you come up with a more elegant way, I am happy to hear about that.

tests/transformers/test_relevant_feature_extractor.py
jneuff: I am not sure how to keep it simple and make it obvious. But you could add a comment explaining what you assert here.

tsfresh/transformers/relevant_feature_augmenter.py
jneuff: We don't have to tackle this now, but what I meant is this: We'd get more clarity implementing `RelevantFeatureAugmenter` using `extract_features` and `select_features` instead of `FeatureAugmenter` and `FeatureSelector`.

tests/transformers/test_relevant_feature_extractor.py
jneuff:
```python
self.assertEqual(sum(["pre_feature" == column for column in transformed_X.columns]), 1)
```
would also help.

MaxBenChrist: Good idea!

Issue #86, opened 2016-11-30
Rolling window of time series


xgdgsc:
I want to extract features from a rolling window of a table with columns of several timeserieses and do some prediction based on the timeseries in that window.
Currently, as far as I understand the [doc](http://tsfresh.readthedocs.io/en/latest/text/quick_start.html). I have to extract the timeseries and tile them like in the example, so there would be a lot of duplicate data because the rolling window and doesn' t seem memory efficient. Is there a rolling window API or better ways to do it? 

Thanks!

MaxBenChrist:
we have not yet implemented a rolling window api to do that efficiently.

To implement such a API, one would have to decide for every feature calculator if this calculator can use the result of the last window for the current window. For some features this can be easily done (maximum, mean, ..) but for others this is not trivially possible (median, wavelet coefficients, ...)

What we could do is to pick some feature calculator for which this could be possible and then implement such an API.

ClimbsRocks:
This is pretty critical for all the problems I'm working on as well.

MaxBenChrist:
For most use cases that involve to forecast time series this can reduce the time to calculate the features. 

But as stated above, for a class of features is it mathematically impossible to use auxiliary results from the last window.

MaxBenChrist:
If somebody of you wants to implement this, I would be glad to help you with the design decisions.  I will probably not have time for this during the next months

MaxBenChrist:
Maybe we should provide a wrapper for the translation of timestamp, value combinations into rolling window time series. 

It is pretty straightforward to implement with http://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.shift.html applied to a groupby. (http://pandas.pydata.org/pandas-docs/version/0.17.0/generated/pandas.core.groupby.DataFrameGroupBy.shift.html)

Afterwards one has to drop NaNs.

MaxBenChrist:
Lets tackle this, the following code will do the treansformation:

```
import numpy as np
import pandas as pd

cid = np.repeat([10, 500]*2, 4)
csort = [1,2,3,4]*4
 cval = [11, 9, 67, 45] * 4
ckind= np.repeat(["a", "b"], 8)
df = pd.DataFrame({"id": cid, "sort": csort, "val": cval, 'kind': ckind})

n = max(cid)

lst_df = []

for i in range(n):
    df_temp = df.groupby(["id", "kind"]).shift(-i)
[...]

PR #170, opened 2017-03-18
Rolling timeseries

nils-braun:
This implements a "rolling" parameter to the extract function. Should solve #86 and alike.

I am still writing some documentation and an example notebook, but this is how the code will look like.


nils-braun:
... and I will fix the merge conflicts soon...

nils-braun:
I have now added documentation to the code and to the read-the-docs part. Do you think it is needed to also include an example jupyter notebook?

docs/text/rolling.rst
```diff
@@ -0,0 +1,160 @@
...
+
+How to handle rolling time series
+=================================
+
+In many application with time series on real-world problems, the "time" column
+(we will call it time in the following, although it can be anything)
+gives a certain sequential order to the data. We can exploit this sequence to generate
+more input data out of single time series, by *rolling* over the data.
+
+Imagine the following situation: you have the data of EEG measurements, that
```
MaxBenChrist: this example confused me a little bit. 

E.g. what do you mean by "if you have stopped the measurement"?

MaxBenChrist: Maybe we can say something in the lines of: the rolling mechanism takes a time series "x" and creates n new time series x^k, each a data point of "x", lets say "x_k"and "x_{k-1}, x_{k-2}..."

I think it would be benificial to add some formulas, what do you think?

So 

rolling_mechanism(x) gives list of time series [x^1,x^2, x^3, ..] where

x^k = (x_k, x_{k-1}, x_{k-2}, ...)

tsfresh/utilities/dataframe_functions.py
```diff
@@ -322,4 +326,26 @@ def normalize_input_to_internal_representation(df_or_dict, column_id, column_sor
         if kind_to_df_map[kind][column_value].isnull().any():
             raise ValueError("You have NaN values in your value column.")
 
+    # Roll the data frames if requested
+    if rolling:
```
MaxBenChrist: we kind of assume that the time stamps are equal. 

If not, this is not returning what the users may think.

MaxBenChrist:
I am not sure about one architectural decision that you made: You included the shift from time series in `normalize_input_to_internal_representation`. Another way would be to have it in the utilities module, which I would prefer actually.

There are several arguments for both options: 

Nils way (in `normalize_input_to_internal_representation`):
good: we can later optimize the calculation of features for rolled time series
good: we could have a memory optimized representation of the rolled time series
bad: transformation is hidden from user, debugging hard for the user (have I selected the right shifting mechanism)

Max way (as external script in `dataframe_utilities`):
good: better separation of functionality
good: user can play with shifted time series and save the DataFrame with the the shifted time series 

I want to keep the "core" of tsfresh as small as possible. 

I think we will still have to make a few additions to handle cases where the time series have different2 time stamps and i want to keep that code out of the core.

What do you think guys?

tsfresh/utilities/dataframe_functions.py
MaxBenChrist: I am not yet sure how to properly handle those cases

nils-braun:
I see your point Max. Just two remarks (although I am also not sure what is better here): what made me take this descission was the fact, that we had so many issues concerning rolling TS that I thought we want to make this a core functionality. The second thought was, that if I implement this in the normalization, I would only have to roll one kind of data format and not all three of them...

tsfresh/utilities/dataframe_functions.py
nils-braun: We would need specific knowledge on the data domain to handle this properly...

docs/text/rolling.rst
nils-braun: Jep, sounds good

tsfresh/utilities/dataframe_functions.py
MaxBenChrist: yes. For the time being we can just add a check or throw an error if the time stamps are not sampled uniformly.

MaxBenChrist:
> we had so many issues concerning rolling TS that I thought we want to make this a core functionality.

it is still in the tsfresh package, just not in the extract_features.

> I implement this in the normalization, I would only have to roll one kind of data format and not all three of them...

Your point about only having to implement it once for all three formats is good. So far, we have it for a stacked DataFrame. I don't think that it would be much work to apply it to a flat DataFrame. (we could reduce the flat DataFrame to a index column, do the rolling like it was a stacked one and then just get the values from the different time series columns, should be only 3-4 lines more + unit tests )

jneuff:
Nice work @nils-braun! As @MaxBenChrist said, we could also move the rolling functionality into a different function. But I'm fine with both. For the non-uniformly sampled cases a warning is enough, I'd say.

kempa-liehr:
@nils-braun: Thanks for implementing the rolling functionality.

I agree with @MaxBenChrist and @jneuff  and support refactoring the rolling functionality into a separate function.

nils-braun:
Ok, this is how it would look like with a separate function.

nils-braun:
So which one do you like more? The separate implementation or the incorporated one? 
Either way, I think we should finish this up now.

MaxBenChrist:
wow Nils. looks really good. 

I will polish the documentary and then we can merge this.

MaxBenChrist:
I will open another pr for the refactoring.

PR #171, opened 2017-03-25
fitting of polynomial on NaNs does not make sense

MaxBenChrist:
this will fix issue 169

@kempa-liehr : regarding the polynomial fit on signal and delta , is there another way that comes to your mind then dropping nans?


MaxBenChrist:
@nils-braun good point: we should trigger a warning

MaxBenChrist:
@kempa-liehr : 

I want to have a release today due to the index bug. Would make sense to include this one as well.

Issue #165, opened 2017-03-14
Merge fail when Index has same Name as ID column


MaxBenChrist:
I was stumbling over a Bug when the Index of the pandas.DataFrame time series container had the same name as the ID column.

I will try to reproduce the bug and to find a minimal example.

MaxBenChrist:
I reproduced it by adding the following cell to the robot_failure_notebook

```
import random
import string 

idcol = df.id
df = df.drop("id", axis=1)   
df["id"] = [random.choice(string.ascii_lowercase) for _ in range(df.shape[0])]
df = df.set_index("id")
df["id"] = idcol.values
```

PR #172, opened 2017-03-25
we do not need index for internal representation

MaxBenChrist:
fixes issue #165


MaxBenChrist:
however, we need the index for the merging of results.

nils-braun:
Is dropping the name really the best way to do this?

MaxBenChrist:
The index that we later return in the feature matrix is based on the id column and not the index of the time series container data frame that is used as input. so yes, I think dropping the name should be fine.

Issue #175, opened 2017-03-29
problem in calculating index_mass


MoffyJiang:
Hi, our company is trying to apply tsfresh on our credit scoring models for time series data but one problems came up.  During calculating index_mass, np.argmax(mass_centralized >= config["q"])+1) is trying to get the index of a time series for a single sample when the cumsum of this feature is greater than given quantile. however, the results is not a 'relative' index within a time series data for a single sample but relative to the index of the whole sample. 
For example, in the show case robot_execution_failures, extracted feature 'a__index_mass_quantile__q_0.1' increases when sample id increases. In my experience, using a pure random id to classify samples always has good statistical evaluation results but the models themselves are meaningless. 
We are thinking that is it a better way to reset dataframe index while calculating index mass? So that  the np.argmax will get the the index regardless of which part of the whole sample is the sample from  but just the position of the time in the series.

MaxBenChrist:
Hi @MoffyJiang 

may I ask why you closed this issue?

I think the point that you mentioned is valid. I will investigate this.

MoffyJiang:
Hi sorry I saw #162 and I thought it was the same issue as mine. But after I read it carefully I found that was about id not index....sorry

MaxBenChrist:
I now how to fix this. We can just call `.value` before `np.argmax`

Thanks for reporting, I will add a pr later.

MoffyJiang:
ah yes! thanks for your help!

MaxBenChrist:
#162 was about the id column but the behavior is caused by the index

MaxBenChrist:
Please check if your problems also occur with version 0.7.0, if they do reopen this issue. 

again, thanks for reporting this one

MoffyJiang:
thanks for your help!

PR #176, opened 2017-03-29
argmax uses global index

nils-braun:
This should solve #175 by adding .value (as Max said in #175). I have also added a test to not let this happen again.


MaxBenChrist:
haha Nils, the man who submits pr faster than his shadow. ;)

There are also another few occurrences of argmax, we should check them. Further 

https://github.com/blue-yonder/tsfresh/blob/master/tests/feature_extraction/test_extraction.py#L114-LL16

nils-braun:
@MaxBenChrist In the other two cases of calling np.argmax, we use np.asarray before, which should have more or less the same effect as using `.value`.

MaxBenChrist:
> @MaxBenChrist In the other two cases of calling np.argmax, we use np.asarray before, which should have more or less the same effect as using .value.

You are right. 

Can you add another drop index to https://github.com/blue-yonder/tsfresh/blob/master/tests/feature_extraction/test_extraction.py#L98 ?

tests/feature_extraction/test_feature_calculations.py
```diff
@@ -320,6 +320,16 @@ def test_mass_quantile(self):
         six.assertCountEqual(self, list(res.index), expected_index)
         self.assertAlmostEqual(res["TEST__index_mass_quantile__q_0.5"], 0.5, places=1)
 
+        # Test for parts of pandas series
```
MaxBenChrist: ?

tests/transformers/test_full_pipeline.py
```diff
@@ -37,14 +37,6 @@ def test_relevant_extraction(self):
                                   'a__autocorrelation__lag_1',
                                   'a__binned_entropy__max_bins_10',
                                   'a__count_above_mean',
-                                  'a__index_mass_quantile__q_0.1',
```
MaxBenChrist: Why did you remove them?

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -788,11 +788,13 @@ def index_mass_quantile(x, c, param):
...
     res = pd.Series()
 
+    x = pd.Series(x)
+
     if s == 0: # all values in x are zero or it has length 0
         for config in param:
             res["{}__index_mass_quantile__q_{}".format(c, config["q"])] = np.NaN
     else: # at least one value is not zero
-        mass_centralized = np.cumsum(np.abs(x)) * 1.0 / s
+        mass_centralized = (np.cumsum(np.abs(x)) * 1.0 / s).values
```
MaxBenChrist: I would add the `.value`to the x inside the abs call. I think it is more readable

tests/transformers/test_full_pipeline.py
nils-braun: As I have said in the it message, they are not relevant anymore

tests/feature_extraction/test_feature_calculations.py
nils-braun: Well, the issue was not spotted because we calculate the feature always on the full series

MaxBenChrist: Why do you need a vector with over 150 values? are 3 not enough?

nils-braun: I just copied the example from above and added some values. It does not matter how much.

MaxBenChrist: I was thinking about keeping the runtime of the tests low

PR #181, opened 2017-03-29
moved imputing out of the loop

MaxBenChrist:
This would be the easiest solution to move the imputing out of the loops where the id chunks are aggregated.

What do you think?


nils-braun:
I thought there would be a reason to have this in the code before the concationating, because we would impute too much if we do this afterwards - but I can not remember the exact reason. I just remember that we did have the discussion before. Is your new solution faster than the old one or why did you do this?

MaxBenChrist:
My solution should be slower. 

I think the motivation was that when you impute on chunks you can start imputing before having the whole column.

MaxBenChrist:
So, @nils-braun @jneuff, can we merge this?

nils-braun:
Sorry to ask again, but: why do we need this at all?

MaxBenChrist:
I dont understand your question. Are you referring to the imputing?

nils-braun:
No, I just wanted to know why we need your changes at all :-) Because if it does not give more features and make things slower... ;-)

Issue #177, opened 2017-03-29
Add maximum_number_of_timeshifts parameter to roll_time_series


MaxBenChrist:
Add maximum_number_of_timeshifts parameter to roll_time_series method instead of calculating the max lag

https://github.com/blue-yonder/tsfresh/blob/master/tsfresh/utilities/dataframe_functions.py#L414-L4120

I fear the current version explodes on long time series

PR #184, opened 2017-04-01
Maximum rolling

nils-braun:
Solves #177

### DIFF
--- a/tests/utilities/test_dataframe_functions.py
+++ b/tests/utilities/test_dataframe_functions.py
@@ -213,38 +213,53 @@ class RollingTestCase(TestCase):
 
         df_full = pd.concat([first_class, second_class], ignore_index=True)
 
-        df = dataframe_functions.roll_time_series(df_full, column_id="id", column_sort="time",
-                                                  column_kind=None, rolling_direction=1)
-
         correct_indices = (["id=1, shift=3"] * 1 +
                            ["id=1, shift=2"] * 2 +
                            ["id=1, shift=1"] * 3 +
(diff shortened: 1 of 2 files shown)


tsfresh/utilities/dataframe_functions.py
```diff
@@ -360,6 +361,9 @@ def roll_time_series(df_or_dict, column_id, column_sort, column_kind, rolling_di
     :type column_kind: basestring or None
     :param rolling_direction: The sign decides, if to roll backwards or forwards in "time"
     :type rolling_direction: int
+    :param maximum_number_of_timeshifts: If None, shift as often as possible. If not None, shift only
```
MaxBenChrist: isn't "If not None, shift only up to maximum_number_of_timeshifts" more precise?

tests/utilities/test_dataframe_functions.py
```diff
@@ -230,6 +230,22 @@ def test_positive_rolling(self):
         self.assertListEqual(list(df["b"].values),
                              [5, 5, 6, 5, 6, 7, 12, 5, 6, 7, 8, 12, 13])
 
+        df = dataframe_functions.roll_time_series(df_full, column_id="id", column_sort="time",
```
MaxBenChrist: i think we should add more test (e.g. in the other direction and with maximum_number_of_timeshifts>1)

tsfresh/utilities/dataframe_functions.py
nils-braun: Let's use both :-)

PR #186, opened 2017-04-02
Make extraction faster

nils-braun:
This one was some major work, so let me describe everything in detail:
I have done some profiling to increase the runtime performance of the feature extraction. 

I have done this using two tools (as reference for later): cProfile + pyprof2calltree for vizualisation and kernprof (from line_profiling), e.g. with

    python -m cProfile -o prof.out tests/examples/test_robot_execution_failures.py
    pyprof2calltree -i prof.out -k
    kernprof -l -v tests/examples/test_robot_execution_failures.py

Everything was done on our default robots failures dataset (I know that this may not be the perfect example for out usage, but I need to use something that is standardized).

After identifying the main runtime issues, I have refactored them step by step. I have made sure, that the results stayed the same during all steps using the full robots dataset and our unit tests.

Here are the results:

| What | Before | After |
| ------- | --------- | ------- |
| extraction on full robot set (without parallelization) | 39.51s | 20.68s ( - 48% !!!) |
| run all unittest | 41.91s | 38.73s | 

Parallelisation was turned off for the first measurement, to not rely on kernel process management etc.

The changes to the functions were in detail (all in s for the first 20 ids of the robot set):

| Name | Before | After |
| ---- | ---- | ---- |
| large_standard_dev | 0.33 | 0.076 |
| symmetry_looking | 0.758 | 0.046 |
| number_peaks | 3.901 | 0.160 |
| index_mass_quantile | 0.926 | 0.039 |
| cwt_coeff | 1.813 | 0.129 |
| ar_coeff | 0.626 | 0.1629 |

After that, we only have 4 functions left, that have "large" performance issues:
* augmented_dicker_fuller 14 %
* mean_abs_change_qunatiles 25 %
* autocorrelation 14 %
* quantile 6 %

The percentage numbers are the percentage they use up during the full extraction function.

TLDR: I increased the runtime by approx 50% on our test dataset while remaining the content of the calculations.

### DIFF
--- a/tsfresh/feature_extraction/extraction.py
+++ b/tsfresh/feature_extraction/extraction.py
@@ -120,7 +120,8 @@ def extract_features(timeseries_container, feature_extraction_settings=None,
         result = _extract_features_parallel_per_sample(kind_to_df_map, feature_extraction_settings,
                                                        column_id, column_value)
     elif parallelization == 'no_parallelization':
-        result = _extract_features_serial(kind_to_df_map, feature_extraction_settings, column_id, column_value)
+        result = _extract_features_parallel_per_kind(kind_to_df_map, feature_extraction_settings,
+                                                     column_id, column_value, serial=True)
     else:
         raise ValueError("Argument parallelization must be one of: 'per_kind', 'per_sample'")
 
@@ -132,7 +133,7 @@ def extract_features(timeseries_container, feature_extraction_settings=None,
     return result
 
 
-def _extract_features_parallel_per_kind(kind_to_df_map, settings, column_id, column_value):
+def _extract_features_parallel_per_kind(kind_to_df_map, settings, column_id, column_value, serial=False):
     """
     Parallelize the feature extraction per kind.
 
@@ -147,6 +148,9 @@ def _extract_features_parallel_per_kind(kind_to_df_map, settings, column_id, col
     :param settings: settings object that controls which features are calculated
     :type settings: tsfresh.feature_extraction.settings.FeatureExtractionSettings
 
+    :param serial: Do not parallelize the extraction. This can be handy if (1) you want to debug something
+       (2) you want to profile something or (3) your environment does not support multiprocessing
+
     :return: The (maybe imputed) DataFrame containing extracted features.
     :rtype: pandas.DataFrame
     """
@@ -159,8 +163,14 @@ def _extract_features_parallel_per_kind(kind_to_df_map, settings, column_id, col
     chunksize = helper_functions.calculate_best_chunksize(kind_to_df_map, settings)
 
     total_number_of_expected_results = len(kind_to_df_map)
-    extracted_features = tqdm(pool.imap_unordered(partial_extract_features_for_one_time_series, kind_to_df_map.items(),
-                                                  chunksize=chunksize), total=total_number_of_expected_results,
+
+    if serial:
+        map_function = map
+    else:
+        map_function = partial(pool.imap_unordered, chunksize=chunksize)
+
+    extracted_features = tqdm(map_function(partial_extract_features_for_one_time_series, kind_to_df_map.items()),
+                              total=total_number_of_expected_results,
                               desc="Feature Extraction", disable=settings.disable_progressbar)
 
     pool.close()
@@ -174,45 +184,6 @@ def _extract_features_parallel_per_kind(kind_to_df_map, settings, column_id, col
 
     pool.join()
     return result
-
-
-def _extract_features_serial(kind_to_df_map, settings, column_id, column_value):
-    """
-    Feature extraction without parallelization.
-
-    :param kind_to_df_map: The time series to compute the features for in our internal format
-    :type kind_to_df_map: dict of pandas.DataFrame
-
-    :param column_id: The name of the id column to group by.
-    :type column_id: str
-    :param column_value: The name for the column keeping the value itself.
-    :type column_value: str
-
-    :param settings: settings object that controls which features are calculated
-    :type settings: tsfresh.feature_extraction.settings.FeatureExtractionSettings
-
-    :return: The (maybe imputed) DataFrame containing extracted features.
-    :rtype: pandas.DataFrame
-    """
-    partial_extract_features_for_one_time_series = partial(_extract_features_for_one_time_series,
-                                                           column_id=column_id,
-                                                           column_value=column_value,
-                                                           settings=settings)
-
-    total_number_of_expected_results = len(kind_to_df_map)
-    extracted_features = tqdm(map(partial_extract_features_for_one_time_series, kind_to_df_map.items()),
-                              total=total_number_of_expected_results,
-                              desc="Feature Extraction", disable=settings.disable_progressbar)
-
-    # Concatenate all partial results
-    result = pd.concat(extracted_features, axis=1, join='outer').astype(np.float64)
-
-    # Impute the result if requested
-    if settings.IMPUTE is not None:
-        settings.IMPUTE(result)
-
-    return result
-
 
 def _extract_features_parallel_per_sample(kind_to_df_map, settings, column_id, column_value):
     """
--- a/tsfresh/feature_extraction/extraction.py
+++ b/tsfresh/feature_extraction/extraction.py
@@ -114,14 +114,14 @@ def extract_features(timeseries_container, feature_extraction_settings=None,
 
     # Calculate the result
     if parallelization == 'per_kind':
-        result = _extract_features_parallel_per_kind(kind_to_df_map, feature_extraction_settings,
-                                                     column_id, column_value)
+        result = _extract_features_per_kind(kind_to_df_map, feature_extraction_settings,
+                                            column_id, column_value)
     elif parallelization == 'per_sample':
         result = _extract_features_parallel_per_sample(kind_to_df_map, feature_extraction_settings,
                                                        column_id, column_value)
     elif parallelization == 'no_parallelization':
-        result = _extract_features_parallel_per_kind(kind_to_df_map, feature_extraction_settings,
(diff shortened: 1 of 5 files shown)


MaxBenChrist:
> TLDR: I increased the runtime by approx 50% on our test dataset while remaining the content of the calculations.

Your mean Decreased, right?

 I will go over this Pr tomorrow. Looks like a new release is coming in ;)

tsfresh/feature_extraction/extraction.py
```diff
@@ -174,6 +176,44 @@ def _extract_features_parallel_per_kind(kind_to_df_map, settings, column_id, col
     return result
 
 
+def _extract_features_serial(kind_to_df_map, settings, column_id, column_value):
```
MaxBenChrist: I mean you said that it is for debugging purposes, but why exactly?

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -939,32 +947,36 @@ def ar_coefficient(x, c, param):
...
+
+    for parameter_combination in param:
+        k = parameter_combination["k"]
+        p = parameter_combination["coeff"]
+
+        column_name = "{}__ar_coefficient__k_{}__coeff_{}".format(c, k, p)
 
-    for k in df_cfg["k"].unique():
-        coeff = df_cfg[df_cfg["k"] == k]["coeff"]
         try:
```
MaxBenChrist: i would move the try block one alignment deeper to line 965

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -939,32 +947,36 @@ def ar_coefficient(x, c, param):
...
-
-            for p in coeff:
-                if p <= k:
-                    try:
-                        res_tmp["{}__ar_coefficient__k_{}__coeff_{}".format(c, k, p)] = mod[p]
-                    except IndexError:
-                        res_tmp["{}__ar_coefficient__k_{}__coeff_{}".format(c, k, p)] = 0
-                else:
-                    res_tmp["{}__ar_coefficient__k_{}__coeff_{}".format(c, k, p)] = np.NaN
+            if k not in calculated_ar_params:
```
MaxBenChrist: very elegant

MaxBenChrist:
We should try not to drop the coverage...

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -762,11 +766,20 @@ def number_peaks(x, n):
     :return: the value of this feature
```
MaxBenChrist: I thought about this feature a little bit. I think there is great potential to speed up calculations here if we also make this an apply one.

Also I have an idea for a speed up: At the moment we compare too many values (`x_reduced > np.roll`). Instead we only have to compare neighbours. We could save only if the neighbour is greater or smaller (-1 if smaller, 1 if bigger). Then we are searching for subsequences. I feel that this way the calculations can be greatly speed up.

tsfresh/feature_extraction/extraction.py
nils-braun: Two things here:
(1) profiling gets a lot harder (if not impossible) if you have multiprocessing. For example, tools like line_profiling etc do not work as expected
(2) If you do runtime studies, you do not want to be dependent on the CPUs scheduler and the OS, so you try to have the same setup all the time => singleprocessing should be better here

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -762,11 +766,20 @@ def number_peaks(x, n):
...
+    x_reduced = x[n:-n]
+
+    res = None
     for i in range(1, n + 1):
-        res += [(x > np.roll(x, i))[n:-n]]
-        res += [(x > np.roll(x, -i))[n:-n]]
-    return sum(reduce(lambda a, b: a & b, res))
+        result_first = (x_reduced > np.roll(x, i)[n:-n])
+
+        if res is None:
```
nils-braun: No, because you need to do a &= with it and 0 has the size of 1 (a single value) whereas result_first has a finite length (an array). This will break.

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -762,11 +766,20 @@ def number_peaks(x, n):
...
-    res = list()
+    x = np.asarray(x)
+    x_reduced = x[n:-n]
+
+    res = None
     for i in range(1, n + 1):
-        res += [(x > np.roll(x, i))[n:-n]]
-        res += [(x > np.roll(x, -i))[n:-n]]
-    return sum(reduce(lambda a, b: a & b, res))
+        result_first = (x_reduced > np.roll(x, i)[n:-n])
```
nils-braun: The method stored each temporary result of the > operation in a list and reduced them at once in the end. This is not needed as the operation can be done immediately.
The [n:-n] and the > operation are actually quite costly (I have profiled it), so I tried to reduce the amount of times we have to do this.

tsfresh/feature_extraction/feature_calculators.py
nils-braun: Yes, this is probably true. But as stated in the conversation of this PR, I have already reduced this function to a quite small and reasonable time. We should concentrate on the last 4 runtime issues (augmented_dicker_fuller 14 %, mean_abs_change_qunatiles 25 %, autocorrelation 14 %, quantile 6 %) if we want to make the extraction even faster.

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -762,11 +766,20 @@ def number_peaks(x, n):
...
+        if res is None:
+            res = result_first
+        else:
+            res &= result_first
+
+        res &= (x_reduced > np.roll(x, -i)[n:-n])
+    return sum(res)
 
 
 @set_property("fctype", "apply")
```
nils-braun: Yes, this is a very crucial point. pandas is very nice, but also very slow. Just using the numpy methods speeds up things tremendously.

See my comment in the conversation for more on this.

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -844,39 +861,30 @@ def cwt_coefficients(x, c, param):
...
 
-        for w in df_cfg[df_cfg["widths"] == widths]["w"].unique():
+    for parameter_combination in param:
+        widths = parameter_combination["widths"]
+        w = parameter_combination["w"]
+        coeff = parameter_combination["coeff"]
 
-            # get the coefficients corresponding to this array of widths
-            coeff = df_cfg[(df_cfg["widths"] == widths) & (df_cfg["w"] == w)]["coeff"].unique()
+        if widths not in calculated_cwt:
```
nils-braun: and faster, yes. This was the main intention :-)
All changes in this PR are just to make things faster (and cleaner, but this was just a side product).

nils-braun:
> from what i saw, you used the following techniques to speed up calculations

> * cache results by making it apply from aggregate (symmetry_looking)
> * do not sort parameters, use dict instead (ar_coefficient, ctw coefficients)
> * ?? (number_peaks)

One of the major things for increasing the performance was the reduction of the usage of pandas dataframes and series. They are very handle, but unfortunately very slow - especially when it comes to appending or concationating of results. Here, it is much better to store all results in a python list/dict and create a Series/DataFrame when everything is finished. Also, applying mean/std etc. and indexing is much faster on pure numpy arrays than on pandas Series.
Another optimization was to not use DataFrames for the configurations (e.g. in cwt_coefficients), because we would need to groupby/index/mask many times which is slow. Better use a cache for the calculated values and work through all configuration parameters one-by-one.

nils-braun:
And I do not know why the coverage has dropped...

tsfresh/feature_extraction/extraction.py
MaxBenChrist: Ok sure, it makes sense from a . I feel that we have a code duplication here. the main difference are the `map` versus `pool.imap_unordered` right? I want to prevent code duplication here. 

Do you think you can modify the `_extract_features_parallel_per_kind` method to do it in serial to use that for debugging?

tsfresh/feature_extraction/feature_calculators.py
MaxBenChrist: ok

tsfresh/feature_extraction/feature_calculators.py
MaxBenChrist: ok

tsfresh/feature_extraction/feature_calculators.py
MaxBenChrist: okay, I will open an issue for the comparing of neighbours.

MaxBenChrist: This pr is great already ;D

MaxBenChrist:
> Another optimization was to not use DataFrames for the configurations (e.g. in cwt_coefficients), because we would need to groupby/index/mask many times which is slow. Better use a cache for the calculated values and work through all configuration parameters one-by-one.

This was an aha-moment for me. Dictionaries are great.

MaxBenChrist:
> And I do not know why the coverage has dropped...

Normally, you can just click on the details field, the coveralls page shows you the lines where coverage is missing. But here it says that it can't find the source file? strange

tsfresh/feature_extraction/extraction.py
nils-braun: Yes, this sounds like a very good idea!

MaxBenChrist:
So, the only thing that is missing is the adaption of the _extract_features_parallel_per_kind method to do it in serial. Then we can merge this thing here. Maybe this also helps with the coverage

MaxBenChrist:
Yes, coverage increased again :)

nils-braun:
Ok, I will merge now and we do a new release?

PR #188, opened 2017-04-03
Fix issue 187 imputing on float64

### LINKED ISSUE #187: The impute method does not work properly on np.float64 or np.int64
It returns the same DataFrame. Casting it as `np.float32` seem to resolve this

MaxBenChrist:
fixes #187


MaxBenChrist:
well, iloc is not helping...

PR #201, opened 2017-05-14
include unit test for num_cwt_peaks

mynameisvinn:
included unit test for num_cwt_peaks function.


MaxBenChrist:
Looks same, but why doed the coverage decrease?!

PR #215, opened 2017-06-16
fix implementation of mannwhitneyu function

smueller18:
The tsfresh master is not compatible with scipy v0.17.0 as set as minimum in the `requirements.txt`.
In this version, the default of parameter `alternative` of function `scipy.stats.mannwhitneyu(x, y, use_continuity=True, alternative='two-sided')` is set to `two-sided`. In this case, the doubling of the p-value would be done twice by calling `tsfresh.feature_selection.significance_tests.target_binary_feature_real_test(x, y, test)`.
Since scipy v0.18.0, the default value changed to ``None`` and leaving the parameter `alternative` on that default value results in a deprecated warning. So the best solution here is to add the parameter `alternative='two-sided` and remove the manual doubling of the p-value. With this fix, all recent allowed version of scipy calculate the right p-value.


MaxBenChrist:
Hi @smueller18, thanks for the pr and the detailed explanation. 

Coverage reports that the coverage dropped on one line. I assume it is the newline that you added to tsfresh/feature_selection/significance_tests.py. Can you remove that? 

I will merge your pr after that.

smueller18:
@MaxBenChrist done

MaxBenChrist:
Well I do not want to be so picky but according to the gui, the last commit

ac8b527d50695e486ae217c502ceda86fc3e999d

is not changing anything ?!

MaxBenChrist:
bildschirmfoto 2017-06-16 um 13 28 38

smueller18:
I removed the line and commited directly over the browser.
Did't know that github.com adds a newline to every file which is commited over the browser (also like githubs Atom editor used for the commit before).
37de74dcdefe80ce10c385c0d5d3951b16a6f320 fixes it now.

MaxBenChrist:
> I removed the line and commited directly over the browser.
Did't know that github.com adds a newline to every file which is commited over the browser (also like githubs Atom editor used for the commit before).

me neither. anyway, welcome @smueller18  to the selected circle of tsfresh contributors ;)

Issue #205, opened 2017-05-27
Performance comparison with pandas built in functions


liorshk:
Hi,
I have made some performance comparison of 3 basic feature extractors: mean, sum of values, and standard deviation between your feature extractors and pandas and I found some major comparison differences.

I changed the code in _extraction.py_ at line _345_
```python
if isDirect:
  del column_name_to_aggregate_function[column_prefix+'__sum_values']
  del column_name_to_aggregate_function[column_prefix+'__standard_deviation']
  del column_name_to_aggregate_function[column_prefix+'__mean']
  extracted_features =  pd.DataFrame(index=dataframe[column_id].unique())
  extracted_features[column_prefix+'__mean']= dataframe.groupby(column_id)[column_value].mean()
  extracted_features[column_prefix+'__sum_values']= dataframe.groupby(column_id)[column_value].sum()
  extracted_features[column_prefix+'__standard_deviation']= dataframe.groupby(column_id)[column_value].std()
else:
  extracted_features = dataframe.groupby(column_id)[column_value].aggregate(column_name_to_aggregate_function)
```
As you can see, I only called pandas built in function instead of calling the implementation of the feature extractor.

The following plot compares different time series lengths with 1000 ids
![TimeSeriesLength](https://cloud.githubusercontent.com/assets/7127502/26523950/59968040-432d-11e7-91c4-73f29dc24248.png)

As you can see, there is a major comparison difference.
The reason for that is that probably because pandas functions are [optimized](https://pandas.pydata.org/pandas-docs/stable/enhancingperf.html).

I think that the feature extractors should be optimized using [numba](http://numba.pydata.org/) or cython

nils-braun:
Thank you for this nice study @liorshk!
We will definitely have a look into this. For this we will have to change some internals in the feature calculators, but this should be doable.

MaxBenChrist:
@liorshk: Nice observation.

But, I do not see the urge to optimize this part. 

Yes, the functions  `.mean()`, `.max()`, `.min()`, `.std()` and `.median()` are optimized on a groupby object (so `gp.f()` is faster than `gp.apply(f)`).

However, if we want to exploit that, we would have to implement some annoying routines, making the code unnecessary complicated. We probably would have to add a fourth kind of feature calculator. 

The return is not worth it from my point of view.

nils-braun:
Well, we could at least use numba or something similar to speed up the calculation by itself. Let's study this in greater detail.

[...]

PR #217, opened 2017-06-18
Remove pandas internal implementation

nils-braun:
Following the discussion in #205, I have change some internal structures. Generally speaking, the extraction is now simplified:
* all calculators either return a single number or a list of tuples (parameter, result), depending on whether they are a "combiner" or "simple". There is no need for distinguishing between "apply" and "aggregate" as we will see below.
* extraction works now as follows: split the input df into single chunks by "kind" and "id". The extraction is then run on parallel on each chunk without a groupy - apply, but just using a simple loop over all feature calculators.
* this means, our parallelisation is now much easier and cleaner (there is no difference between by_sample or by_kind, just parallel now)
* the handling of the correct feature naming is now mostly handled by the feature extraction and not by the calculators.

Some benefits of the new solution:
* logic is much clearer now
* better separation between calculators and extraction itself
* smaller code basis
* and last but not least: another speed boost!

Speed boost:
Very simple test script:
```
from tsfresh.examples.robot_execution_failures import download_robot_execution_failures, load_robot_execution_failures

import time

download_robot_execution_failures()
df, y = load_robot_execution_failures()

start = time.clock()
df = extract_features(df[:100], column_id='id', column_sort='time')
end = time.clock()

print(100 * (end - start))
```
See below how much faster the new implementation is!


nils-braun:
All right, some performance studies with the newest code in this PR are attached:
timing
You see the performance in calculating the features (using multiprocessing in the default settings) of different subsets of the robot dataset using the master (commit a5...) and this PR (commit e2...). You see that for all lengths a speedup of at least 1.5 (up to 3!) is achieved. The script used for calculating these plots is also in this PR.
Additionally, as I have already said, I like the data flow now much better and the "strange" and "artificial" difference between aggregate and apply is also gone.
Documentation is also updated, so this PR is ready for merging after the review :-)

nils-braun:
Thank you again @NoamGit and @liorshk for starting the discussion which lead eventually to this PR.

docs/text/parallelization.rst
```diff
@@ -29,31 +29,8 @@ Parallelization of Feature Extraction
...
-
--   For small data sets the difference between parallelization per kind or per sample should be negligible.
--   For data sets with one kind of time series parallelization per sample results in a decent speed up that grows
-    with the number of samples.
--   The more kinds of time series the data set contains, the more samples are necessary to make parallelization
-    per sample worthwhile.
--   If the data set contains more kinds of time series than available cpu cores, parallelization per kind is
-    the way to go.
+To do performance studies and profiling, it sometimes quite useful to turn off parallelization at all. This can be
+handled by the boolean parameter `parallelization`.
```
MaxBenChrist: what about just one parameter `n_jobs` ? 

we could decide that for `n_jobs=1` no parallelization is used.

MaxBenChrist:
Amazing work @nils-braun. As this pr is really big I want to go over it several times. So far everything seems valid

tsfresh/feature_extraction/extraction.py
```diff
@@ -214,171 +190,112 @@ def _extract_features_per_kind(kind_to_df_map,
...
 
     if not chunksize:
-        chunksize = _calculate_best_chunksize(kind_to_df_map, n_processes)
+        chunksize = _calculate_best_chunksize(data_in_chunks, n_processes)
 
-    total_number_of_expected_results = len(kind_to_df_map)
+    total_number_of_expected_results = len(data_in_chunks)
 
-    if serial:
+    if not parallelization:
```
MaxBenChrist: here we could use a check if `n_jobs==1`

tsfresh/feature_extraction/extraction.py
```diff
@@ -214,171 +190,112 @@ def _extract_features_per_kind(kind_to_df_map,
...
-    function. The result objects are temporarily stored in a fifo queue from which they can be retrieved in order
-    of submission.
+    return result
 
-    :param kind_to_df_map: The time series to compute the features for in our internal format
-    :type kind_to_df_map: dict of pandas.DataFrame
 
-    :param column_id: The name of the id column to group by.
-    :type column_id: str
+def _do_extraction_on_chunk(chunk, default_fc_parameters, kind_to_fc_parameters):
```
MaxBenChrist: i do not like the parameter name chunk, as this chunk thing actually contains `chunk_id`, `chunk_kind` and the `data`

tsfresh/feature_extraction/extraction.py
```diff
@@ -132,48 +123,38 @@ def extract_features(timeseries_container, default_fc_parameters=None,
 
     # Always use the standardized way of storing the data.
     # See the function normalize_input_to_internal_representation for more information.
-    kind_to_df_map, column_id, column_value = \
-        dataframe_functions.normalize_input_to_internal_representation(df_or_dict=timeseries_container,
-                                                                       column_id=column_id,
+    df_melt, column_id, column_kind, column_value = \
```
MaxBenChrist: `df_melt` contains the chunks, right?

maybe we can construct an object for the chunks (sorry for my java ;-) )

docs/text/parallelization.rst
nils-braun: Yes, seems reasonable. I would actually prefer to use `n_jobs = 0` for no parallelisation, as one could test the overhead of parallelisation when comparing `n_jobs = 0` and `n_jobs = 1`.

tsfresh/feature_extraction/extraction.py
nils-braun: Yes, me neither. Do you have a better idea?

tsfresh/feature_extraction/extraction.py
nils-braun: Well, at this point it is still a "normal" pandas dataframe - so no need for a custom object here.

It was just reordered, to be in the typical form: `id kind time value`.

docs/text/parallelization.rst
MaxBenChrist: Great idea! I like it

tsfresh/feature_extraction/extraction.py
MaxBenChrist: alright. what about the good old timeseries_container as name for `df_melt`?

tsfresh/feature_extraction/extraction.py
MaxBenChrist: not at this point...

tsfresh/feature_extraction/extraction.py
```diff
@@ -214,171 +190,112 @@ def _extract_features_per_kind(kind_to_df_map,
...
-    :type column_id: str
+def _do_extraction_on_chunk(chunk, default_fc_parameters, kind_to_fc_parameters):
+    """
+    Main function of this module: use the feature calculators defined in the
+    default_fc_parameters or kind_to_fc_parameters parameters and extract all
+    features on the chunk.
 
-    :param column_value: The name for the column keeping the value itself.
-    :type column_value: str
+    The chunk consists of the chunk id, the chunk kind and the data (as a Series),
```
MaxBenChrist: you could add what `chunk_id` and `chunk_kind` are. Ints? str?

tsfresh/feature_extraction/extraction.py
jneuff: Why `df_melt`? I don't get the name.

tsfresh/feature_extraction/extraction.py
jneuff: I actually think that it's a good name. A chunk consists of an id, a kind and the data.

tsfresh/feature_extraction/extraction.py
```diff
@@ -214,171 +190,112 @@ def _extract_features_per_kind(kind_to_df_map,
...
+    The <parameters> are in the form described in :mod:`~tsfresh.utilities.string_manipulation`.
 
-    :param chunksize: The size of one chunk for the parallelisation
-    :type chunksize: None or int
+    :param chunk: A tuple of chunk_id, chunk_kind, data
+    :param default_fc_parameters: A dictionary of feature calculators.
+    :param kind_to_fc_parameters: A dictionary of fc_parameters for special kinds or None.
+    :return: A list of calculated features.
+    """
+    chunk_id, chunk_kind, data = chunk
```
jneuff: However, I would rename these. What about: `sample_id, kind, data = chunk`?

nils-braun: `df_melt` refers to the pandas method, that does exactly, what is done in the normalize function (see https://pandas.pydata.org/pandas-docs/stable/generated/pandas.melt.html).

tsfresh/feature_extraction/extraction.py
nils-braun: `df_melt` refers to the pandas method, that does exactly, what is done in the normalize function (see https://pandas.pydata.org/pandas-docs/stable/generated/pandas.melt.html).

tsfresh/feature_extraction/extraction.py
nils-braun: Well, actually I do not know. ID can be anything (pandas allows ints as well as strings or other objects) and the kind can also be anything (even an array).

nils-braun:
All right! Thank you everyone! I will merge this, when the build is ready!

MaxBenChrist:
This is so amazing! Great work @nils-braun! 👍

Next step: tsfresh in a cluster? ;-)

nils-braun:
Bäh, I will have to fix the unittests because of the renaming :-D
Will do so this evening...

nils-braun:
All right, I will merge in the next half an our, if no one disagrees. This is another small API change, so we should maybe push the version.

Issue #197, opened 2017-04-26
roll_time_series creates same values


TKlerx:
Hi all,

thanks for the great library and the new feature of rolling time series.
Unfortunately, there seems to be a bug in the new feature:
Instead of using the values of the previous timestamps, always the same values are used for a point in time t.
I attached a notebook where I implemented the rolling time series on my own and compared it with the tsfresh version.
[bug-tsfresh-rolling.zip](https://github.com/blue-yonder/tsfresh/files/958964/bug-tsfresh-rolling.zip)
I could supply a test case with and create a pull request if you wish.

Versions:
Win10 x64
tsfresh 0.7.1
python = 3.5 and 3.6

The expected output of the example given at https://tsfresh.readthedocs.io/en/v0.7.1/text/rolling.html should should be

id | t | variable | value
------------ | ------------- | ------------ | ------------- 
1  |2 	|x_shift0 	|2.0 
1 |2 	|x_shift1 	|1.0 	
1 |2 	|y_shift0 	|6.0 	
1 |2 	|y_shift1 	|5.0 	

but the output is

id | t | variable | value
------------ | ------------- | ------------ | ------------- 
1 |2 	|x _shift0|2.0 
1|2 	|x_shift1 |2.0 	
1 |2 	|y_shift0 |6.0 	
1|2 	|y_shift1 |6.0

nils-braun:
Hi @TKlerx!
Thank you for your question. And you even included a notebook - this is how we like our bug reports :+1: 

However, I think I do not understand your question fully. The time column is also shifted, so the time of `t = 2` has always the same `x` and `y` values - we just "delete" the "later" rows (or the earlier ones, depending on the direction). 
If you look into `df_tsf_rolling[(df_tsf_rolling.id == "id=1, shift=1") & (df_tsf_rolling.variable == "x")].sort_values(by=['t','id'])` and `df_tsf_rolling[(df_tsf_rolling.id == "id=1, shift=0") & (df_tsf_rolling.variable == "x")].sort_values(by=['t','id'])` in your notebook, you may see how it works.

Please ask, if you still have a question.

TKlerx:
Hi @nils-braun,

thank you for your quick answer. You are correct, I misunderstood the API. I thought that I would get the values of id=1 from t=1 added to id=2 and t=2.

As you mentioned, one has to group by the new id.

Is there any possibility to define the maximum length of the resulting time series?
I would like to use it in the following scenario:
[...]

Issue #174, opened 2017-03-25
Improve documentation + provide notebook for rolling mechanism


MaxBenChrist:
the rolling mechanism may be hard to understand. 

We could provide a notebook with examples + better documentation

MaxBenChrist:
An explanation similar to 

If I find time during the next days, I am ghttps://www.mathworks.com/help/econ/rolling-window-estimation-of-state-space-models.html

would be good

MaxBenChrist:
I started sketching something on branch notebook_rolling. Feel free to also contribute to that branch

PR #221, opened 2017-06-29
Update docs of rolling page

TKlerx:
Hi everyone,

as mentioned in #197 and #174, this is the PR for updating the docs of rolling (sorry for taking so long).
There is a "comment" to discuss whether the first scenario described really fits.

I hope this somehow helps and would be happy to improve this PR.

### DIFF
--- a/docs/text/rolling.rst
+++ b/docs/text/rolling.rst
@@ -21,10 +21,8 @@ the last 10 steps.
 A rolling mechanism creates such time series for every time step by creating sub time series of the sensor data of the
 last 10 time steps.
 
-<!---
-The above scenario cannot be solved by tsfresh, can it?
-How do you create a rolling time series that always contains the current value and the last 10 steps?
--->
+The scenario described above cannot be easily computed by tsfresh in the current version.
+However, it is planned to be included in one of the upcoming releases (> v0.8.1).
 
 Another example can be found in streaming data, e.g. in Industry 4.0 applications.
 Here you typically get one new data row at a time and use this to for example predict machine failures. To train your model,
--- a/docs/text/rolling.rst
+++ b/docs/text/rolling.rst
@@ -183,7 +183,9 @@ that are supported by tsfresh.
 
 Parameters and Implementation Notes
 ------
-The above example was used for demonstrating how the rolling of time series works.
+The above example demonstrates the overall rolling mechanism, which creates new time series.
+Now we discuss the naming convention for such new time series:
+
 For identifying every subsequence, tsfresh introduces a qualifier ("shift") that shows how far the time series was shifted.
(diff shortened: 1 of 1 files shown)


docs/text/rolling.rst
```diff
@@ -21,6 +21,11 @@ the last 10 steps.
 A rolling mechanism creates such time series for every time step by creating sub time series of the sensor data of the
 last 10 time steps.
 
+<!---
+The above scenario cannot be solved by tsfresh, can it?
```
MaxBenChrist: yes, it can. 

the `maximum_number_of_timeshifts` parameter of the `roll_time_series` controls how many ticks you look into the past (here 10)

docs/text/rolling.rst
```diff
@@ -174,4 +179,85 @@ If you set rolling to -1, you end up with features for the time series, rolled i
 +----+------+----+----+
 
 We only gave an example for the flat DataFrame format, but rolling actually works on all 3 :ref:`data-formats-label`
-that are supported by tsfresh.
\ No newline at end of file
+that are supported by tsfresh.
+
+Parameters and Implementation Notes
+------
+The above example was used for demonstrating how the rolling of time series works.
```
MaxBenChrist: The above example already demonstrates the rolling of time series.

docs/text/rolling.rst
TKlerx: Yes, but you cannot do the following (correct me if I am wrong):
Extract all sub time series of length 10 from a given time series.
But that is what you usually want to do for training. Take into account the current values of the time series and the most recent (e.g. 10) values of the past.

The `maximum_number_of_timeshifts` controls how many shifts (of length 1) to perform into the past.

docs/text/rolling.rst
TKlerx: Thanks, done!

docs/text/rolling.rst
nils-braun: @TKlerx yes, you are totally correct, this can not be handled by the methods in our package.

TKlerx: That's exactly the thing I am looking for. If I find the time, I will try to implement it.
Is there any preferred way how and where to place it?

nils-braun: Sorry, I missed answering this :-)
Yes, if it would be possible to include this (maybe as a parameter) to the already implemented rolling function, this would be very neat, If you have time, feel free to do so. If not, I can work on this.

nils-braun:
@MaxBenChrist Can we merge?

MaxBenChrist:
Please do. not merge, i think the comment starting in line 24 should be removed

PR #227, opened 2017-07-10
P value augmented dickey fuller

MaxBenChrist:
I added the pvalue and the used lag as features. 

they were calculated anyway so we get more features for free. 

To implement this I had to change the type of augmented_dickey_fuller from simple to combiner

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ install:
 - pip install .
 - pip install -r requirements.txt
 - pip install -r test-requirements.txt
+- pip freeze
 env:
 - NUMBER_OF_CPUS=1
 script:
--- a/tsfresh/feature_extraction/feature_calculators.py
+++ b/tsfresh/feature_extraction/feature_calculators.py
@@ -296,7 +296,9 @@ def augmented_dickey_fuller(x, param):
     series sample. This feature calculator returns the value of the respective test statistic.
 
     See the statsmodels implementation for references and more details.
-
+    
+    :param x: the time series to calculate the feature of
+    :type x: pandas.Series
     :param param: contains dictionaries {"attr": x} with x str, either "teststat", "pvalue" or "usedlag"
     :type param: list
     :return: the value of this feature
--- a/tsfresh/feature_extraction/feature_calculators.py
+++ b/tsfresh/feature_extraction/feature_calculators.py
@@ -289,8 +289,8 @@ def mean_autocorrelation(x):
         return np.nanmean(r / var)
 
 
-@set_property("fctype", "simple")
-def augmented_dickey_fuller(x):
+@set_property("fctype", "combiner")
+def augmented_dickey_fuller(x, param):
     """
     The Augmented Dickey-Fuller test is a hypothesis test which checks whether a unit root is present in a time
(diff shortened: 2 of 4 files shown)


MaxBenChrist:
Well, that is weird. locally, the tests are passing, on travis, the build fails. 

I will check if there is a problem with the stats model package version

jneuff:
The relevant test also fails on my machine (linux).

MaxBenChrist:
That is weird. On my Mac it passes.

.travis.yml
```diff
@@ -15,6 +15,7 @@ install:
 - pip install .
 - pip install -r requirements.txt
 - pip install -r test-requirements.txt
+- pip freeze
```
nils-braun: Is this somehow related to this PR?

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -289,26 +289,32 @@ def mean_autocorrelation(x):
...
+        res = np.NaN, np.NaN, np.NaN
     except ValueError:  # occurs if sample size is too small
-        return np.NaN
+        res = np.NaN, np.NaN, np.NaN
+
+    return [("attr_{}".format(config["attr"]),
+                  res[0] if config["attr"] == "teststat"
+             else res[1] if config["attr"] == "pvalue"
+             else res[2] if config["attr"] == "usedlag" else np.NaN)
+            for config in param]
```
nils-braun: Isn't this simpler with a dict or a list? This seems a bit "ugly" :-D

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -289,26 +289,32 @@ def mean_autocorrelation(x):
...
 
-@set_property("fctype", "simple")
-def augmented_dickey_fuller(x):
+@set_property("fctype", "combiner")
+def augmented_dickey_fuller(x, param):
     """
     The Augmented Dickey-Fuller test is a hypothesis test which checks whether a unit root is present in a time
     series sample. This feature calculator returns the value of the respective test statistic.
 
     See the statsmodels implementation for references and more details.
```
nils-braun: where is the docu of x?

MaxBenChrist: went missing :D, will add it

.travis.yml
MaxBenChrist: yes. somehow, the unit tests were locally passing but not on travis

I use the pip freeze at the end of the build to inspect which versions of the packages are installed

nils-braun: Well, ok. But still, you can delete this now, can't you? :-D

MaxBenChrist: I would leave it for future inspections . in my opinion it improves the travis log.

tsfresh/feature_extraction/feature_calculators.py
MaxBenChrist: well, the dictionary does not cover the last else part

MaxBenChrist:
Can you life with both the `pip freeze` and the `else` construction @nils-braun ?

nils-braun:
Yep, I'll fix it after the release

Issue #222, opened 2017-07-03
Bug in `mean_autocorrelation`?


nikhaas:
OS: Ubuntu 16.04, Python 3.5, tsfresh 0.8.1

There might be a bug in the method `mean_autocorrelation` ([Link to line of code](https://github.com/blue-yonder/tsfresh/blob/master/tsfresh/feature_extraction/feature_calculators.py#L288))

The calculation of the autocorrelation for all lags 0...k simultaneously following the equation for [autocorrelation estimation](https://en.wikipedia.org/wiki/Autocorrelation#Estimation) (as mentioned in the documentation) can be achieved by using `np.correlate`.

But the results of the `tsfresh` code do not match the result when averaging the results of [`statsmodels.tsa.stattools.acf`](http://www.statsmodels.org/dev/generated/statsmodels.tsa.stattools.acf.html), see this example with the time series `x = [1, 2, -3]` from the [tsfresh test for mean autocorrelation](https://github.com/blue-yonder/tsfresh/blob/master/tests/feature_extraction/test_feature_calculations.py#L140):

```python
# Implementation of tsfresh
x = [1, 2, -3]
var = np.var(x)
n = len(x)
r = np.correlate(x - np.mean(x), x - np.mean(x), mode='full')
r = r[0: (n - 1)] / np.arange(n - 1, 0, -1)
print(np.nanmean(r / var))
# out: -0.589285714286

# Using statsmodels acf function, lag 0 is omitted by indexing!
print(np.nanmean(acf(x, unbiased=True)[1:]))
# out: -0.535714285714
```
Be aware that lag 0 (which is trivially 1) is omitted during averaging, but is an output of the `acf` function.

Reversing the order of the divisors provided by `np.arange` from

```python
r = r[0: (n - 1)] / np.arange(n - 1, 0, -1)
```

to

```python
r = r[0: (n - 1)] / np.arange(1, n, 1)
```
achieves the expected result `-0.535714285714`. This is because the application of `np.correlate` to a timeseries itself yields as a result a symmetric output, for example

```python
x = [1, 2, -3]
np.correlate(x - np.mean(x),x - np.mean(x), mode='full')
# out: array([ -3.,  -4.,  14.,  -4.,  -3.])
# accroding to lag: [2, 1, 0, 1, 2]
```
[...]

PR #231, opened 2017-07-10
use acf from statsmodels to calculate mean_autocorrelation

MaxBenChrist:
solves issue #222


MaxBenChrist:
I do not understand why the coverage decreases. 

It could be that the an unrelated test, e.g. for the `augmented_dickey_fuller` throws an error non deterministically during the testing. Then the coverage drops and one line is less "tested".

jneuff:
> I do not understand why the coverage decreases.

> It could be that the an unrelated test, e.g. for the augmented_dickey_fuller throws an error non deterministically during the testing. Then the coverage drops and one line is less "tested".

No, it's just due to the decrease of code in the feature calculator. This reduces the tested lines versus the untested.

PR #232, opened 2017-07-10
Add support for one-vs-rest multiclass selection

jneuff:
I still need to update the multiclass notebook and the docs.


tsfresh/convenience/relevant_extraction.py
```diff
@@ -142,7 +146,8 @@ def extract_relevant_features(timeseries_container, y, X=None,
                             test_for_real_target_real_feature=test_for_real_target_real_feature,
                             fdr_level=fdr_level, hypotheses_independent=hypotheses_independent,
                             n_jobs=n_jobs,
-                            chunksize=chunksize)
+                            chunksize=chunksize,
+                            multiclass=multiclass)
```
MaxBenChrist: maybe `is_multiclass` to reflect that it is a boolean flag?

tsfresh/feature_selection/selection.py
```diff
@@ -20,7 +20,8 @@ def select_features(X, y, test_for_binary_target_binary_feature=defaults.TEST_FO
                     test_for_real_target_binary_feature=defaults.TEST_FOR_REAL_TARGET_BINARY_FEATURE,
                     test_for_real_target_real_feature=defaults.TEST_FOR_REAL_TARGET_REAL_FEATURE,
                     fdr_level=defaults.FDR_LEVEL, hypotheses_independent=defaults.HYPOTHESES_INDEPENDENT,
-                    n_jobs=defaults.N_PROCESSES, chunksize=defaults.CHUNKSIZE):
+                    n_jobs=defaults.N_PROCESSES, chunksize=defaults.CHUNKSIZE,
+                    multiclass=False):
```
MaxBenChrist: same as above, would prefer name that reflects its boolean nature

tsfresh/feature_selection/selection.py
```diff
@@ -124,7 +128,20 @@ def select_features(X, y, test_for_binary_target_binary_feature=defaults.TEST_FO
...
         y = pd.Series(y, index=X.index)
 
-    df_bh = check_fs_sig_bh(X, y, n_jobs, chunksize, fdr_level, hypotheses_independent,
-                            test_for_binary_target_real_feature)
-
-    return X.loc[:, df_bh[df_bh.rejected].Feature]
+    if multiclass is True:
+        unique_labels = y.unique()
+        if len(unique_labels) < 3:
+            raise RuntimeError("Cannot perform multiclass selection on binary target")
```
MaxBenChrist: really? why? 

It can happen in a CV for a non binary target that the fold only contains two classes, this will break the whole cross validation. can we ditch that check?

jneuff: I don't like the idea of having two codepaths for binary classification. But you are right, this might just happen during cross validation. I'll remove the check.

However, this got me thinking.. Currently, we implicitly assume classification if the target is binary, e.g. `len(set(y)) == 2`. What about ditching that too?
Instead of the `multiclass` or `is_multiclass` parameter, we could have a `selection` parameter that defaults to `'regression'` but could be set to `'classification'` instead. Then, if `selection='regression'` and the target consists of `int` only, we produce a message proposing to set `selection='classification'`.

MaxBenChrist: 0.9.0 will have some breaking changes anyway, so now is the time to change the API.

I like your proposal.

I would default that parameter to "None", meaning that the method tries to infer the type of target by looking at the dtype. I would not only look for type `int` but also look for targets that can be casted to int, something like `[1.0, 0.0, 1.0]`. 

But, I am not sure if `selection` is the right name. Maybe something like `ml_task`?

jneuff: Yes, `ml_task` is better. Also, inference as a default is a good idea. The options then are `'auto'`, `'regression'` and `'classification'`.  I'll implement it tomorrow.

tests/feature_selection/test_selection.py
```diff
@@ -3,28 +3,54 @@
 # Maximilian Christ (maximilianchrist.com), Blue Yonder Gmbh, 2016
 
 import numpy as np
-from unittest import TestCase
 import pandas as pd
```
nils-braun: why have you deleted the test class?
I always like the classes, as they summarize things quite nicely.

tests/feature_selection/test_selection.py
```diff
@@ -3,28 +3,54 @@
...
+from pytest import raises
+from future.utils import lrange
 
 from tsfresh.feature_selection.selection import select_features
 
 
-class SelectionTestCase(TestCase):
+def test_assert_list():
+    with raises(TypeError):
```
nils-braun: what is against self.assertRaises?

jneuff: Nothing. I just prefer the tests to be 'flat', i.e. not part of a class here, as the tests don't share any data.

tests/feature_selection/test_selection.py
jneuff: In this case, all tests in the file were part of the same class. Then I prefer not to have the class, as it doen't add any value. But you are right, they can help group similar tests. I will readd them now that there is another function tested in the same file.

tests/feature_selection/test_selection.py
jneuff: Nothing, I just prefer the pytest function.

jneuff:
The PR is not ready for a merge:

- I need to update the docs. I am thinking about removing the multiclassification notebook and adding examples to the docs on how to use the parameter `ml_task`.
- The Transformers still do not support multiclassification and have different semantics about when to do regression. I will take care of that in this PR.

jneuff:
Still not done. But we can already discuss the change in naming I propose here.

tsfresh/feature_selection/feature_selector.py
```diff
@@ -102,15 +102,10 @@ def check_fs_sig_bh(X, y,
             "Feature",
             "type" (binary, real or const),
             "p_value" (the significance of this feature as a p-value, lower means more significant)
-            "rejected" (if the Benjamini Hochberg procedure rejected this feature)
+            "relevant" (True if the Benjamini Hochberg procedure rejected the null hypothesis for this feature)
```
nils-braun: You could explain the null hypothesis here again for people not reading the rest :-)

tsfresh/feature_selection/feature_selector.py
```diff
@@ -216,9 +211,9 @@ def _calculate_p_value(feature_column, y, target_is_binary, test_for_binary_targ
 
 def benjamini_hochberg_test(df_pvalues, hypotheses_independent, fdr_level):
     """
-    This is an implementation of the benjamini hochberg procedure that calculates which of the hypotheses belonging
-    to the different p-Values from df_p to reject. While doing so, this test controls the false discovery rate,
-    which is the ratio of false rejections by all rejections:
+    This is an implementation of the benjamini hochberg procedure that determines if the null hypothesis
+    for a given feature can be rejected. For this the test regards the features' p-values
```
nils-braun: regards?

tsfresh/feature_selection/selection.py
```diff
@@ -124,7 +136,141 @@ def select_features(X, y, test_for_binary_target_binary_feature=defaults.TEST_FO
...
+    :type y: pandas.Series
+    :return: 'classification' or 'regression'
+    :rtype: str
+    """
+    if y.dtype.kind in np.typecodes['AllInteger']:
+        ml_task = 'classification'
+    else:
+        ml_task = 'regression'
+
+    _logger.warning('Infered {} as machine learning task'.format(ml_task))
```
nils-braun: Is this really worth a "warning"?

tsfresh/feature_selection/selection.py
```diff
@@ -124,7 +136,141 @@ def select_features(X, y, test_for_binary_target_binary_feature=defaults.TEST_FO
...
+    :return: The combined relevance table
+    :rtype: pandas.DataFrame
+    """
+    def _append_label_to_p_value_column(a):
+        label, df = a
+        return df.rename(columns={'p_value': 'p_value_{}'.format(label)})
+
+    def _combine(a, b):
+        a.relevant |= b.relevant
+        return a.join(b.iloc[:,3])
```
nils-braun: here you assume a certain order of the columns. Maybe it would be better to work with name here (either when dropping or when selecting columns).

tsfresh/feature_selection/selection.py
```diff
@@ -124,7 +136,141 @@ def select_features(X, y, test_for_binary_target_binary_feature=defaults.TEST_FO
...
 
-    return X.loc[:, df_bh[df_bh.rejected].Feature]
+    relevance_tables = map(_append_label_to_p_value_column, relevance_tables_with_label)
+    relevance_table = reduce(_combine, relevance_tables)
+    relevance_table['p_value'] = relevance_table.iloc[:, 3:].values.min(axis=1)
```
nils-braun: Just a question of my understanding: is this what the user expects when he does a multiclass classification? The feature is rejected if it was rejected in one (single) of the binary classifications?

tsfresh/transformers/feature_selector.py
```diff
@@ -89,6 +89,11 @@ def __init__(self, test_for_binary_target_binary_feature=defaults.TEST_FOR_BINAR
 
         :param chunksize: Size of the chunks submitted to the worker processes
         :type chunksize: int
+
+        :param ml_task: The intended machine learning task. Either `'classification'`, `'regression'` or `'auto'`.
+                    Defaults to `'auto'`, meaning the intended task is inferred from `y`.
+                    If `y` has a boolean or integer dtype, the task is assumend to be classification, else regression.
+        :type ml_task: str
```
nils-braun: You are right, maybe we should add a method do "copy" documentation from the underlaying functions, for that we do not have to duplicate it all the time.
We could then maybe think about using **kwargs here.

But this is nothing for the next release or your PR ;-)

tsfresh/feature_selection/selection.py
jneuff: Yes, currently the rejection of H0 happens per class and then features that are considered relevant for one class are considered relevant for all classes. But I thought about that too, we could run the Benjamini-Hochberg-Test on the minimal p-values from all classes. That way, the FDR will be respected globally instead of just per class. If someone wants the features relevant to each class, they could still do their one-vs-rest selection by hand. What's your opinion @nils-braun and @MaxBenChrist ?

I'm going to change a lot more in the `feature_selection` module anyway, as currently the type of the feature ('binary', 'real', 'const') is determined for each of the multiple classes again..

tsfresh/transformers/feature_selector.py
jneuff: Yeah, that would be great! But it that's for another time :-D

tsfresh/feature_selection/selection.py
jneuff: Yes, I do not like that either. The direct alternative here would be to do string comparison (`startswith('p_value')`) because I do not know the exact name of the column. That I wanted to avoid. However, I will need to do some more rewriting of the module and the final version will likely not rely on this function.

tsfresh/feature_selection/selection.py
jneuff: I'm open to changing it to `info` level. The thing is, our inference of the machine learning task is very opinionated and might be suprising for some. Imagine passing in housing prices as integers not knowing you end up performing multiclass selection.
Still, I think the inference is good this way. We just have to make it very clear to the user.

tsfresh/feature_selection/feature_selector.py
jneuff: I'll change it to 'examines'.

tsfresh/feature_selection/feature_selector.py
jneuff: Will do.

jneuff:
> I have already proposed this right in the beginning, but I can not remember why we did not choose to get rid of the "rejected", which was always quite puzzling for me :-)

We had a lot of choices to make in the beginning. Not all of them were good :-D

tsfresh/feature_selection/selection.py
MaxBenChrist: I am in favor of the warning, we also warn the user in the impute function.

tsfresh/feature_selection/selection.py
MaxBenChrist: > But I thought about that too, we could run the Benjamini-Hochberg-Test on the minimal p-values from all classes. That way, the FDR will be respected globally instead of just per class.

Mathematically, I do not that that is the right way. I think, what you have to do is to combine all p_values and then do BH test. So if you have three classes, you combine 3*p (p=number of features) p-Values to one BH test. You keep the features, if at least one of its p-values got rejected

jneuff:
Finally I found time to implement all changes. I still have to verify that the notebooks work. I couldn't get the notebook tests run on my setup (will post the details as issue).

Here's what I changed:

- rename lots of stuff in the `feature_selection` module
- split up and refactored some tests
- implement the selection for one-vs-rest multiclassification
- change the robot execution failure example to return all labels (probably the notebook will fail)

I think we should mention the multiclassification selection somewhere in the docs. Still need to decide how and where. Ideas are welcome!

MaxBenChrist:
Awesome work @jneuff, I will have a look at it Sunday night + then trigger the v0.9.0 release.

tsfresh/feature_selection/relevance.py
```diff
@@ -0,0 +1,254 @@
...
+
+    In the same way we can state the hypothesis when the feature is binary
+
+        :math:`H_0 =  \\left( T_{\\text{feature}=1} = T_{\\text{feature}=0} \\right)`
+
+        :math:`H_1 = \\left( T_{\\text{feature}=1} \\neq T_{\\text{feature}=0} \\right)`
+
+    Here :math:`T` is the distribution of the target.
+
+    TODO: And for real valued?
```
MaxBenChrist: Have a look at the paper ;)

tests/feature_selection/test_relevance.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def test_restrict_ml_task_options(self, X, y_binary):
+        with pytest.raises(ValueError):
+            calculate_relevance_table(X, y_binary, ml_task='some_other_task')
+
+    def test_constant_feature_irrelevant(self, y_binary):
+        X = pd.DataFrame([1, 1, 1], columns=['feature_binary'])
+
+        relevance_table = calculate_relevance_table(X, y_binary)
+        assert "feature_binary" == relevance_table.index[0]
```
MaxBenChrist: personally, I am used to interpret the output of unit tests assert. 

We could have this here as well by making `TestCalculateRelevanceTable` a subclass of `unittest.TestCase` and then call `self.assert..`like in the other tests.

tests/feature_selection/test_significance_tests.py
```diff
@@ -0,0 +1,158 @@
...
+
+@pytest.fixture()
+def real_target_not_related():
+    return pd.Series(np.random.normal(0, 1, 250))
+
+
+class TestUnsignificant:
+    @pytest.fixture()
+    def minimal_p_value_for_unsignificant_features(self):
+        np.random.seed(seed=42)
```
MaxBenChrist: why do you set the seed here?

tsfresh/feature_selection/relevance.py
```diff
@@ -0,0 +1,254 @@
...
+        for label in y.unique():
+            tmp = _calculate_relevance_table_for_binary_target(
+                table_real, table_binary, X, (y == label),
+                test_for_binary_target_real_feature=test_for_binary_target_real_feature,
+                hypotheses_independent=hypotheses_independent, fdr_level=fdr_level, map_function=map_function
+            )
+            tables.append(tmp)
+        relevance_table = combine_relevance_tables(tables)
+    elif ml_task == 'regression':
+        _calculate_real_feature = partial(target_real_feature_real_test, y=y)
```
MaxBenChrist: in the case of the classification, we have a subfunction (`_calculate_relevance_table_for_binary_target`) in the case of the regression not. 

any reason for that?

tests/feature_selection/test_relevance.py
```diff
@@ -0,0 +1,133 @@
...
+
+        relevance_table = calculate_relevance_table(X, y_real, n_jobs=0)
+
+        assert 0.5 == relevance_table.loc['feature_binary'].p_value
+        assert 0.7 == relevance_table.loc['feature_real'].p_value
+        significance_test_feature_binary_mock.assert_called_once_with(X['feature_binary'], y=y_real)
+        significance_test_feature_real_mock.assert_called_once_with(X['feature_real'], y=y_real)
+
+
+class TestCombineRelevanceTables:
```
MaxBenChrist: don't you want to test if the relevance_tables are combined? so you need two relevance tables with different indices?

tsfresh/examples/robot_execution_failures.py
```diff
@@ -102,10 +102,7 @@ def load_robot_execution_failures():
             if line[0] not in ['\t', '\n']:
                 cur_id += 1
                 time = 0
-                if line.strip() == 'normal':
-                    id_to_target[cur_id] = 0
-                else:
-                    id_to_target[cur_id] = 1
+                id_to_target[cur_id] = line.strip()
```
MaxBenChrist: can you clarify that change?

tests/feature_selection/test_relevance.py
jneuff: I'm not sure there is an actual difference. `pytest` rewrites `assert` statements before executing them. See https://docs.pytest.org/en/latest/assert.html#assert for more details.
For me using the plain assert makes the tests much clearer.

tests/feature_selection/test_significance_tests.py
jneuff: Well spotted! I naively thought that the fixtures are executed in the order they appear as arguments. However after googling a bit, I found out that's not the case. I'll fix it. The solution should be that the fixtures relying on the seed to be set, should depend on a fixture that does that, thus guaranteeing the order of execution.

tsfresh/feature_selection/relevance.py
jneuff: I'd have created a more general function that would fit both cases, taking the significance tests as parameters. But as the `target_binary_feature_real_test` has an additional parameter I wanted to avoid the complexity of checking for that case.
However, reviewing it, I see that I could just pass in the significance tests as configured partials. Then I can remove the duplication.

tests/feature_selection/test_relevance.py
jneuff: I'm not sure what you mean here. The index of a relevance table are the feature names. So to have several of them combined, the indices have to be the same across the tables.

tsfresh/examples/robot_execution_failures.py
jneuff: Yes, here we return the class label. Before we only returned failure/no failure. As we are now capable of selection for multiclassification, the robot execution failure dataset can serve as an example for that.

MaxBenChrist: can you add a Boolean flag that controls if a binary or a multi classification target is returned?

tests/feature_selection/test_relevance.py
MaxBenChrist: Ok. I was thinking that the combine method is concatenating and not combining the results. 

The unit tests are fine then

tests/feature_selection/test_significance_tests.py
MaxBenChrist: But why do we have to set a seed? the calculation of the p-value should not depend on a random number...

tsfresh/examples/robot_execution_failures.py
jneuff: Good idea!

tests/feature_selection/test_significance_tests.py
jneuff: It's not for the p-value, but for the creation of the random features and targets.

tests/feature_selection/test_relevance.py
MaxBenChrist: I would be in favor of having a consistent way for the unit tests. I agree that the putests are not as clean but they have advantage of 

> pytest has support for showing the values of the most common subexpressions including calls, attributes, comparisons, and binary and unary operators. 

.. which we rely on in the other modules.

MaxBenChrist: julius convinced me in person about pytest :)

Issue #228, opened 2017-07-10
Bug: Calculation of Fourier coefficients is not independent from other Fourier coefficients


MaxBenChrist:
I was working on unit tests for the `fft_coefficient` feature calculator, when I stumbled about the following bug:

``` Python

i = np.linspace(-1, 1, num=100)
x = np.sin(i*np.pi)

for max_lag in range(1, 10):
    ...:     print(np.fft.rfft(x, max_lag).real)
# will print
[ -1.22464680e-16]
[-0.06342392  0.06342392]
[-0.19001637  0.09500819]
[-0.37926762  0.12659245  0.12608271]
[-0.6304156   0.15831485  0.15689295]
[-0.94244905  0.19039278  0.18734766  0.18696817]
[-1.31411151  0.22311752  0.21741968  0.21651855]
[-1.74390642  0.25685141  0.24710692  0.24544456  0.24510063]
[-2.23010315  0.29202479  0.27642732  0.27368103  0.27291845]
```

So what is happening here? The value of a fft coefficient depends on the maxlag that is used to calculate the fft.

The bad line is here: https://github.com/blue-yonder/tsfresh/blob/master/tsfresh/feature_extraction/feature_calculators.py#L742

Why is this bad? 

This means that the value of e.g. the Fourier coefficient for coefficient==3 depends on the configuration for the other features. This breaks our convention that the value of a feature only depends on his configuration

How can we fix this? I see multiple solutions

- make the `fft_coefficient` a "simple" function calculator instead of "combine" (this is computational expensive because it will calculate one fft for EVERY coefficient..)
- add an additional parameter `maxlag` (not user friendly)
- use length of individual time series as maxlag (will Also increase the computational complexity)

I tend to include the maxlag parameter as I do not want to increase the runtime, speed is everything ;)

It would be very helpful to have somebody with a background in signal processing to comment on this. I had a lecture on Fourier transformations but that was 5 years ago, so my knowledge is limited

nikhaas:
I give my 2cents to this issue, although I am not an expert in signal processing. Tbh, my following conclusions are drawn from ~1h of research on Fast Fourier Transformation (FFT). Please correct me if I am mistaken.

[...]

PR #234, opened 2017-07-11
Fix the trunction bug for fft_coefficients

MaxBenChrist:
I ditched the "max_lag" as it truncated the input.

I added some unit tests and updated the documentation

See issue #228 for a detailed discussion

### DIFF
--- a/tsfresh/feature_extraction/feature_calculators.py
+++ b/tsfresh/feature_extraction/feature_calculators.py
@@ -737,17 +737,13 @@ def fft_coefficient(x, param):
     :return type: pandas.Series
     """
 
-    coefficients = set([config["coeff"] for config in param])
-    for coeff in coefficients:
-        if coeff < 0:
-            raise ValueError("Coefficients must be positive or zero.")
+    assert min([config["coeff"] for config in param]) >= 0, "Coefficients must be positive or zero."
+    assert set([config["attr"] for config in param]) == set("imag", "real"), 'Attribute must be "real" or "imag"'
 
-    maximum_coefficient = max(max(coefficients), 1)
-    fft = np.fft.rfft(x, min(len(x), 2 * maximum_coefficient))
+    fft = np.fft.rfft(x)
 
-    res = [fft[q] if q < len(fft) else 0 for q in coefficients]
-    res = [r.real if isinstance(r, complex) else r for r in res]
-    index = ["coeff_{}".format(q) for q in coefficients]
+    res = [getattr(fft[config["coeff"]], config["attr"]) for config in param]
+    index = ["coeff_{}__attr_{}".format(config["coeff"], config["attr"]) for config in param]
     return zip(index, res)
 
 
--- a/tsfresh/feature_extraction/feature_calculators.py
+++ b/tsfresh/feature_extraction/feature_calculators.py
@@ -742,7 +742,7 @@ def fft_coefficient(x, param):
 
     fft = np.fft.rfft(x)
 
-    res = [getattr(fft[config["coeff"]], config["attr"]) for config in param]
+    res = [getattr(fft[config["coeff"]], config["attr"]) if config["coeff"] < len(fft) else np.NaN for config in param]
     index = ['coeff_{}__attr_"{}"'.format(config["coeff"], config["attr"]) for config in param]
     return zip(index, res)
 
--- a/tsfresh/feature_extraction/feature_calculators.py
+++ b/tsfresh/feature_extraction/feature_calculators.py
@@ -743,7 +743,7 @@ def fft_coefficient(x, param):
     fft = np.fft.rfft(x)
 
     res = [getattr(fft[config["coeff"]], config["attr"]) for config in param]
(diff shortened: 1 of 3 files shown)


tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -726,25 +726,24 @@ def fft_coefficient(x, param):
...
+    assert set([config["attr"] for config in param]) <= set(["imag", "real"]), 'Attribute must be "real" or "imag"'
 
-    maximum_coefficient = max(max(coefficients), 1)
-    fft = np.fft.rfft(x, min(len(x), 2 * maximum_coefficient))
+    fft = np.fft.rfft(x)
 
-    res = [fft[q] if q < len(fft) else 0 for q in coefficients]
-    res = [r.real if isinstance(r, complex) else r for r in res]
-    index = ["coeff_{}".format(q) for q in coefficients]
+    res = [getattr(fft[config["coeff"]], config["attr"]) if config["coeff"] < len(fft) else np.NaN for config in param]
```
jneuff: So now we return `NaN` instead of 0, if the coefficient does not exist? Seems reasonable!

tsfresh/feature_extraction/settings.py
```diff
@@ -117,7 +119,7 @@ def __init__(self):
             "ar_coefficient": [{"coeff": coeff, "k": k} for coeff in range(5) for k in [10]],
             "mean_abs_change_quantiles": [{"ql": ql, "qh": qh}
                                           for ql in [0., .2, .4, .6, .8] for qh in [.2, .4, .6, .8, 1.]],
-            "fft_coefficient": [{"coeff": coeff} for coeff in range(0, 10)],
+            "fft_coefficient": [{"coeff": k, "attr": a} for a, k in product(["real", "imag"], range(100))],
```
jneuff: That would be 20x as many features from this calculator. That seems quite a lot for the default settings. How likely are they gonna be `NaN` anyway?

MaxBenChrist: Those features come for free! We calculate the coefficients up to the length of the time series.

They will be np.NaN if and only if  `len(x) < coeff`.

MaxBenChrist:
About the number of features. Computationswise, there is nearly no difference if we extract 100 or 1000 dft coefficients.

tsfresh/feature_extraction/settings.py
jneuff: Yes, I got that. But I don't have any experience working with the Fourier Transform. What I mean is this: For the average time series (OK, there is no average time series..), how many FFT coefficients are we likely to find? So, if we take a 100 and most of the time most of them are NaN, we could also just take 10.

jneuff:
> About the number of features. Computationswise, there is nearly no difference if we extract 100 or 1000 dft coefficients.

That's not what's bothering me. I mean, do we want to blow up the resulting DataFrame like that?

tsfresh/feature_extraction/settings.py
MaxBenChrist: It only depends on the length of the time series. For a time series of length 10, you will get exactly 10 dft coefficients. For one of length 100, exactly 100

MaxBenChrist: Sorry, that is not correct. If the time series is real and its length is `n`, then the length of the fft transformation is (n/2)+1 if n is even. If n is odd, the length is (n+1)/2. (due to symmetries in the calculation of the dft). you have to compare `coeff` to the length of the fft, not to the length of the fft

jneuff: Cool, that means we could derive the number of coefficients to return as features from the length of the time series. I think that would be better than to return lots of `NaN` or skipping possibly meaningful coefficients. But of course it would be problematic to vary the features depending on the length of the time series, as the length could differ between samples. This would be especially problematic with regards to train / test splits.

MaxBenChrist: yes. I had the same idea. get the maximal length of the time series, and then use that to calculate max coeff. I thought long about this, but I do not think there is a good way to realize that.

We could add a function in the future that infers the extraction settings from a given dataset.

But, I am reading in the details of the DFT. I am not sure yet, but from what I remember the higher order coefficients correspond to the energy in higher frequencies. As most time series suffer from noise, Im expect higher order coefficients not to be too relevant.

jneuff:
So the discussion is just about the number of coefficients to return with the default settings. If these coefficients turn out to be irrelevant, they are not selected anyway. The only concern would be the size of the intermediate DataFrame and excess time spending on imputation (only if the time series lengths differ per sample).

I think we can merge this PR. We could cut down the number of coefficients later, if actual problems arise.

MaxBenChrist:
> So the discussion is just about the number of coefficients to return with the default settings. If these coefficients turn out to be irrelevant, they are not selected anyway. The only concern would be the size of the intermediate DataFrame and excess time spending on imputation (only if the time series lengths differ per sample).

> I think we can merge this PR. We could cut down the number of coefficients later, if actual problems arise.

alright. Thanks for reviewing it @jneuff

PR #235, opened 2017-07-11
added notebook to inspect dft features

MaxBenChrist:


MaxBenChrist:
Not merging. at this moment. @jneuff this is what I was referring to, the higher order dft coefficients correspond to higher frequencies. In the notebook it can be used to filter out the noise. 

I will comment that notebook a little bit more

MaxBenChrist:
Can be merged

PR #236, opened 2017-07-12
update human activity notebook with new impute

Baisang:
hi, i noticed in the latest version of tsfresh you use the new method of passing in your impute function to `extract_features` instead of setting it in the `extraction_settings`. So I updated the notebook so others would see this (it tripped me up for a bit, as I was wondering why my data wasn't being imputed).

let me know if there's anything else i need to do for this PR.


MaxBenChrist:
Ah, we seem to haved overlooked that. Thanks @Baisang

PR #243, opened 2017-07-14
Rewrite the DFT notebook

nikhaas:
* theoretical background for the discrete fourier transform
* Analysis of a superposition of periodic signals
* analysis of noise, polynomial and truncated periodic signals
* conclusion with the main points at the end
* after submission, an issue will be opened to incorporate
normalization of FFT coefficients in tsfresh


MaxBenChrist:
Great notebook @nikhaas.

MaxBenChrist:
In the 13th cell, you are missing to plot the raw signal

You are tackling a lot of points in your notebook. The images are great to get a feeling for the dft, but your conclusions cell contain a few things that raise more questions than they answer.

> If you are interested in the symmetry of the DFT and the important but unmentioned Nyquist Frequency, you can find many valuable resources in the wild world wide web

add that to further research/reading

> White noise can be seen on all "frequencies": 

from my experience with industrial or financial time series, the noise is most often in the higher frequencies. 

I also would spent that much time on explaining the symmetry, for the user it is an implementation detail

nikhaas:
> In the 13th cell, you are missing to plot the raw signal

I corrected that

> In general, I would spent that much time on explaining the symmetry, for the user it is an implementation detail. add that point to further research/reading

This part is truncated and added below the header "Further Research".

> from my experience with industrial or financial time series, the noise is most often in the higher frequencies.

I don't think so. It can be seen on all frequencies and, according to [this post on SO](https://dsp.stackexchange.com/questions/8418/what-is-the-phase-and-magnitude-response-of-white-noise):
> The *expected* magnitude response of white noise is flat
> In fact, the Fourier transform of white noise is... white noise!

(I did not see that this holds only for the *expected* magnitude). To your experience from industrial applications: Is it possible that the gaussian noise was covered by deterministic signals in the lower frequencies? So it seems to you that the noise does only appear in the higher frequencies? I plotted the DFT of gaussian noise in the notebook, and i can see no pattern in this example. However, one could derive this analytically, but this might be beyond the scope of this PR / *tsfresh* in general.

> delete that. every feature can be tuned by data scientists

Done


```python
import numpy as np

x1 = np.linspace(0, 1, 5)
x2 = np.linspace(0, 1, 10)

np.fft.rfft(x1) 
# returns array([ 2.500+0.j        , -0.625+0.8602387j , -0.625+0.20307481j])

np.fft.rfft(x2) that
# returns array([ 5.00000000+0.j        , -0.55555556+1.70982419j, 
# -0.55555556+0.76465662j, -0.55555556+0.40363474j,
# -0.55555556+0.18051094j, -0.55555556+0.j        ])
```
> Same signal.

I do not agree with you. At least for the DFT, these are not equal. You can interpret the signals as
- `x2` is `x1` with doubled sampling rate ( e.g. 5Hz vs 10Hz) OR
- with both sampling rates being equal (5Hz) you observed `x2` for 2 seconds and `x1` for 1 second, and the signals are different

However, if you do normalization with `n`, you can at least compare the 0th coefficient, which is the sum of the signal values, being a somehow trivial feature (and possibly duplicated by *tsfresh*).

```python
np.abs(np.fft.rfft(x1) / 5)
# Out: array([ 0.5       ,  0.2126627 ,  0.13143278])
np.abs(np.abs(np.fft.rfft(x2) / 10))
# Out: array([ 0.5       ,  0.17978155,  0.09451676,  0.06867044,  0.05841457,  0.05555556])
```
but sadly, using `norm="ortho"` (which is 1/sqrt(n)), there is no exploitable information

```python
np.abs(np.fft.rfft(x1, norm="ortho"))
# Out: array([ 1.11803399,  0.47552826,  0.29389263])
np.abs(np.fft.rfft(x2, norm="ortho"))
# Out: array([ 1.58113883,  0.56851919,  0.29888823,  0.21715501,  0.18472308,
#         0.17568209])
```
at least, normalization is valuable if you compare time series with the same length:

```python
y1 = np.linspace(0,1,5)

y2 = np.linspace(0,2,5)

np.fft.rfft(y1, norm="ortho")
# Out: array([ 1.11803399+0.j        , -0.27950850+0.38471044j,
#        -0.27950850+0.09081782j])
np.fft.rfft(y2, norm="ortho")
# Out: array([ 2.23606798+0.j        , -0.55901699+0.76942088j,
#        -0.55901699+0.18163563j])
```

> Also, you are missing one important insight, the first real valued dft coefficients is the sum of the signal.

I added this to the conclusion.

nikhaas:
I commented in #247 on the non-passing `nbconvert` tests, which might be related to notebook-metadata kernel specifications.

MaxBenChrist:
> (I did not see that this holds only for the expected magnitude). To your experience from industrial applications: Is it possible that the gaussian noise was covered by deterministic signals in the lower frequencies? So it seems to you that the noise does only appear in the higher frequencies? I plotted the DFT of gaussian noise in the notebook, and i can see no pattern in this example. However, one could derive this analytically, but this might be beyond the scope of this PR / tsfresh in general.

Just start playing with real data and you will see what I mean. the following snippets pulls some stock data.

``` Python

import pandas_datareader.data as web
import datetime
import matplotlib.pylab as plt
import seaborn as sns
import numpy as np

start = datetime.datetime(2016, 1, 1)
end = datetime.datetime(2016, 3, 27)
f = web.DataReader("F", 'google', start, end)

def plot_x_and_rfft(x):

    fft = np.fft.rfft(x.values)
    plt.figure(figsize=(12, 4))
    
    plt.subplot(311)
    plt.title("Time Series")
    plt.plot(x)
    
    plt.subplot(312)
    plt.plot(fft.real, label="real", marker="o")
    plt.plot(fft.imag, label="imag", marker="o")
    plt.title("Real and imaginary part of DFT coefficients A(k)")
    plt.legend()

    plt.subplot(313)
    plt.plot(fft.real[1:], label="real", marker="o")
    plt.plot(fft.imag[1:], label="imag", marker="o")
    plt.title("Real and imaginary part of DFT coefficients A(k), without first")
    plt.legend()
    
    plt.tight_layout()
    plt.show()

plot_x_and_rfft(f["High"])
```

MaxBenChrist:
> at least, normalization is valuable if you compare time series with the same length:

So, you are saying, that normalization is useful, but only if length is fixed. Then you normalize by 1/length or 1/sqrt(length.), so the normalization will be equal for all ts of same length => is just a constant factor, right?

> I do not agree with you. At least for the DFT, these are not equal. You can interpret the signals as
> x2 is x1 with doubled sampling rate ( e.g. 5Hz vs 10Hz) OR
> with both sampling rates being equal (5Hz) you observed x2 for 2 seconds and x1 for 1 second, and the signals are different

I was talking about the first case. But, as you also pointed out, the dft are totally different

MaxBenChrist:
@nikhaas, Can you pull the master into this branch? Afterwards the build will pass and we can merge this pr!

nikhaas:
Done. Maybe I should not have merged the master in the first place. What's your policy here? Always merge the newest master?

Btw, I used `sns.set_palette('colorblind')` in the code. This might be good for people that have problems distinguishing the colors red and green.

nikhaas:
The decrease in coverage does not depend on the notebook. Shall I Merge the newest master again? Or roll back all merged master branches to the last "all tests passing" state?

MaxBenChrist:
> The decrease in coverage does not depend on the notebook. Shall I Merge the newest master again? > Or roll back all merged master branches to the last "all tests passing" state?

No need for that. The coverage drop is unrelated. 

I am happily merging this. only one last thing

I think that 

> If one is interested to compare amplitudes among time series with different lengths, one needs to normalize the FFT coefficients by time series length

is misleading. Can you delete that?

Further, the point 

> White noise can be seen on all "frequencies"

Is misleading in my opinion. Maybe just delete it, no better phrase comes to my mind at the moment.

MaxBenChrist:
Awesome. Welcome to the selected circle of tsfresh contributors ;-) @nikhaas

PR #248, opened 2017-07-16
from mean_abs_Change_quantiles made general class of features change_quantiles

MaxBenChrist:


MaxBenChrist:
I also fixed the not directly related bug in 639bf717fc3a2bf6149e95021baed6a94c391200

tsfresh/utilities/string_manipulation.py
```diff
@@ -48,10 +48,19 @@ def convert_to_output_format(param):
...
     :param param: The dictionary of parameters to write out
     :type param: dict
 
     :return: The string of parsed parameters
     :rtype: str
     """
-    return "__".join(str(key) + "_" + str(param[key]) for key in sorted(param))
+
+    def add_paranthesis_if_string_value(x):
+        if isinstance(x, numbers.Number):
```
nils-braun: Don't you want to turn this the other way round: only add " on str (or basestring)? Because we may also have lists and tuples and stuff, where we do not want to have parenthesis.

Also, it is par*e*nthesis, not par*a*nthesis ;-)

MaxBenChrist: Yeah you are right, well spotted!

MaxBenChrist: corrected it

Issue #200, opened 2017-05-13
Can anyone provide a example of a regression problem?


NickYi1990:
I really do not know how to use this package to solve a regression problem, like predict a company in which day earned how many money.

But i think this package must can be used to solve this kind of problem, is there any example?

mynameisvinn:
this package "contains many feature extraction methods" that characterize time series data. it wont, however, fit regression models (or any other statistical models) for prediction.

MaxBenChrist:
Vinn is right, tsfresh is just used to do the preprocessing. It will map the time series into the feature space, on this feature space you can use regression models from other libraries, e.g. Tensorflow or sklearn.

I will let this issue open, We need some notebooks showing how to use tsfresh for time series forecasting g

NickYi1990:
thanks!

yenicelik:
Has there been any progress on this ever since the last comment?

PR #249, opened 2017-07-16
Notebook rolling

MaxBenChrist:
It is not ready to be merged. 

However, I would like to get your feedback on this. What do you think about that `make_forecasting_frame` method?

The  `notebooks/timeseries_forecasting_basic_example.ipynb` notebook can be used to play a little bit with the method.


nils-braun:
I will try to have a look into this tomorrow...

TKlerx:
Hi @MaxBenChrist,

sorry that I did not build this as (more or less) promised.

I created a notebook that follows your idea but can also handle two signals at a time and create the forecasting frame. It's still a draft.

Are you interested in it? If yes, where shall I put it?

MaxBenChrist:
Hi @TKlerx 

have you changed much in my notebook?

TKlerx:
Hi,

no, not much, just a bit. And I added some assertions for "testing".

MaxBenChrist:
> But I would be happy to help/support ;)

Awesome. I do not have much time at the moment to test the forecasting API. feedback would be really helpful

TKlerx:
I am struggling with the input format of dict or data frame. Creating the windows is already done in my notebook.
I will try to create a PR for the master branch that contains the rolling functionality from the notebooks.

trevorwelch:
Pulled this down and got it running. This is a workable foundation for forecasting problems. However, the current version is not usable for classification problems.

Some thoughts:

1) `extract_features` drops a row from the input `x`. 
In the notebook, for example: x starts out with shape `(200,)`.
After `make_forecasting_frame` and `extract_features` it has shape `(199, 275)`.
Rather than dropping the rows, it should maintain the original indices and set empty values as a result of lagged data frames  to `NaN` like `.rolling` in pandas. Otherwise, it is challenging to work around for most machine learning problems where `X` is your data and `y` is your ground truth/target. Assume you are working from a dataset with 400 rows and `X` are your features, `y` is your ground truth for classification. If your `X` ends up with 399 rows after feature extraction and your `y` is at 400, this is a problem. 

2) It seems like it would be desirable if `make_forecasting_frame` + `extract_features` would generate the best features for `X` with respect to `y`, as opposed to assuming that its a forecasting problem. Of course, in a forecasting/regression problem, the best features are usually those that coincide with accurately predicting the next window of the time-series, so you don't really need a separate `y`. In classification, obviously you need a separate `y` target that your model is trying to predict. One way to achieve this with `tsfresh` would be: add a forecasting flag that if set, uses `n` number of following time-series rows to generate a `y`. If forecasting is 0, then tsfresh expects to receive an explicit `y` data frame or array. This seems like it would be the most `sklearn`-like approach, and would be flexible to both forecasting and classification problems. Probably in this case it should be called `make_rolling_frame` or `make_lag_frame` as to be general to either forecasting or classification.

MaxBenChrist:
This pr is nearly ready to merge. I only have to adapt the id column of the section in the docs and add a reference to the `make_forecasting_frame`.

 @jneuff or @nils-braun could have a look at this pr?

MaxBenChrist:
@TKlerx 

With what are you struggling? have you had a look at the notebook `notebooks/timeseries_forecasting_basic_example.ipynb` ?

MaxBenChrist:
@trevorwelch: I thought about your remark `1.`. I feel that dropping that the first column is more user friendly.

TKlerx:
@MaxBenChrist I was struggling with implementing the forecasting frame for more than one feature in the right way at the right place and how to check for dict or data frame (and the right format of the data frame).
Unfortunately, I did not have the time to implement anything the last few days.

Maybe, I will find some time in the near future.

MaxBenChrist:
> @MaxBenChrist I was struggling with implementing the forecasting frame for more than one feature in the right way at the right place and how to check for dict or data frame (and the right format of the data frame).
Unfortunately, I did not have the time to implement anything the last few days.

You mean creating the forecasting frames for multiple time series? Yes, that is not supported in the current version. But even though it would be easy to add it, it does not fit the forecasting task.

I cannot think of an application where you would train a multivariate forecasting model.

TKlerx:
What about this application?

If you want to perform predictive maintenance and thus, estimate the number of days until breakdown, you may need to incorporate multiple signals and you would predict a single value (days until breakdown).

MaxBenChrist:
> What about this application?
> If you want to perform predictive maintenance and thus, estimate the number of days until breakdown, you may need to incorporate multiple signals and you would predict a single value (days until breakdown).

Ok, good point, but I think that this is not a time series forecasting task per se. lets say you use 10 different temperature sensors to predict failures or remaining lifetime of the machine. It is true that  a rolling mechanism is needed to construct the model. However, this is not a time series forecasting task because you do not forecast the same time series (the temperature) but an external variable (the life time).

It is more of a regression instead of a time series classification task.

TKlerx:
> Ok, good point, but I think that this is not a time series forecasting task per se. lets say you use 10 different temperature sensors to predict failures or remaining lifetime of the machine. It is true that a rolling mechanism is needed to construct the model. However, this is not a time series forecasting task because you do not forecast the same time series (the temperature) but an external variable (the life time).
It is more of a regression instead of a time series classification task.

That's true ;)

If I find the time, I will implement it and create a PR. Suggestions how to call it and where to put it?

MaxBenChrist:
@TKlerx 

You can just call `make_forecasting_frame` for every time series and then concat the resulting dfs. should be only a few lines of code + unit tests.
I would put it into utilities.

Looking forward to the pr :)

MaxBenChrist:
ready for merge :)

MaxBenChrist:
Solves issues #200 and #174

docs/text/forecasting.rst
```diff
@@ -179,91 +190,126 @@ If you set rolling to -1, you end up with features for the time series, rolled i
...
-| 1,shift=0 | t2   | 2  | 6  |
+| t4        | t4   | 4  | 8  |
 +-----------+------+----+----+
-| 1,shift=0 | t3   | 3  | 7  |
+
 +-----------+------+----+----+
-| 1,shift=0 | t4   | 4  | 8  |
+| id        | time | x  | y  |
++===========+======+====+====+
+| t8        | t8   | 10 | 12 |
```
nils-braun: isn't there one missing?

tests/utilities/test_dataframe_functions.py
```diff
@@ -562,3 +611,36 @@ def test_no_finite_values_yields_0(self):
...
+        expected_y = pd.Series(data=[1, 2, 3], index=pd.DatetimeIndex(["2011-01-01 01:00:00", "2011-01-01 02:00:00",
+                                                                       "2011-01-01 03:00:00"]), name="val")
+        expected_df = pd.DataFrame({"id": pd.DatetimeIndex(["2011-01-01 01:00:00", "2011-01-01 02:00:00",
+                                                            "2011-01-01 03:00:00"]),
+                                    "kind": ["test"]*3, "val": [0., 1., 2.],
+                                    "time": pd.DatetimeIndex(["2011-01-01 00:00:00", "2011-01-01 01:00:00",
+                                                              "2011-01-01 02:00:00"])
+                                    })
+        assert_frame_equal(df.sort_index(axis=1), expected_df.sort_index(axis=1))
+        assert_series_equal(y, expected_y)
```
nils-braun: are the three test you are giving really our typical use case?

tsfresh/utilities/dataframe_functions.py
```diff
@@ -409,27 +408,96 @@ def roll_time_series(df_or_dict, column_id, column_sort, column_kind, rolling_di
...
         return df_temp.dropna()
 
-    return pd.concat([roll_out_time_series(time_shift) for time_shift in range_of_shifts],
-                     ignore_index=True)
+    df_shift = pd.concat([roll_out_time_series(time_shift) for time_shift in range_of_shifts], ignore_index=True)
+
+    return df_shift.sort_values(by=[column_id, column_sort])
+
+
+def make_forecasting_frame(x, kind, max_timeshift, rolling_direction):
```
nils-braun: what if one wants to process data frames with this?

MaxBenChrist: Then you could use the rolling function, couldn't you? 

Also, a typical forecasting tasks involves a singular time series, see for example the google stock notebook. Our container format was developed for multiple time series (of different kinds and ids), so I feel that there is no need to support it here

tests/utilities/test_dataframe_functions.py
MaxBenChrist: why not?

docs/text/forecasting.rst
MaxBenChrist: you are right. will add it

tsfresh/utilities/dataframe_functions.py
nils-braun: Yes, this answers more or less also my question above - I think I was more stuck into the rolling use case. You are right!

docs/text/forecasting.rst
nils-braun: Ah sorry, I have shifted while reading. This is correct.

nils-braun:
Ok, this is now also ready to merge from my side

MaxBenChrist:
> Ok, this is now also ready to merge from my side

thx for the review, will merge once travis is done :)

trevorwelch:
Thanks for this! 

So, I just wanted to clarify -- it appears that the rolling method drops the first row of the time series, is that correct? This seems fine for most Regression tasks, but is definitely not a desirable result for a classification task. It would probably be good to specify that somewhere so people aren't wondering why there's a shape mismatch if they try to use the rolling methods prior to feeding the augmented `X` and `y` to a classifier :)

nils-braun:
Concerning the deletion of the first row: the new function for making forecasts ist just intended to be used in these typical forecasting scenarios. If you want to do classification or regression of external features, there is still the roll function in the utilities, that can be used for that (actually, the forecast function is just a wrapper around the rolling function plus deleting the first row...)

PR #252, opened 2017-07-19
Added meaningful feature names to robot failure example

kempa-liehr:
Meaningful feature names are useful in order to explain the data mining aspect of refresh


MaxBenChrist:
Are the names from their documentation? 

Can you please also correct the naming of the sensors in the following phrase

> The data consists of timeseries for 88 robots (id 1 - 88). For each robot, each timepoint (time) contains datapoints from 6 sensors (a - f) leading up to a success or failure (y).

it still uses a-f instead of F_1, F_2

bytheway, what does F_ and T_ stand for?

Finally, can you suppress the warnings by using parameter `show_warnings=True`

Thx

kempa-liehr:
Thanks for the review!

Yes, the column names are from the documentation.

MaxBenChrist:
Can you adapt the unit tests? They still test for "a", "b" etc.

kempa-liehr:
Of course.

MaxBenChrist:
Have you merged the newest master in this branch?

kempa-liehr:
Yes, I've merged the newest master in this branch. There seems to be a problem with nbconvert and Python 2.7.

MaxBenChrist:
Awesome

Issue #247, opened 2017-07-16
Test ipython notebooks for python3


MaxBenChrist:
I had to turn off the testing of python notebooks for py3. 

It would be good to also check the notebooks unter python3

nikhaas:
Why did you need to turn off the py3 testing? This might be related to my experience.

My [PR for a new notebook produced an error](https://github.com/blue-yonder/tsfresh/pull/243). This is because I use Anaconda with it's notebook extensions - I can choose an Anaconda Kernel among the "normal" python2 and python3 kernels. This information is stored in the metadata which is located at the bottom of the notebook. [Link to travis error log](https://travis-ci.org/blue-yonder/tsfresh/jobs/254186231#L984)

This is an example notebook metadata spec from [robot_failure_example.ipynb](https://raw.githubusercontent.com/blue-yonder/tsfresh/master/notebooks/robot_failure_example.ipynb):
```json
"metadata": {
  "kernelspec": {
   "display_name": "Python 2",
   "language": "python",
   "name": "python2"
  },
  "language_info": {
   "codemirror_mode": {
    "name": "ipython",
    "version": 2
   },
   "file_extension": ".py",
   "mimetype": "text/x-python",
   "name": "python",
   "nbconvert_exporter": "python",
   "pygments_lexer": "ipython2",
   "version": "2.7.11"
  }
 }
```
A local execution of the test suite yielded no success either. No notebook passed the test. This might be because my "Standard" Python2 environment (I am on Linux) does not have all the requirements, but `nbconvert` seems to use this environment.

It would be nice if one could specify which kernel to use for notebook execution via `nbconvert` during testing.

You might want to overwrite the metadataspecification in the Jupyter Notebook. This can be done temporarily by a flag. You can modify [test_notebooks.py](https://github.com/blue-yonder/tsfresh/blob/master/tests/test_notebooks.py) with the `"--ExecutePreprocessor.kernel_name=python"` flag:

```python
def _notebook_run(path):
    """
    Execute a singular ipython notebook at path and returns the cells and outputs

    :returns (parsed nb object, execution errors)
    """

    dirname, _ = os.path.split(path)

    with tempfile.NamedTemporaryFile(suffix=".ipynb") as fout:
        args = ["jupyter", "nbconvert",
                "--to", "notebook", "--execute",
                "--ExecutePreprocessor.timeout=600",
                "--ExecutePreprocessor.kernel_name=python",  ##!! New !!
                "--output", fout.name, path
[...]

PR #258, opened 2017-07-21
Py3nbtests

MaxBenChrist:


MaxBenChrist:
Locally, this passes with both python2 (2.7.13)  and python3 (3.6.2). 

The idea is that we specify the kernel when executing the notebooks as discussed in #247. This will override the kernel specified in the metadata in the header of the python notebook.

This should also work with anaconda installations as we specify `ipython==5.3.0` in `/test-requirements.txt`, the ipython package should contain the python2 and python3 kernels.

MaxBenChrist:
Well, it looks like the temp file is a bytes and not a string object in 3.5.2....

MaxBenChrist:
Finally!

PR #267, opened 2017-07-25
add notebook with principal component analysis, closes #259

### LINKED ISSUE #259: Include Principal Component Analysis in tsfresh
I did a PCA analysis after the feature selection process of tsfresh and I think it is a good idea to implement this code also in the tsfresh project.
I created a notebook file for that purpose. Because I cannot upload .ipynb files here, have a look at https://github.com/smueller18/tsfresh/blob/pca/notebooks/pca.ipynb. My suggestion is to implement the class PCAForPandas from cell [1] somewhere in the tsfresh project and to include the notebook without cell [1] for the notebook collection.
I can also include the class if you say me where to place. Documentation and testing is not included so far and has to be done before merging.

The notebook explains how to implement #42 .

smueller18:
as discussed in #259


MaxBenChrist:
Thx @smueller18 .

Two remarks:

- I think it is called "Principal Component Analysis on past features" instead of "Principal Component Analysis of past features"
- the notion of past and future is catchy, but in machine learning those are more often called train and test set, maybe you can add that to the headlines?

smueller18:
Will change this after my vacation.

smueller18:
remarks done

MaxBenChrist:
looks good. What I am missing is a cell at the top that explains the purpose of the notebook. 

What do you want to achieve with that notebook?

MaxBenChrist:
I will comment the notebook by myself

PR #268, opened 2017-07-25
New feature extractor as discussed, energy ratio by chunks

Ezekiel-Kruglick:
Added the energy_ratio_by_chunks feature extractor, as well as relevant parameter definitions and tests.

I tried to mimic existing format. This made the input param a little odd looking - glad to change it but wanted to stay with project format on first pass. Same for tests, usually I would do many short tests but went with one long one as per the other tests.


Ezekiel-Kruglick:
I see the errors. I cannot replicate the failure of the notebook, when I run it on my machine the notebook works fine. Preumably PerColumnImputerTestCase::test_NAN_preset_contains_more_columns_than_dataframe_to_fit  has something to do with the NaN-proofing I did, I will have to look into that. Not sure about the python 3.5.2 stuff, I will have to look into that too - I run 2.7

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -1504,3 +1504,42 @@ def agg_linear_trend(x, param):
...
+    res_data = []
+    res_index = []
+    N = int(param[0]["chunk_count"])
+
+    if N <= 0:
+        return "energy_ratio_input_invalid", 0.0
+
+    full_series_energy = np.nansum(x**2)
+    segment_length = len(x)//N
+    for i in range(N):
```
MaxBenChrist: I feel that we should get rid of the loop and try to vectorize this.

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -1504,3 +1504,42 @@ def agg_linear_trend(x, param):
...
+        start = i*segment_length
+        end = (i+1)*segment_length
+        res_data.append(np.nansum(x[start:end]**2.0)/full_series_energy)
```
MaxBenChrist: i think that the `append` operation is costly because it creates a new list in every loop iteration

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -1504,3 +1504,42 @@ def agg_linear_trend(x, param):
...
+    """
+
+    res_data = []
+    res_index = []
+    N = int(param[0]["chunk_count"])
+
+    if N <= 0:
+        return "energy_ratio_input_invalid", 0.0
+
+    full_series_energy = np.nansum(x**2)
```
MaxBenChrist: No need to use `nansum`, we assume that the time series data is clean

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -1504,3 +1504,42 @@ def agg_linear_trend(x, param):
...
+    :type chunk_count: integer
+    :return: the different feature values
+    :return type: tuples of feature name, feature value
+    """
+
+    res_data = []
+    res_index = []
+    N = int(param[0]["chunk_count"])
+
+    if N <= 0:
```
MaxBenChrist: it is better to use `assert` here.

Later users can start the script with `python -o ` which will optimize all assert calls out, so the program will even run faster

MaxBenChrist:
Well done @Ezekiel-Kruglick. I am really interested in seeing more features from you.

Only one thing. In its current version, the `energy_ratio_by_chunks` returns multiple features. We have a paradigm in tsfresh that for every parameter combination exactly one feature is returned. 

That means that for your feature, we need an additional parameter that denotes for which bin you are returning the relative energy. In the current version, you are returning `N` features.

tsfresh/feature_extraction/feature_calculators.py
Ezekiel-Kruglick: Will do, no problem

tsfresh/feature_extraction/feature_calculators.py
Ezekiel-Kruglick: Okay, will change that

tsfresh/feature_extraction/feature_calculators.py
Ezekiel-Kruglick: I actually modeled the loop on some of the others (especially those that bring in parameter combos). I'll have to think about this  one, but I think we could perhaps vectorize with a running average and doing sums and differences for the results, although somewhere there has to be result code that runs N times...

tsfresh/feature_extraction/feature_calculators.py
Ezekiel-Kruglick: This part is exactly like the feature before it :) 

I went and did some research and didn't see many obvious faster approaches in the standard library. If collections is already a requirement (I'm on mobile, can check later) we could use deque which does fast appends. Or we could put them in a dict then convert the dict to list at the end?

Ezekiel-Kruglick:
"So I propose to use two parameters i and N. then the feature returns the relative energy for the ith out of N bins." - Yup I can do that. See, learning these paradigm conventions for a library is always part of the first contribution. I'll make all the changes above then see if I can run down those test errors. I haven't see the approach here of running jupyter notebooks and checking the exit code as a test, I have managed to reproduce the failure of the test now but when I run the jupyter notebook in question in interactive mode I can run all cells without any errors. If you have any thoughts on that I welcome then. Otherwise I'll get these changes in and try to see what I can do about the notebook test.

tsfresh/feature_extraction/feature_calculators.py
MaxBenChrist: Well actually, Python appends is a amoritized 0(1) costs (see https://wiki.python.org/moin/TimeComplexity). 

So, this should not be much of a problem here.

MaxBenChrist:
> Otherwise I'll get these changes in and try to see what I can do about the notebook test.

It could be related to your feature calculator returning multiple features for one setting configuration. I would start by only returning one value and then see if the problem with the notebooks persist.

Ezekiel-Kruglick:
Changes should be complete. Passes all tests on my machine (Python 2.7) and I believe I corrected the one Python 3 compatibility issue Travis-CI found.

MaxBenChrist:
Awesome @Ezekiel-Kruglick.

I fixed the merge conflict and ramped up the number of bins to 10.

If you have some more ideas for features, we happily receive your prs or issues

PR #281, opened 2017-08-01
Ensure relevant augmenter imputes

jneuff:
There actually was a test named `test_impute_works` but it didn't seem to do anything related to impute.


nils-braun:
Jep, fine with this.

tests/transformers/test_relevant_feature_augmenter.py
```diff
@@ -106,3 +88,19 @@ def test_evaluate_only_added_features_false(self):
...
         self.assertEqual(sum(["pre_keep" == column for column in transformed_X.columns]), 1)
         self.assertEqual(sum(["pre_drop" == column for column in transformed_X.columns]), 0)
+
+
+    @mock.patch('tsfresh.transformers.feature_selector.calculate_relevance_table')
+    def test_does_impute(self, calculate_relevance_table_mock):
+        df = pd.DataFrame([[1, 1, 1], [2, 1, 1]], columns=['id', 'time', 'value'])
+        X = pd.DataFrame(index=[1])
+        y = pd.Series([0, 1])
+        fc_parameters = {"autocorrelation": [{'lag': 2}]}
```
MaxBenChrist: I think to trigger a `np.NaN`, the lag needs to be greater than Len(x)

MaxBenChrist: this unit test should maybe mock that the autocorrelation to make sure that it really returns a np.NaN

nils-braun: Actually, we have already some tests that perform imputing on the data - so this does not have to be tested again, I think

MaxBenChrist: Do we really have a test that checks if the `RelevantFeatureAugmenter` is imputing?

Issue #223, opened 2017-07-03
Feature Request: Partial Autocorrelation Function (PACF)


nikhaas:
For a timeseries with an unknown data generating model, the *autocorrelation function (ACF)* and [*partial autocorrelation function (PACF)*](https://en.wikipedia.org/wiki/Partial_autocorrelation_function) help in identifying the order of an Autoregressive `ARMA(p, q)` model. The [Box-Jenkins method  uses ACF and PACF](https://en.wikipedia.org/wiki/Box%E2%80%93Jenkins_method#Autocorrelation_and_partial_autocorrelation_plots) for this purpose. Depending on the shape of the ACF and PACF, one can derive a model for fitting the data.

The difference between ACF and PACF:

- The ACF calculates the linear relationship between the values at timestep `t` and `t-k` (if one looks at the autocorrelation of lag `k`). So this is an univariate regression problem, which can be solved by Ordinary Least Squares (OLS).
- The PACF calculates the linear relationship between the values at timestep `t` **up to** `t-k` (if one looks at the partial autocorrelation of lag `k`). So this is a **multivariate** regression problem, which can be solved by Ordinary Least Squares (OLS), too. See this [post at SO](https://stats.stackexchange.com/questions/18599/pacf-manual-calculation) for clarification.

#### Current Status

Afaics,`tsfresh` already provides the functions

- [`mean_autocorrelation`](https://github.com/blue-yonder/tsfresh/blob/master/tsfresh/feature_extraction/feature_calculators.py#L264)
- [`autocorrelation`](https://github.com/blue-yonder/tsfresh/blob/master/tsfresh/feature_extraction/feature_calculators.py#L1158)

If I derive it correctly from the statement `"autocorrelation": [{"lag": lag} for lag in range(10)],` from the [`feature_extraction.settings`](https://github.com/blue-yonder/tsfresh/blob/master/tsfresh/feature_extraction/settings.py#L108), the latter is done for the lags 0 (which is trivially 1) to 9.

`tsfresh` also fits an AR process by the method [`ar_coefficient`](https://github.com/blue-yonder/tsfresh/blob/master/tsfresh/feature_extraction/feature_calculators.py#L945). I need to confess that I don't get which params are calculated. It says in the [feature_extraction settings](https://github.com/blue-yonder/tsfresh/blob/master/tsfresh/feature_extraction/settings.py#L117)
```python
[...]

PR #283, opened 2017-08-02
Add partial autocorrelation feature calculator

nikhaas:
* The PACF can be used to determine the lag p of an AR(p) process

* Comprehensive docstring

* Many tests added. Some may be discarded

* Use the option `method="ld"` which is fastest and does not throw
Singular Matrix Errors compared to "yw"

* The OLS method is buggy / wrong (see statsmodels issue #3839)

* Solves tsfresh issue #223

### DIFF
--- a/tsfresh/feature_extraction/feature_calculators.py
+++ b/tsfresh/feature_extraction/feature_calculators.py
@@ -24,7 +24,7 @@ import pandas as pd
 from scipy.signal import welch, cwt, ricker, find_peaks_cwt
 from statsmodels.tsa.ar_model import AR
 from statsmodels.tsa.stattools import adfuller
-from statsmodels.tsa.stattools import acf
+from statsmodels.tsa.stattools import acf, pacf
 from scipy.stats import linregress
 
 # todo: make sure '_' works in parameter names in all cases, add a warning if not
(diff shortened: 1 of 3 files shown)


tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -292,6 +292,58 @@ def agg_autocorrelation(x, param):
...
+    :type param: list
+    :return: the value of this feature
+    :return type: float
+    """
+    # Check the difference between demanded lags by param and possible lags to calculate (depends on len(x))
+    max_demanded_lag = max([lag["lag"] for lag in param])
+    n = len(x)
+
+    # Check if list is too short or consists only of NaNs
+    if n <= 1 or sum(np.isnan(x)) == n:
```
MaxBenChrist: we assume in tsfresh that the time series are clean, meaning that it should only contain finite values

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -292,6 +292,58 @@ def agg_autocorrelation(x, param):
...
+        pacf_coeffs = [np.nan] * (max_demanded_lag + 1)
+    else:
+        if (n <= max_demanded_lag):
+            max_lag = n - 1
+        else:
+            max_lag = max_demanded_lag
+        pacf_coeffs = list(pacf(x, method="ld", nlags=max_lag))
+        pacf_coeffs = pacf_coeffs + [np.nan] * max(0, (max_demanded_lag - max_lag))
+
+    return [("lag_{}".format(lag["lag"]), pacf_coeffs[lag["lag"]]) for lag in param]
```
MaxBenChrist: is pacf starting at lag 0 or 1?

if it starts at lag 1, you have to shift the selection, so `pacf_coeffs[lag["lag"]]-1`, don't you?

nikhaas: `pacf` returns lag 0, too, i.e. the first value of the result is always 1.

tsfresh/feature_extraction/feature_calculators.py
nikhaas: So i delete the check if the series consists only of NaNs. I'll remove the unit test for this case, too.

tsfresh/feature_extraction/feature_calculators.py
MaxBenChrist: Alright

PR #284, opened 2017-08-02
Add dickey_fulley LinAlg and ValueError test

nikhaas:


tests/feature_extraction/test_feature_calculations.py
```diff
@@ -195,6 +195,14 @@ def test_augmented_dickey_fuller(self):
         self.assertLessEqual(res["attr_pvalue"], 0.05)
         self.assertEqual(res["attr_usedlag"], 0)
 
+        # Check if LinAlgError and ValueError are catched
```
MaxBenChrist: You could also achieve this by mocking the `adfuller` method

MaxBenChrist:
We are getting close to 99% coverage, great @nikhaas

tests/feature_extraction/test_feature_calculations.py
nikhaas: What do you mean with "mocking"?

MaxBenChrist: http://www.voidspace.org.uk/python/mock/patch.html

you patch the function, change it to one that raises the error of interest, always, no madder which input

nikhaas: Thx!

PR #289, opened 2017-08-04
Bugfix - augmented Dickey-Fuller feature extractor improperly quoted parameter

ckchow:
The aforementioned feature extractor would put `attr_usedlag` in the column title, causing `get_config_from_string` to parse the attribute correctly. Added quotes like in the FFT feature extractor to fix this, so the resulting column title should be `attr_"usedlag"` instead.


ckchow:
I originally thought this was a bug with `get_config_from_string`, maybe it would be nice to test serializing + deserializing the default extractor params for every feature extractor so this doesn't happen when people add new extractors.

MaxBenChrist:
Thx for fixing this @ckchow I will add a unit test later

PR #290, opened 2017-08-05
Add check for nans

jneuff:
I added the check and refactored the surrounding tests for clarity of intent and legibility.


MaxBenChrist:
Great, @jneuff.

I am fine with merging this after the build passes

PR #292, opened 2017-08-08
Listed contributors

kempa-liehr:
Contributers ordered as shown by https://github.com/blue-yonder/tsfresh/graphs/contributors


MaxBenChrist:
Well, is there a way to automate the construction of the list?

nils-braun:
I think the automatation is possible - but may be hard, as we definitely do not want something like "Coveralls" etc. in it. Lets keep it like this. Maybe people can feel free to add them by themselves.

PR #296, opened 2017-08-15
Bugfix for preventing the subsequent spawning of Python processes, if…

kempa-liehr:
… feature_extraction is called subsequently

It would be cool, if someone had an idea for unit testing this.


MaxBenChrist:
I am not sure I understand which bug this should catch?

kempa-liehr:
This bug occurs, if someone is experimenting with tsfresh.feature_extraction.extraction.extract_features() e.g. in a Jupiter notebook.

In a multi-core environment every call to this function creates some processes, which are not correctly terminated and occupy memory resources.

MaxBenChrist:
Ok. 

Are you sure about the order of the calls?

kempa-liehr:
Yes. I've tested the bugfix in a large scale project, which (before the bugfix) crashed the server again and again. After this fix, everything was fine.

kempa-liehr:
But, travis still needs to be fixed.

MaxBenChrist:
> But, travis still needs to be fixed.

You have to catch the case `n_jobs == 0`. For that case, the pool object does not exist

MaxBenChrist:
> Yes. I've tested the bugfix in a large scale project, which (before the bugfix) crashed the server again and again. After this fix, everything was fine.

This seems to be the way to go. However, before merging this pr I will read into this issue a little bit

MaxBenChrist:
We should probably also close the pool inside `calculate_relevance_table`

kempa-liehr:
Now, travis should be fine and the pool inside `calculate_relevance_table` has also been terminated.

PR #297, opened 2017-08-15
Motif discovery

Ezekiel-Kruglick:
Initial motif capabilities. Currently all still manual and separate from other extractors. PR so we can back on it together.


Ezekiel-Kruglick:
Looks like something is incompatible with 3.5. I'll look into it

MaxBenChrist:
Yes. for example xrange was removed.

There is an easy solution: Just use `six` https://pythonhosted.org/six/ 

It contains a `six.range` and `six.assertCountEqual()` that will work with both py2 and py3. 👍

Ezekiel-Kruglick:
Six looks neat, thanks for pointing me at that. Apologies I didn't implement that before the merge, real life has limited my contribution time the past few days.

MaxBenChrist:
No problem. I will fix it during the next half hour ;)

Issue #276, opened 2017-07-31
Speed up tests suite on Travis


MaxBenChrist:
Currently the Travis build takes way to long.

Also, our tests 

- are not parallelized (we had memeory issues that forced us to limit number of CPUs to 1)
- are not sorted/clustered into unit, integrations and notebook tests.

Doing both should improve the runtime of a travis run and allow more agile development.

see also

https://docs.travis-ci.com/user/speeding-up-the-build/

jneuff:
Good idea! Actually doing a proper split of unit and integration tests would also be helpful for local development. We can tackle this gradually.

nils-braun:
I think speed is my favourite topic :-D I will see what I can do here.

MaxBenChrist:
splitting the test suite into unit tests and integration tests helps. 

Travis gives you only 2 cores, which we already exploit by executing tests in parallel. 

To speed up the testing further, we would have to slim down the tests. The ipython notebooks for example take quite a while to execute

nils-braun:
I would suggest the following solution:
* split up into unit tests and integration tests (as we already do)
* only run the unit tests in pull requests and all tests on master - or even only once per day
* make sure the unit tests have also a good coverage (and not only the integration tests). We are in the moment at ~92%.
* fix the max_langevin and the estimate_friedrich test, as those take ~80% of the whole unit test testing time.

MaxBenChrist:
* only run the unit tests in pull requests and all tests on master - or even only once per day

I am against this. We will end up with a polluted master from time to time

* fix the max_langevin and the estimate_friedrich test, as those take ~80% of the whole unit test testing time.

You sure? I think the notebooks take most of the time

nils-braun:
> You sure? I think the notebooks take most of the time
Yes, I was talking about unit tests - not integration tests.

The thing is: our full test suite takes ~30 minutes. This is definitely too long. I can not wait 30 minutes to have a result on a PR.

MaxBenChrist:
> > You sure? I think the notebooks take most of the time
> > Yes, I was talking about unit tests - not integration tests.
>The thing is: our full test suite takes ~30 minutes. This is definitely too long. I can not wait 30 minutes to have a result on a PR.

but what is the point in getting a fast result after 6 min that is not complete?

MaxBenChrist:
see this pr for example https://travis-ci.org/blue-yonder/tsfresh/builds/269136674?utm_source=github_status&utm_medium=notification

[...]

PR #300, opened 2017-08-17
Travis test

nils-braun:
I have done (more or less) two things in this PR:
* separated out integration and unit tests
* make travis handle both tests independently
This leads to the fact, that a first result is already available after ~5 minutes (unit tests only) and the integration tests are run afterwards together with the unittests to also define the coveralls.
All in all, the total build time is still quite large and we may think about reducing the notebooks, that get tested.


MaxBenChrist:
Related to issue #276

MaxBenChrist:
The cause for its failing is really weird. When I call `pytest tests/units/feature_extraction/test_extraction.py` everything works fine, but when I call `python setup.py test`, like on travis, python fails to create the profiler file

MaxBenChrist:
using absolute paths is not solving this.
in fact, even trying to get the current working directory by `os.getcwd()` failed.

Also, i tried to add `__init__.py` files in the test directories. With no success

MaxBenChrist:
Ok. finally I made it. 

The solution was to create a temporary directory and save the files there.

PR #305, opened 2017-08-26
Fixing TODOs

Ezekiel-Kruglick:
Very major improvements to the _candidates_top_uniques function, thank you for the insightful comments! Your request to not generate those lists led me down three iterations of clearer thought and I really like the new way - it's a lot faster and much easier to read.

Upgraded the tests based on that, and also pulled candidate generation out to it's own method so that I could test it a bit before the uniqueness filter is applied.

Note improvements in use of as_strided as well after some research. I'm not sure if that is done yet or not, I'm still thinking about it.

Still working on the other TODOs


MaxBenChrist:
Great work @Ezekiel-Kruglick! 

The `TestMotifSubelements.test_count_motifs` is failing, even though we are using the six.integer_types. not sure about that. can you look into that?

Ezekiel-Kruglick:
Yup, I'll look into that. I need to figure out a good work flow for testing on Python 3 at my dev machine so I don't have to wait for Travis-CI to detect non-2.7 issues. I will work on it.

MaxBenChrist:
Personally I have two virtual environments, one with Python2.7 called venv another venv3 with Python3.4. when  you create a virtual environment you can specify which Python interpreter should be used. 

I have two bash shells open, the environments activated in each. Then I call Python setup.py test in both. This will make sure that the tests pass on both Python versions.

Ezekiel-Kruglick:
I have py2 and py3 running in venvs now for the future, thanks for the tip. Guess you got to checking it before me though :)

PR #308, opened 2017-08-29
Speed up tests

nils-braun:
Using some "fancy" if conditions with travis, smaller statistics for some tests and general refactoring, I could speed up our unittests frm 27 minutes to 6 minutes on travis.
We are only running the "small" unittests on branches and PR now and the full integration tests on master.
Also deployment is only enabled on tags (as before).


nils-braun:
Well, we do have a coverage drop here of course, because the coverage is now only calculated on the unit tests - not on the integration tests.

MaxBenChrist:
> Well, we do have a coverage drop here of course, because the coverage is now only calculated on the unit tests - not on the integration tests.

I do not think that this pr is the way to go. We will end up with masters with dropped coverages or where the integration tests are failing.

nils-braun:
ok, I have fixed the coverage problem. The only file, that really changed between unit and integration tests is the driftbuf_simulation - which is just an example and is not needed to have full coverage in my opinion.

nils-braun:
Perfect  - unit tests are already finished - integration is still running :-)

PR #314, opened 2017-09-12
Spectral Centroid

aloosley:
I added (and tested) a spectral centroid feature calculator in combination with the fft_coefficient feature calculator.  Maybe such a thing is also useful to others?

### DIFF
--- a/tsfresh/feature_extraction/feature_calculators.py
+++ b/tsfresh/feature_extraction/feature_calculators.py
@@ -827,7 +827,7 @@ def fft_coefficient(x, param):
     # Spectral Centroid:
     fft_abs = np.abs(fft)
     res += [fft_abs[:len(x)//2].dot(np.arange(0, len(x)//2, 1)) / fft_abs[:len(x)//2].sum()]
-    index += ['centroid']
+    index += ['coeff_centroid__attr_abs']
 
     return zip(index, res)
 
--- a/tsfresh/feature_extraction/feature_calculators.py
+++ b/tsfresh/feature_extraction/feature_calculators.py
@@ -827,7 +827,7 @@ def fft_coefficient(x, param):
     # Spectral Centroid:
     fft_abs = np.abs(fft)
     res += [fft_abs[:len(x)//2].dot(np.arange(0, len(x)//2, 1)) / fft_abs[:len(x)//2].sum()]
-    index += ['coeff_centroid__attr_abs']
+    index += ['centroid']
 
     return zip(index, res)
 
--- a/tsfresh/feature_extraction/feature_calculators.py
+++ b/tsfresh/feature_extraction/feature_calculators.py
@@ -799,7 +799,7 @@ def fft_coefficient(x, param):
     :type x: pandas.Series
     :param param: contains dictionaries {"coeff": x, "attr": s} with x int and x >= 0, s str and in ["real", "imag",
         "abs", "angle"]
-    :type param: list
+    :type param: list (e.g. [{"coeff": 0, "attr": "real"}, {"coeff": 0, "attr": "imag"}, ...]
     :return: the different feature values
     :return type: pandas.Series
     """
@@ -823,7 +823,30 @@ def fft_coefficient(x, param):
     res = [complex_agg(fft[config["coeff"]], config["attr"]) if config["coeff"] < len(fft)
            else np.NaN for config in param]
     index = ['coeff_{}__attr_"{}"'.format(config["coeff"], config["attr"]) for config in param]
+
+    # Spectral Centroid:
+    fft_abs = np.abs(fft)
+    res += [fft_abs[:len(x)//2].dot(np.arange(0, len(x)//2, 1)) / fft_abs[:len(x)//2].sum()]
+    index += ['centroid']
+
     return zip(index, res)
+
+
+@set_property("fctype", "simple")
+def spectral_centroid(x):
+    """
+    Returns the spectral centroid, which is the average frequency in the absolute fourier transform spectrum.
+
+    :param x: the time series to calculate the feature of
+    :type x: pandas.Series
+    :return: the value of this feature
+    :return type: float
+    """
+
+    fft_abs = abs(np.fft.rfft(x))
+
+    # Spectral centroid (weights average over abs values):
+    return fft_abs[:len(x)//2].dot(np.arange(0, len(x)//2, 1)) / fft_abs[:len(x)//2].sum()
 
 
 @set_property("fctype", "simple")
--- a/tests/units/feature_extraction/test_feature_calculations.py
+++ b/tests/units/feature_extraction/test_feature_calculations.py
@@ -451,6 +451,7 @@ class FeatureCalculationTestCase(TestCase):
         self.assertAlmostEqual(res['coeff_0__attr_"real"'], sum(x), places=6)
         self.assertAlmostEqual(res['coeff_0__attr_"angle"'], 0, places=6)
         self.assertAlmostEqual(res['coeff_0__attr_"abs"'], sum(x), places=6)
+        self.assertAlmostEqual(res['centroid'], 0.896937, places=2)
 
         x = [0, 1, 0, 0]
         res = pd.Series(dict(fft_coefficient(x, param)))
@@ -463,6 +464,7 @@ class FeatureCalculationTestCase(TestCase):
         self.assertAlmostEqual(res['coeff_1__attr_"real"'], 0, places=6)
         self.assertAlmostEqual(res['coeff_2__attr_"imag"'], 0, places=6)
         self.assertAlmostEqual(res['coeff_2__attr_"real"'], -1, places=6)
+        self.assertAlmostEqual(res['centroid'], 0.5, places=2)
 
         # test what happens if coeff is biger than time series lenght
         x = range(5)


MaxBenChrist:
Hey, great idea @aloosley, we are always happy to see more feature calculators.

Can you add some unit tests?

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -823,10 +823,33 @@ def complex_agg(x, agg):
     res = [complex_agg(fft[config["coeff"]], config["attr"]) if config["coeff"] < len(fft)
            else np.NaN for config in param]
     index = ['coeff_{}__attr_"{}"'.format(config["coeff"], config["attr"]) for config in param]
+
```
MaxBenChrist: This can be removed (possibly a development artefact)

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -823,10 +823,33 @@ def complex_agg(x, agg):
...
+    res += [fft_abs[:len(x)//2].dot(np.arange(0, len(x)//2, 1)) / fft_abs[:len(x)//2].sum()]
+    index += ['centroid']
+
     return zip(index, res)
 
 
 @set_property("fctype", "simple")
+def spectral_centroid(x):
+    """
+    Returns the spectral centroid, which is the average frequency in the absolute fourier transform spectrum.
```
MaxBenChrist: can you add a reference to a paper where this feature is used or explained?

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -823,10 +823,33 @@ def complex_agg(x, agg):
...
+    # Spectral Centroid:
+    fft_abs = np.abs(fft)
+    res += [fft_abs[:len(x)//2].dot(np.arange(0, len(x)//2, 1)) / fft_abs[:len(x)//2].sum()]
+    index += ['centroid']
+
     return zip(index, res)
 
 
 @set_property("fctype", "simple")
+def spectral_centroid(x):
```
MaxBenChrist: we could even have a more generalized feature calculator "aggregated_fft", which performs aggregations on the fft coefficient.

One of those is the spectral centroid, others could be variance?

tsfresh/feature_extraction/feature_calculators.py
aloosley: Yes, remove please.

tsfresh/feature_extraction/feature_calculators.py
MaxBenChrist: maybe you got some more ideas @aloosley ?

aloosley: Yes.  I would support this.  I am currently trying to reproduce some old results that a few (apparently unknown) students made before me.  Allegedly they measured several FT moments including the variance, skew, and kurtosis from the fft and used them to discriminate between vibration types.  

Unfortunately, nothing is documented nor do I have access to their implementation.  Given my goals, I probably won't implement these in the near-term but if anybody else gets to it first, note the following.  Apparently there are useful and less-useful means of calculating them.  For example, spectral kurtosis could be implemented based on this document: https://hal.archives-ouvertes.fr/hal-00021302/document

tsfresh/feature_extraction/feature_calculators.py
aloosley: https://en.wikipedia.org/wiki/Spectral_centroid

aloosley:
For testing, I'm currently following the instructions in this page:

http://tsfresh.readthedocs.io/en/latest/text/how_to_contribute.html?highlight=tests

tsfresh/feature_extraction/feature_calculators.py
MaxBenChrist: Thanks! 👍

MaxBenChrist:
> For testing, I'm currently following the instructions in this page:
> http://tsfresh.readthedocs.io/en/latest/text/how_to_contribute.html?highlight=tests

we introduced a separation of integration and unit tests, the tests for this feature calculator belong to the unit tests.
we should update that page.

tsfresh/feature_extraction/feature_calculators.py
MaxBenChrist: > Unfortunately, nothing is documented nor do I have access to their implementation. Given my goals, I probably won't implement these in the near-term but if anybody else gets to it first, note the following. Apparently there are good and less-good ways of calculating moments from the fft.

you won't have to implement them. just take the numpy implementation of variance, skew, kurtosis etc. and apply them to the fft vector. most work will be to write the unit tests I guess.

tsfresh/feature_extraction/feature_calculators.py
MaxBenChrist: cc @nikhaas

tsfresh/feature_extraction/feature_calculators.py
aloosley: Yes, you are correct that one can leverage scipy.stats to grab the higher order moments of any distribution.  

However, suffice it to say there are apparently a number of subtleties in calculating these features with respect to their physical interpretation.  First, depending on the need, it doesn't seem necessarily best that one should ignore the complex angular part of the fourier spectrum (Ref [1]).  Second, apparently there is some variance in how people implement spectral kurtosis.  Ref [2]'s implementation is slightly different (by an additive factor be a big deal) from scipy's two implementations.  I would have to give some thought to these implementations before feeling comfortable pull requesting them into tsfresh.

Refs:
[1] https://hal.archives-ouvertes.fr/hal-00021302/document
[2] https://de.mathworks.com/matlabcentral/fileexchange/48914-welch-s-estimate-of-the-spectral-kurtosis

But yes, I naïvely think somebody for the time being could just use scipy.stats's implementation and go with it.

aloosley:
> we introduced a separation of integration and unit tests, the tests for this feature calculator belong to the unit tests.
we should update that page.

Can I have a hint as to what other tests I should do?  ✌️ 

Also, what is the best way to make a new feature calculator without having to redo the np.fft.rfft calculation?

Thanks for the help.

MaxBenChrist:
> Can I have a hint as to what other tests I should do? ✌️

you should add some sane defaults to the settings somewhere here https://github.com/blue-yonder/tsfresh/blob/master/tsfresh/feature_extraction/settings.py#L105:L140

i think there are no additional integration tests are necessary. the tests that we have will cover your method.

> Also, what is the best way to make a new feature calculator without having to redo the np.fft.rfft calculation?

unfortunately this is not possible without a breaking change. As I see it, we will have two feature calculator in the future

- fft_coefficient: already implemented, will perform aggregation of individual coefficients
- fft_aggregated: this is the new one, it will perform aggregations of the whole fft vector, so on multiple coefficients

fot the time beeing, I am fine with that

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -823,10 +823,33 @@ def complex_agg(x, agg):
...
+    :param x: the time series to calculate the feature of
+    :type x: pandas.Series
+    :return: the value of this feature
+    :return type: float
+    """
+
+    fft_abs = abs(np.fft.rfft(x))
+
+    # Spectral centroid (weights average over abs values):
+    return fft_abs[:len(x)//2].dot(np.arange(0, len(x)//2, 1)) / fft_abs[:len(x)//2].sum()
```
nikhaas: I interpret the `cc @nikhaas` from Max as a hinto to give my 2cents.

First of all, I see that you are only interested in "well behaved" frequencies, because you discard the last entry if `fft_abs` has an odd number of items. From the docs of [`np.fft.rfft`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.fft.rfft.html), we see that this last frequency is complex in general.

Your [reference from wikipedia](https://en.wikipedia.org/wiki/Spectral_centroid) seems only meaningful in the time domain. But as always with DFT, we carry no information about the time domain. I've also seen a [slightly different implementation](https://gist.github.com/endolith/359724/aa7fcc043776f16f126a0ccd12b599499509c3cc), where the author uses `np.arange(0,1,len(x)//2)` instead of `np.arange(0, len(x)//2, 1)`. But I prefer your implementation because of the association to the bin. Transforming back in the time domain is done by `val * seconds * sampling_freq / bin_no`.

However, I agree with Max: We could normalize all DFT values which yields a discrete probability distribution of the bins and then calculate some moments (if this is useful at all).

tsfresh/feature_extraction/feature_calculators.py
MaxBenChrist: >  I would have to give some thought to these implementations before feeling comfortable pull requesting them into 

👍

tsfresh/feature_extraction/feature_calculators.py
MaxBenChrist: > I interpret the cc @nikhaas from Max as a hinto to give my 2cents.

I cced you because you both seem to work on the same problem (classification of vibration signals)

> DFT, we carry no information about the time domain

Why? we assume that the signal is uniformly sampled for DFT right?

> However, I agree with Max: We could normalize all DFT values which yields a discrete probability distribution of the bins and then calculate some moments (if this is useful at all).

my intention was just to have a feature calculator that calculates transformations of the whole fft vector, which does not mean that it have to be normalized per se.

nikhaas: > Why? we assume that the signal is uniformly sampled for DFT right?

That's true.

> my intention was just to have a feature calculator that calculates transformations of the whole fft vector, which does not mean that it have to be normalized per se.

This might need some discussion. I can think about relevant aggregates w/o normalization (like max or argmax) and w/ normalization (centroid, kurtosis)..
Ok then, thanks for the hint!

aloosley:
Hi guys,

I started a new branch that I'll pull request shortly.  It puts all the aggregated fft features under a combiner feature called:

```
@set_property("fctype", "combiner")
def fft_aggregated(x, param):
```

as recommended above.  This will include variance, skew, etc.  along with the centroid I already calculated.  I am just writing some tests now.

aloosley:
Please also note @MaxBenChrist , the documentation for adding the new feature name below is confusing:

```
Finally, you have to add custom settings if your feature is a simple or combiner feature with parameters. To do so, just append your feature with sane default parameters to the name_to_param dictionary inside the tsfresh.ComprehensiveFCParameters constructor:
```

The `name_to_param` dict and `ComprehensiveFCParameters` is not under `tsfresh.ComprehensiveFCParameters` as suggested.  The docs should be updated to reflect this dict is under `tsfresh.feature_extraction.settings.ComprehensiveFCParameters`.

MaxBenChrist:
I closed this as the other pr contains the relevant changes.

> The name_to_param dict and ComprehensiveFCParameters is not under tsfresh.ComprehensiveFCParameters as suggested. The docs should be updated to reflect this dict is under tsfresh.feature_extraction.settings.ComprehensiveFCParameters.

Thanks for pointing this out, we will have to update the documentation.

PR #316, opened 2017-09-12
Distributed

nils-braun:
This PR introduces a new way how we handle multiprocessing: a class called "Distributor" handles everything and can be easily changed to use e.g. dask, Amazon cloud computing, local multiprocessing e.g.
A first implementation for `map`, `multiprocessing` and `dask` is already given.

I have still not studied the impact on the runtime, but it should actually have more or less none.
We need unittests and an updated documentation for this, but in the moment I do not have the time to do this.


nils-braun:
Do not care too much on the commit messages.... I will remove the nonsense once we merge it :-)

MaxBenChrist:
Great! 👍  Soon we will have a distributed version of tsfresh.

As discussed, I will continue to work on this.

MaxBenChrist:
> This PR introduces a new way how we handle multiprocessing: a class called "Distributor" handles everything and can be easily changed to use e.g. dask, Amazon cloud computing, local multiprocessing e.g.

On first side, I was not sure if we need that abstraction. However, I changed my mind, the abstraction you choose is really nice and way better than my implementation. :+1

nils-braun:
Thanks! Nice to hear that 😊

tsfresh/feature_extraction/extraction.py
```diff
@@ -193,56 +200,57 @@ def _do_extraction(df, column_id, column_value, column_kind,
...
+                                                 progressbar_title="Feature Extraction", address=distributor)
+        else:
+            raise ValueError(distributor + " is not a valid ip address")
 
-    # Map over all those chunks and extract the features on them
-    result = tqdm(map_function(extraction_function, data_in_chunks),
-                  total=total_number_of_expected_results,
-                  desc="Feature Extraction", disable=disable_progressbar)
+    if not isinstance(distributor, Distributor):
+        raise ValueError("the passed distributor is neither None nor an IP address or a Distributor object")
```
nils-braun: Actually, I do not like mixing distributor instances and strings as an argument.
I would prefer that if users want a local dask cluster (which is somehow advance anyway), they just pass their own already created instance...

tsfresh/utilities/distribution.py
```diff
@@ -0,0 +1,252 @@
+# -*- coding: utf-8 -*-
```
nils-braun: I am actually thinking about moving this out into its own module - because it has nothing to do with tsfresh and may be useful to others also.
But lets talk about this after this PR.

tsfresh/utilities/distribution.py
```diff
@@ -0,0 +1,252 @@
...
+    def close(self):
+        self.client.close()
+
+
+class ClusterDaskDistributor(Distributor):
+    """
+    Distributor using a dask cluster, meaning that the calculation is spread over a cluster
+    """
+    def __init__(self, n_workers, disable_progressbar, progressbar_title, address):
+        Distributor.__init__(self, n_workers, disable_progressbar, progressbar_title)
```
nils-braun: the n_workers has no meaning here, or? So better not use it as a constructor argument and set it to None or something.

tsfresh/utilities/distribution.py
```diff
@@ -0,0 +1,252 @@
...
+        result = self.client.gather(self.client.map(func, partitioned_chunks))
+        return result
+
+    def close(self):
+        self.client.close()
+
+
+class ClusterDaskDistributor(Distributor):
+    """
+    Distributor using a dask cluster, meaning that the calculation is spread over a cluster
```
nils-braun: Do you want to tell the user, how this is setup before and what this ip address means?

tsfresh/feature_extraction/extraction.py
MaxBenChrist: It makes sense from the user perspective. I am still working on this branch. Give me a few days ;)

MaxBenChrist:
I am closing this because I am still working on the branch. As said above, give me a few days to get everything working, add the documentation, examples and testing. I will reopen once I have a state that is ready for discussion

tsfresh/feature_extraction/extraction.py
nils-braun: Sure, just wanted to give my opinion :-)

PR #322, opened 2017-09-25
New Feature: fft_aggregated

aloosley:
Hi all,

As mentioned in the [Spectral Centroid PR](https://github.com/blue-yonder/tsfresh/pull/314), I decided it would be cleaner to just restart the PR.  As requested, I have added a new combiner feature called fft_aggregated.  

I added tests for it, they pass.  There are other tests that I'm not sure are related to this new feature that are not passing for me.

Best.

### DIFF
--- a/tests/units/feature_extraction/test_feature_calculations.py
+++ b/tests/units/feature_extraction/test_feature_calculations.py
@@ -482,37 +482,61 @@ class FeatureCalculationTestCase(TestCase):
         ]
         expected_index = ['aggtype_"centroid"', 'aggtype_"variance"', 'aggtype_"skew"', 'aggtype_"kurtosis"']
 
-        x = range(10)
+        x = np.arange(10)
         res = pd.Series(dict(fft_aggregated(x, param)))
         six.assertCountEqual(self, list(res.index), expected_index)
         self.assertAlmostEqual(res['aggtype_"centroid"'], 1.135, places=3)
         self.assertAlmostEqual(res['aggtype_"variance"'], 2.368, places=3)
-        self.assertAlmostEqual(res['aggtype_"skew"'], 2.1, places=3)
-        self.assertAlmostEqual(res['aggtype_"kurtosis"'], 4.467, places=3)
+        self.assertAlmostEqual(res['aggtype_"skew"'], 1.249, places=3)
+        self.assertAlmostEqual(res['aggtype_"kurtosis"'], 3.643, places=3)
 
-        x = range(5)
+        # Scalar multiplying the distribution should not change the results:
+        x = 10*x
         res = pd.Series(dict(fft_aggregated(x, param)))
         six.assertCountEqual(self, list(res.index), expected_index)
-        self.assertAlmostEqual(res['aggtype_"centroid"'], 0.563, places=3)
-        self.assertAlmostEqual(res['aggtype_"variance"'], 0.557, places=3)
-        self.assertAlmostEqual(res['aggtype_"skew"'], 1.396, places=3)
-        self.assertIsNaN(res['aggtype_"kurtosis"'])
+        self.assertAlmostEqual(res['aggtype_"centroid"'], 1.135, places=3)
+        self.assertAlmostEqual(res['aggtype_"variance"'], 2.368, places=3)
+        self.assertAlmostEqual(res['aggtype_"skew"'], 1.249, places=3)
+        self.assertAlmostEqual(res['aggtype_"kurtosis"'], 3.643, places=3)
 
-        x = [0, 1, 0, 0]
-        res = pd.Series(dict(fft_aggregated(x, param)))
-        six.assertCountEqual(self, list(res.index), expected_index)
-        self.assertAlmostEqual(res['aggtype_"centroid"'], 1., places=3)
-        self.assertAlmostEqual(res['aggtype_"variance"'], 0.667, places=3)
-        self.assertAlmostEqual(res['aggtype_"skew"'], 0., places=3)
-        self.assertIsNaN(res['aggtype_"kurtosis"'])
-
+        # The fft of a sign wave is a dirac delta, variance and skew should be near zero, kurtosis should be near 3:
+        # However, in the discrete limit, skew and kurtosis blow up in a manner that is noise dependent and are
+        # therefore bad features, therefore an nan should be returned for these values
         x = np.sin(2 * np.pi / 10 * np.arange(30))
         res = pd.Series(dict(fft_aggregated(x, param)))
         six.assertCountEqual(self, list(res.index), expected_index)
-        self.assertAlmostEqual(res['aggtype_"centroid"'], 3., places=6)
-        self.assertAlmostEqual(res['aggtype_"variance"'], 0., places=6)
-        self.assertAlmostEqual(res['aggtype_"skew"'], 4., places=6)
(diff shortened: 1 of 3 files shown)


tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -826,6 +826,53 @@ def complex_agg(x, agg):
...
+    :type param: list
+    :return: the different feature values
+    :return type: pandas.Series
+    """
+
+    assert set([config["aggtype"] for config in param]) <= set(["centroid", "variance", "skew", "kurtosis"]), \
+        'Attribute must be "centroid", "variance", "skew", "kurtosis"'
+
+
+    def get_moment(y, moment):
```
MaxBenChrist: can you please add a docstring here?

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -826,6 +826,53 @@ def complex_agg(x, agg):
...
+        return y.dot(np.arange(len(y)) ** moment) / y.sum()
+
+    def get_centroid(y):
+        return get_moment(y, 1)
+
+    def get_variance(y):
```
MaxBenChrist: can you please add a docstring here?

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -826,6 +826,53 @@ def complex_agg(x, agg):
...
+    """
+
+    assert set([config["aggtype"] for config in param]) <= set(["centroid", "variance", "skew", "kurtosis"]), \
+        'Attribute must be "centroid", "variance", "skew", "kurtosis"'
+
+
+    def get_moment(y, moment):
+        return y.dot(np.arange(len(y)) ** moment) / y.sum()
+
+    def get_centroid(y):
```
MaxBenChrist: can you please add a docstring here?

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -826,6 +826,53 @@ def complex_agg(x, agg):
...
+    def get_variance(y):
+        return get_moment(y, 2) - get_centroid(y) ** 2
+
+    def get_skew(y):
+        return skewness(y)
+
+    def get_kurtosis(y):
+        return kurtosis(y)
+
+    calculation = dict(
```
MaxBenChrist: I like the way how you calculate the attributes by help of the calculation dictionary

tests/units/feature_extraction/test_feature_calculations.py
```diff
@@ -473,6 +473,47 @@ def test_fft_coefficient(self):
...
+            {"aggtype": "variance"},
+            {"aggtype": "skew"},
+            {"aggtype": "kurtosis"}
+        ]
+        expected_index = ['aggtype_centroid', 'aggtype_variance', 'aggtype_skew', 'aggtype_kurtosis']
+
+        x = range(10)
+        res = pd.Series(dict(fft_aggregated(x, param)))
+        six.assertCountEqual(self, list(res.index), expected_index)
+        self.assertAlmostEqual(res['aggtype_centroid'], 1.135, places=3)
```
MaxBenChrist: This looks like you ran the method on the input and fixated the result (in this case `1.135`).

So, this unit test does not check if the calculation is right. Instead it checks if the value of the feature is not changing between tsfresh version.

I would prefer a result that is calculated by hand. Can you think of a border case, where we can calculate the feature by hand?

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -826,6 +826,53 @@ def complex_agg(x, agg):
...
+        centroid=get_centroid,
+        variance=get_variance,
+        skew=get_skew,
+        kurtosis=get_kurtosis
+    )
+
+    fft_abs = abs(np.fft.rfft(x))
+
+    res = [calculation[config["aggtype"]](fft_abs) for config in param]
+    index = ['aggtype_{}'.format(config["aggtype"]) for config in param]
```
MaxBenChrist: the error is related to the inferring of a feature calculators parameter from the column name.

To correct this, you need to add quotation marks around the string paramter, it should be 

```
index = ['aggtype_"{}"'.format(config["aggtype"]) for config in param]
```

MaxBenChrist:
Great work @aloosley, we are close to merge this into master 👍

tsfresh/feature_extraction/feature_calculators.py
aloosley: Done

tsfresh/feature_extraction/feature_calculators.py
aloosley: Done

tsfresh/feature_extraction/feature_calculators.py
aloosley: Done

tsfresh/feature_extraction/feature_calculators.py
aloosley: Thanks.

tsfresh/feature_extraction/feature_calculators.py
aloosley: Done.  I updated the tests accordingly.

MaxBenChrist: this line was not touched, was it?

tests/units/feature_extraction/test_feature_calculations.py
```diff
@@ -473,6 +473,47 @@ def test_fft_coefficient(self):
...
         self.assertIsNaN(res['coeff_10__attr_"real"'])
 
+    def test_fft_aggregated(self):
+        param = [
+            {"aggtype": "centroid"},
+            {"aggtype": "variance"},
+            {"aggtype": "skew"},
+            {"aggtype": "kurtosis"}
+        ]
+        expected_index = ['aggtype_centroid', 'aggtype_variance', 'aggtype_skew', 'aggtype_kurtosis']
```
MaxBenChrist: the expected index needs to have quotation marks around the attributes

aloosley: Ok, I will push now, please note the tests will fail because I'm in the middle of fixing the kurtosis and skew calculations (it was incorrect to use the pandas.Series skew/kurtosis functions that I had committed).

aloosley:
Please give some time for me to finish fixing the skew and kurtosis calculations and to add a few tests.

MaxBenChrist:
> Please give some time for me to finish fixing the skew and kurtosis calculations and to add a few tests.

Sure sure, take your time :)

aloosley:
Everything should be set now.  I also fixed kurtosis and skew calculations for the edge cases where the discrete calculations blow up and are no longer meaningful. For example, with a dirac delta, the 0/0 division for kurtosis should limit to 3 but gives nonsense in the real world case because of noise and rounding. I also added hand calculated test values for x=sin() and x=Normal() cases.

aloosley:
One final push to I just fixed up a few bad comments I had glossed over.

MaxBenChrist:
Wow. Great work @aloosley! 

Happy to merge this one!

aloosley:
Thanks for the back and forth with me.  That was great.

MaxBenChrist:
> Thanks for the back and forth with me. That was great.

Thanks. Looking forward to see more feature calculators from you ;)

PR #333, opened 2017-09-30
Distributed

MaxBenChrist:
So we are getting close to a distributed version of tsfresh.

I refactored the modules, wrote docstrings, added unit tests and added a chapter in the docs that explains how to use the distributor opbjects.

So, what is missing? I think we should still increase the test coverage, for example we can add some tests for the localdaskdistributor.

I am happy about feedback what still needs to be tested


MaxBenChrist:
There is still a small problem with the format of the features that are returned by the two Dask Distributors. I will upload a fix with some unit tests tomorrow

MaxBenChrist:
> There is still a small problem with the format of the features that are returned by the two Dask Distributors. I will upload a fix with some unit tests tomorrow

Done

tsfresh/utilities/string_manipulation.py
```diff
@@ -1,6 +1,13 @@
+# -*- coding: utf-8 -*-
+# This file as well as the whole tsfresh package are licenced under the MIT licence (see the LICENCE.txt)
+# Maximilian Christ (maximilianchrist.com), Blue Yonder Gmbh, 2016
+
 import ast
 import numpy as np
 from six import string_types
+import ipaddress
```
nils-braun: this is not needed anymore. Please remove :-)

tsfresh/feature_extraction/extraction.py
```diff
@@ -208,28 +207,31 @@ def _do_extraction(df, column_id, column_value, column_kind,
...
+                                         progressbar_title="Feature Extraction")
         else:
-            distributor_class = MultiprocessingDistributor
+            distributor = MultiprocessingDistributor(n_workers=n_jobs, disable_progressbar=disable_progressbar,
+                                                     progressbar_title="Feature Extraction")
 
-        distributor = distributor_class(n_workers=n_jobs,
-                                        disable_progressbar=disable_progressbar,
-                                        progressbar_title="Feature Extraction")
+    if not isinstance(distributor, DistributorBaseClass):
```
nils-braun: do we really want to be that strict? If a user implements everything that is needed, it is fine to have his own base class (he is of course responsible on his own).
Because of duck typing, it is barely needed to check for a type in python....

tsfresh/feature_extraction/extraction.py
```diff
@@ -208,28 +207,31 @@ def _do_extraction(df, column_id, column_value, column_kind,
...
     :return: the extracted features
     :rtype: pd.DataFrame
     """
     data_in_chunks = [x + (y,) for x, y in df.groupby([column_id, column_kind])[column_value]]
 
     if distributor is None:
+
         if n_jobs == 0:
-            distributor_class = MapDistributor
+            distributor = MapDistributor(disable_progressbar=disable_progressbar,
```
nils-braun: you could summarize the arguments into a dict and fill it in in both cases, to reduce code duplication.

tsfresh/feature_extraction/extraction.py
```diff
@@ -111,10 +110,10 @@ def extract_features(timeseries_container, default_fc_parameters=None,
     :param profiling_filename: Where to save the profiling results.
     :type profiling_filename: basestring
 
-    :param distributor_class: Advanced parameter: set this to a class name that you want to use as a
+    :param distributor: Advanced parameter: set this to a class name that you want to use as a
              distributor. See the utilities/distribution.py for more information. Leave to None, if you want
              TSFresh to choose the best distributor.
-    :type distributor_class: class
+    :type distributor: class
```
nils-braun: why two descriptions?

nils-braun:
Just for not loosing any information here: you have "just" added tests and documentation (thanks for that!) and not changed anything on the implementation itself, correct?

MaxBenChrist:
> Just for not loosing any information here: you have "just" added tests and documentation (thanks for that!) and not changed anything on the implementation itself, correct?

No, I changed a few things. For example the result list from the Dask Distributors needed to be flattened. Also I changed how the Distributor class is called inside `extract_features`

tsfresh/feature_extraction/extraction.py
MaxBenChrist: I want to catch stupid errors (e.g. user forgetting to instantiate the distributor object) with that strict class check.

I do not think that a user will be able to implement a custom distributor without inheriting from the base class. To write the distributor properly, you really need to understand how tsfresh uses it.

Also, there are not too many cluster frameworks, so also the number of possible distributors is limited. 

All in all, I think this motivates the strict test instead of duck typing.

tsfresh/utilities/string_manipulation.py
MaxBenChrist: Eagle eyes, Done!

tsfresh/feature_extraction/extraction.py
MaxBenChrist: ? not sure I understand your comment here

tsfresh/feature_extraction/extraction.py
MaxBenChrist: I think it is quite easy to read in its current state...

nils-braun:
Ah sorry, I meant in the distributed class itself (distributed.py) :-)
Ok, I have seen your changes in feature_extraction :+1:

MaxBenChrist:
> Ah sorry, I meant in the distributed class itself (distributed.py) :-)
> Ok, I have seen your changes in feature_extraction 👍

Some changes that come to my mind:

I moved the implementation of one of the methods from base class to the child classes.
Also, the constructor of every child class has different arguments now. 
I made the calculate chunksize method non private

PR #339, opened 2017-10-21
Performance improvement for min, max and number_crossing.

dangom:


nils-braun:
Nice! Just out of curiosity: by how much did this increase the performance?

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -1493,11 +1493,11 @@ def number_crossing_m(x, m):
     :param m: the threshold for the crossing
     :type m: float
     :return: the value of this feature
-    :return type: float
+    :return type: int
     """
     x = np.asarray(x)
     x = x[x != m]
-    return sum(np.abs(np.diff(np.sign(x - m))))/2
```
MaxBenChrist: can you explain why your solution is faster?

dangom: I guess really because of the usage of vanilla `sum`, instead of numpy.
Here are some benchmarks:

```python
import numpy as np

x = 2 * np.random.rand(1000) - 1

%timeit sum(np.abs(np.diff(np.sign(x))))/2
%timeit np.sum(np.abs(np.diff(np.sign(x))))/2
%timeit np.where(np.diff(np.sign(x)))[0].size
%timeit np.where(np.diff(np.signbit(x)))[0].size
```

> 114 µs ± 5.49 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
> 13.9 µs ± 469 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
> 17.6 µs ± 667 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
> 10.6 µs ± 464 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

Using `np.sum` instead of sum would be indeed faster than the solution I originally proposed.

Here are results using a vector with 10k elements, instead of 1k:

> 1.25 ms ± 83 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
> 81.4 µs ± 7.86 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
> 216 µs ± 38 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
> 61 µs ± 2.54 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)


And there is actually a [thread at stack overflow](https://stackoverflow.com/questions/3843017/efficiently-detect-sign-changes-in-python) with even faster solutions, but I haven't had the time to test them:

Example without using `np.diff`:
```python
pos = x > m
%timeit np.where(np.bitwise_xor(pos[1:], pos[:-1]))[0].shape
```
For 1k:
> 4.28 µs ± 260 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

For 10k:
> 43.4 µs ± 730 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

dangom:
@nils-braun About 10x performance improvement for small vecs when using numpy's optimized libs instead of vanilla python. Numpy also scales better so the performance improvement is greater for larger time series.

I guess there are probably more feature calculators where such small changes can make a big difference.

tsfresh/feature_extraction/feature_calculators.py
MaxBenChrist: Thanks for the explanation.

just one remark:

```
pos = x > m
%timeit np.where(np.bitwise_xor(pos[1:], pos[:-1]))[0].shape 
```

here the timeit should capture both expressions, shouldn't it? Like
```
%timeit pos = x > m; np.where(np.bitwise_xor(pos[1:], pos[:-1]))[0].shape 
```

MaxBenChrist:
> @nils-braun About 10x performance improvement for small vecs when using numpy's optimized libs instead of vanilla python. Numpy also scales better so the performance improvement is greater for larger time series.

This is great, do you want to extend your pr by the other changes?

tsfresh/feature_extraction/feature_calculators.py
dangom: Right, I'll run more benchmarks later and update the PR once I'm done.

dangom:
Sure thing. I'll update the PR

MaxBenChrist:
Wow. Thanks @dangom.  I am sure there are datasets where those speed ups bear fruits.

PR #352, opened 2017-12-01
adding binder requirements + badge

choldgraf:
This is a quick PR to add some pieces necessary to allow the examples in this repo to run on `mybinder.org`. This would make it possible for people to run your example notebooks interactively in the cloud by clicking on the badge in your README (or just going to the URL).

Basically all I did was copy over your requirements.txt file to a `binder` folder and added a few extra dependencies that were called from the notebooks (including the `tsfresh` library itself). Let me know if there's another way you'd prefer to handle dependencies, or if you have other questions in general. I think it'd be useful for people to be able to run the examples interactively without installing `tsfresh` / dependencies on their own machine!

Here's what would happen if people go to the binder-version of this repo (click link below) though note that currently there are dependencies missing that need to be installed, such as matplotlib (that's what this PR is addressing)

https://mybinder.org/v2/gh/blue-yonder/tsfresh/master?filepath=notebooks

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@
 [![Gitter chat](https://badges.gitter.im/tsfresh/Lobby.svg)](https://gitter.im/tsfresh/Lobby?utm_source=share-link&utm_medium=link&utm_campaign=share-link)
 ![py27 status](https://img.shields.io/badge/python2.7-supported-green.svg)
 [![py352 status](https://img.shields.io/badge/python3.5.2-supported-green.svg)](https://github.com/blue-yonder/tsfresh/issues/8)
+[![Binder](https://mybinder.org/badge.svg)](https://mybinder.org/v2/gh/blue-yonder/tsfresh/master?filepath=notebooks)
 
 
 
--- a/binder/generate_binder_requirements.py
+++ b/binder/generate_binder_requirements.py
@@ -1,0 +1,16 @@
+from glob import glob
+
+use_files = ["../requirements.txt", "../test-requirements.txt"]
+files = set()
+
+# Collect a list of dependencies
+for ifile in use_files:
+    with open(ifile, 'r') as ff:
+        lines = ff.readlines()
+        files = files.union(lines)
+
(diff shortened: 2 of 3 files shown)


nils-braun:
Very nice and good idea! I was actually thinking about doing such a thing by myself. Thank you you have done it!
We have already a requirements file for most of those things (test-requirements.txt). Maybe we can merge them together?
@MaxBenChrist what do you think?

choldgraf:
looping back into this now that the holidays are over. It looks like the Travis error isn't related to this (seems to be related to Dask)

Re: consolidating requirements files, the trick here is that for Binder it _must_ be called `requirements.txt`, so things like `dev-requirements.txt` won't work. To that extent I usually just created a `binder` folder and make sure that the requirements are up-to-date in there.

In the near-ish future, Binder will support the new `pipfile` configuration which will make this a lot simpler.

For this PR, is there anything I should add to the requirements file to make sure the examples will run?

MaxBenChrist:
Sorry for the late answer. Thank you @choldgraf for the great addition.

The travis error is not related to this pr. Once I have more time, I will look into that.

At the moment, the packages that are needed for the tsfresh notebooks are contained in the `test-requierements.txt`. Arguably not the best solution...

If I understand you right, for binder, a script that automatically merges our top level `requirements.txt` with the top level `test-requirements.txt` and put the result  into the `requirements.txt` inside the `binder` folder would be favorable, right?

choldgraf:
yep - that's a good solution. I've included such a script and an updated `requirements.txt` file in the latest push!

choldgraf:
hey folks - is this blocking on me at all? I think it's ready to go from my end

nils-braun:
Hi @choldgraf, no, you have made a wonderful job.
Unfortunately, we are a bit short on time on our side in the moment - and we should check the non-running test. Sorry for the delay :-(

choldgraf:
no problem! was just making sure I wasn't blocking on something :-)

PR #395, opened 2018-07-19
Added code to generate visualization

supergitacc:
Added code to generate visualization for easier reproducibility


MaxBenChrist:
Thanks for the pr @supergitacc, I have to find out why the `RollingTestCase.test_warning_on_non_uniform_time_steps` from the python 3 build fails.

PR #372, opened 2018-03-22
local_dir for distributed LocalCluster

earthgecko:
Fix https://github.com/blue-yonder/tsfresh/issues/371

Pass tempfile local_dir to distributed LocalCluster

And Correct shell syntax in .travis.yml as per documented in #370

PR #404, opened 2018-08-17
fix for index preservation and test cases.

QEDan:
The added tests show a problem where DataFrame indices are converted to float if they are "nearly" floats. The test case uses indices with the format 99999_9999_99. A small change to fix this. Thanks for the great library.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,6 +11,3 @@ tqdm>=4.10.0
 ipaddress>=1.0.18; python_version <= '2.7'
 dask>=0.15.2
 distributed>=1.18.3
-mock>=2.0.0
-nbformat>=4.4.0
-pytest>=3.7.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,3 +11,6 @@ tqdm>=4.10.0
 ipaddress>=1.0.18; python_version <= '2.7'
 dask>=0.15.2
 distributed>=1.18.3
+mock>=2.0.0
+nbformat>=4.4.0
+pytest>=3.7.1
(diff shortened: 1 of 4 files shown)


earthgecko:
Hi @QEDan those test failures are not specific to your build the tests are broken, due to notebooks test time outs that are occurring across the board which are related to travis timeouts, secondly there is a syntax error in the bash in travis.yml which also causes tests to fail (pr #372).  All the warnings hard to say, could very well be package drift and all deprecation warnings, there are quite a few.

I did start to look at getting stuck into it all again this afternoon, in relation to other package drift (#405) however I realised on looking at it I would have to start from scratch to only get to the same place and only then could the package drift since March be addressed.  That is going to take longer than an afternoon.

MaxBenChrist:
Hi @QEDan thanks for the pr, I will need some more time to review it thoroughly

MaxBenChrist:
can you rebase this pr to current master?

MaxBenChrist:
I can confirm that bug. 

The casting as float in that line on the results is causing issues if the id column can be expressed as float. 

What is weird is that on my machine, the bug "fixes" itself after some time. I have to create a new virtual environment to reproduce the bug again.

Thanks again for reporting this @QEDan.

MaxBenChrist:
One last question: how did you find that bug? we are having reports of out of memory issues lately, maybe this causes this

QEDan:
@MaxBenChrist Thanks for merging. I wasn't intentionally bug hunting. I was using tsfresh and was unlucky enough that the data set I was working with hit this edge case that caused it to crash with an error. So, I dug in with a debugger and had to implement this fix locally to get things working. I don't think this is caused by another bug, but just an edge case when you work with DataFrame indices that have certain properties.

MaxBenChrist:
Thanks for feedback. Do you still have the error log that pointed you in this direction?

Von meinem iPhone gesendet

Issue #414, opened 2018-08-21
pandas change - Grouping by a tuple of keys


earthgecko:
As of pandas>=0.23.0 there are additional FutureWarning outputs related to:

```python

>>> dataframe_functions.roll_time_series(df_full, column_id="id", column_sort="time",
...                                            column_kind=None, rolling_direction=1)
/opt/python_virtualenv/projects/tsfresh-branch2540-py352/tsfresh/tsfresh/utilities/dataframe_functions.py:388: FutureWarning: Interpreting tuple 'by' as a list of keys, rather than a single key. Use 'by=[...]' instead of 'by=(...)'. In the future, a tuple will always mean a single key.
  differences = df.groupby(grouper)[column_sort].apply(
/opt/python_virtualenv/projects/tsfresh-branch2540-py352/tsfresh/tsfresh/utilities/dataframe_functions.py:394: UserWarning: Your time stamps are not uniformly sampled, which makes rolling nonsensical in some domains.
  warnings.warn("Your time stamps are not uniformly sampled, which makes rolling "
/opt/python_virtualenv/projects/tsfresh-branch2540-py352/tsfresh/tsfresh/utilities/dataframe_functions.py:400: FutureWarning: Interpreting tuple 'by' as a list of keys, rather than a single key. Use 'by=[...]' instead of 'by=(...)'. In the future, a tuple will always mean a single key.
  grouped_data = df.groupby(grouper)
```

http://pandas.pydata.org/pandas-docs/version/0.23/whatsnew.html#deprecations

> Grouping by a tuple of keys now emits a FutureWarning and is deprecated. In the future, a tuple passed to 'by' will always refer to a single key that is the actual tuple, instead of treating the tuple as multiple keys. To retain the previous behavior, use a list instead of a tuple (GH18314)

These 2 new warnings are present in Travis tests at pandas==0.23.4 - https://travis-ci.org/blue-yonder/tsfresh/jobs/418592164#L1313
The RollingTestCase.test_warning_on_non_uniform_time_steps is failing as result of these 2 additional warnings.

The issue is in tsfresh/utilities/dataframe_functions.py - https://github.com/blue-yonder/tsfresh/blob/master/tsfresh/utilities/dataframe_functions.py#L372

```python

    if column_kind is not None:
        grouper = (column_kind, column_id)
    else:
        grouper = (column_id,)
```
A github search suggests this is the only place where the column_kind and column_id are passed as a tuple.

[...]

Issue #412, opened 2018-08-20
update-dependencies


earthgecko:
The dependencies are moving and requirements being specified with ```>=``` are now causing problems  for new users.  Seeing as the new deps versions are untested deprecation warnings are becoming errors, e.g. #405 

Attempting to update the dependencies and identify errors to debug and fix, this issue is to track all the tests in the associated pull request/s.

earthgecko:
## Not pip cache - numpy 1.11.1 is bundled in the Travis python-2.7.11.tar.bz2

Same result without pip cache folder - https://travis-ci.org/blue-yonder/tsfresh/jobs/421458427#L461

```
9.44s$ pip install numpy scipy coveralls
Requirement already satisfied: numpy in /home/travis/virtualenv/python2.7.11/lib/python2.7/site-packages (1.11.1)
Collecting scipy
  Downloading https://files.pythonhosted.org/packages/2a/f3/de9c1bd16311982711209edaa8c6caa962db30ebb6a8cc6f1dcd2d3ef616/scipy-1.1.0-cp27-cp27mu-manylinux1_x86_64.whl (30.8MB)
```

So checking https://s3.amazonaws.com/travis-python-archives/binaries/ubuntu/14.04/x86_64/python-2.7.11.tar.bz2
as the only explanation is that it is bundled in the archive.

```
cd /tmp
wget https://s3.amazonaws.com/travis-python-archives/binaries/ubuntu/14.04/x86_64/python-2.7.11.tar.bz2
mkdir -p /tmp/travis-python-2.7.11
tar xjf python-2.7.11.tar.bz2 --directory /tmp/travis-python-2.7.11/

..
..
$ mkdir -p /tmp/travis-python-2.7.11
$ tar xjf python-2.7.11.tar.bz2 --directory /tmp/travis-python-2.7.11/
$ ls -al /tmp/travis-python-2.7.11/home/travis/virtualenv/python2.7.11/lib/python2.7/site-packages/numpy-1.11.1.dist-info/metadata.json
-rw-rw-r-- 1 gary gary 1351 Aug 25  2016 /tmp/travis-python-2.7.11/home/travis/virtualenv/python2.7.11/lib/python2.7/site-packages/numpy-1.11.1.dist-info/metadata.json
$
```

earthgecko:
## Review

Travis Python archives bundles have numpy bundled:

- Python 2.7.11 bundled with numpy==1.11.1
[...]

PR #420, opened 2018-08-28
Update dependencies

earthgecko:
As per described in #412 and tested with the testing in pr #414
- Overcome Travis bundled versions of numpy in the Python archives - Python 2.7.11 bundled with numpy==1.11.1 and Python 3.5.2 bundled with numpy==1.14.5
- Fix Sphinx for now to Sphinx==1.6.4 otherwise docs will not build
- Split the pip install from -r requirements.txt -r test-requirements.txt to to two separate installs to allow for the update the bundled numpy versions, specifically for Python 3.5.2 as pip will not allow for multiple declarations of the same package in a single run.


requirements.txt
```diff
@@ -1,5 +1,6 @@
 requests>=2.9.1
-numpy>=1.10.4
+# Travis python-2.7.11.tar.bz2 has numpy 1.11.1 bundled GH #412
+numpy>=1.11.2
```
MaxBenChrist: I am against increasing the numpy version here only to please travis

MaxBenChrist:
I get that we maybe have to test against different numpy versions. We could use https://docs.travis-ci.com/user/build-stages/matrix-expansion/ build stages for that

Issue #385, opened 2018-06-07
improve documentation on chsnks and chunk size parameter


adrianloy:
Hi,

I am unsure what exactly the chunk parameter for the `extract_features` function is. Does it refer to rows? Percentage of the dataframe? in memory size of one chunk? And what happens if the default value of None is set? The docstring isn't very helpful here.

I currently have problems of getting out of memory exceptions, can this parameter help me there? 

Thanks in advance,
Adrian

MaxBenChrist:
Hi @adrianloy,

the documentation could be improved here. 

Its quite easy, one chunk is a singular time series for one id and one kind. If you set the chunk size to 10, then it means that one task is to calculate all features for 10 time series. 

If you set it to None, we use, depending on distributor, heuristics to find the optimal chunk size. If you get out of memory exceptions, you can try it with the dask distributor and smaller chunk sizes. 

How big is your data? how big is your memory?

adrianloy:
Thanks for the clarification! Would be nice to have in the docs indeed.

PR #423, opened 2018-08-30
#385 improve documentation on chsnks and chunk size parameter

earthgecko:
- Updated docs with with more info on chunksize
- Updated chunksize docstrings with more info on chunksize

Modified:
docs/text/parallelization.rst
tsfresh/convenience/relevant_extraction.py
tsfresh/feature_extraction/extraction.py
tsfresh/feature_selection/relevance.py
tsfresh/feature_selection/selection.py
tsfresh/transformers/feature_augmenter.py
tsfresh/transformers/relevant_feature_augmenter.py

### DIFF
--- a/docs/text/parallelization.rst
+++ b/docs/text/parallelization.rst
@@ -22,7 +22,14 @@ otherwise idle) processors.
 
 The chunksize of the Pool's map function is another important parameter to consider. It can be set via the
 `chunksize` field. By default it is up to
-:class:`multiprocessing.Pool` to decide on the chunksize.
+:class:`multiprocessing.Pool` to decide on the chunksize.  The chunksize is the
+size of one chunk that is submitted to the worker process for the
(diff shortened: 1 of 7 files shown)


MaxBenChrist:
Good idea! Can you change that description a little bit? I would write
```
:class:`multiprocessing.Pool` is parallelisation parameter.  
One data chunk is defined as a singular time series for one id and one kind. 
The chunksize is the number of chunks that are submitted as one task to one worker process.  
If you set the chunksize to 10, then it means that one worker task corresponds to calculate all features for 10 id/kind time series combinations.  
If it is set it to None, depending on distributor, heuristics are used to find the optimal chunksize.
The chunksize can have an crucial influence on the optimal cluster performance and should be optimised in benchmarks for the problem at hand.
```

earthgecko:
@MaxBenChrist modified as requested, all done,

PR #429, opened 2018-09-04
Testing numpy latest and 1.10.4 stages #412

earthgecko:
For test purposes only relating to #412

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ install:
 jobs:
   include:
     - stage: Run unit tests and deploy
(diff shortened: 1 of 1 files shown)


earthgecko:
As desired in #412 testing numpy latest and 1.10.4

MaxBenChrist:
Wow, great work @earthgecko! maybe we can even expand this matrix to either use the oldest version of all the packages that we specify (e.g. numpy=1.10.4, scipy=0.17.0) or the latest

Issue #449, opened 2018-10-19
Do the tsfresh packaged provide batch processing over a distributed architecture or not?


ricardovvargas:
Hi. I am confused because the [introduction section](https://tsfresh.readthedocs.io/en/latest/text/introduction.html) of the tsfresh documentation says that "tsfresh is not suitable for batch processing over a distributed architecture, where different time series are fragmented over different computational units" but [this page](https://tsfresh.readthedocs.io/en/latest/text/tsfresh_on_a_cluster.html) says the following: "Indeed, it is possible to extract features with tsfresh in a distributed fashion."

Did I miss something or is this just a case of outdated documentation?

MaxBenChrist:
Both statements are correct and indeed tsfresh will run on cluster. (I ran tsfresh on 100 of machines for weeks on huge time series data sets).

Now, the part 

> "tsfresh is not suitable for batch processing over a distributed architecture, where different time series are fragmented over different computational units"

tries to explain that all time series need to be on one machine, the dask master. So, if your time series are fragmented over the cluster, you will need to consolidate it on one machine before starting the feature extraction. Its a technical detail and the statement is probably a little bit confusing, will see if I can formulate it more clearly.

ricardovvargas:
Humm... Now it is clear. Thanks for this explanation. My mother language is brazilian portuguese and I am not an english grammar specialist but as far as I know the term "where" explain something about the last sentence while the term "when" do a restriction. So I believe that the following minor change would make this part of the documentation much more clear: "tsfresh is not suitable for batch processing over a distributed architecture when different time series are fragmented over different computational units". Anyway, this is just a suggestion. Tsfresh is a very nice package. Congratulations.

ricardovvargas:
Yes, sure. I really intend to use this great package at work (Petrobras S.A) and in my doctoral project and, as a counterpart, collaborate as much as I can. The problem is that I am not so familiar with working with GitHub yet : - ))

After clicking the "New pull request", what am I supposed to do? Do you suggest any tutorial that I can follow to avoid messing up with the project or this repository?

PR #455, opened 2018-10-30
improve documentation (#449)

ricardovvargas:
Hi. This is my first PR in GitHub. Did I do everything correctly? I hope that I can contribute more from now on. : - ))

By the way, can/should I close the issue #449 ? After all, the question is now answered. I am asking that because I am new in GitHub and I noticed that the tsfresh project does not have a code of conduct.

### DIFF
--- a/docs/text/introduction.rst
+++ b/docs/text/introduction.rst
@@ -44,8 +44,7 @@ What not to do with tsfresh?
 Currently, tsfresh is not suitable
 
     * for usage with streaming data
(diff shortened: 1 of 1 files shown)


nils-braun:
Yes, everything is fine!
(Just as a side note although not needed in this case: if you open a PR you normally try to explain what you have done in the pull request, so other people do not need to read through all the changes first to get an overview and then read through the changed again to find out what exactly you have changed. But for this PR it is not needed).
Thanks!

PR #457, opened 2018-11-07
Fix errors in the tests

delyanr:
Changelog:
* Fix a couple of **PermissionError** under WinNT operating systems
* Update the Benjamini-Yekutieli notebook to use **calculate_relevance_table()** instead of the deprecated **check_fs_sig_bh()**
* Update the Time Series Basic Example notebook to feed the **AdaBoostClassifier predict()** method correctly

### DIFF
--- a/tests/integrations/test_notebooks.py
+++ b/tests/integrations/test_notebooks.py
@@ -32,7 +32,12 @@ def _notebook_run(path, timeout=default_timeout):
     except:
         pass
 
-    with tempfile.NamedTemporaryFile(mode='w+t', suffix=".ipynb") as fout:
+    # Ensure temporary files are not auto-deleted as processes have limited 
+    # permissions to re-use file handles under WinNT-based operating systems.
+    fname = ''
+    with tempfile.NamedTemporaryFile(mode='w+t', suffix=".ipynb", delete=False) as fout:
+        fname = fout.name
+        
         args = ["jupyter", "nbconvert",
                 "--to", "notebook", "--execute", execproc_timeout]
         if six.PY2:
@@ -44,6 +49,7 @@ def _notebook_run(path, timeout=default_timeout):
 
         fout.seek(0)
         nb = nbformat.read(fout, nbformat.current_nbformat)
+    os.remove(fname)
 
     errors = [output for cell in nb.cells if "outputs" in cell
               for output in cell["outputs"] \
--- a/tests/units/scripts/test_run_tsfresh.py
+++ b/tests/units/scripts/test_run_tsfresh.py
@@ -18,6 +18,7 @@ class RunTSFreshTestCase(TestCase):
         # Create a temporary directory
         self.test_dir = tempfile.mkdtemp()
         # Change into the tmp dir
+        self.curr_dir = os.getcwd()
         os.chdir(self.test_dir)
 
         def extract_features_mock(df, **kwargs):
@@ -31,6 +32,7 @@ class RunTSFreshTestCase(TestCase):
 
     def tearDown(self):
         # Remove the directory after the test
+        os.chdir(self.curr_dir)
         shutil.rmtree(self.test_dir)
         # Disable the mocking
         self.patcher.stop()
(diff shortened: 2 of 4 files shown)


tests/integrations/test_notebooks.py
```diff
@@ -44,6 +49,7 @@ def _notebook_run(path, timeout=default_timeout):
 
         fout.seek(0)
         nb = nbformat.read(fout, nbformat.current_nbformat)
+    os.remove(fname)
```
MaxBenChrist: this will throw an OSError on non NT systems, right? Because the tempfile does not exist anymore?

tests/units/scripts/test_run_tsfresh.py
```diff
@@ -31,6 +32,7 @@ def extract_features_mock(df, **kwargs):
 
     def tearDown(self):
         # Remove the directory after the test
+        os.chdir(self.curr_dir)
```
MaxBenChrist: what is the motivation behind the new 
`os.chdir`?

tests/integrations/test_notebooks.py
delyanr: I added `delete=False` to the tempfile constructor (see with-statement), so the file will exist on all systems before that `remove()` call. Unfortunately, I cannot test it in other systems.

MaxBenChrist: ok. Why do you remove the file outside of the context manager?

couldn't you do something like 

``` Python
    with tempfile.NamedTemporaryFile(mode='w+t', suffix=".ipynb", delete=False) as fout:
        args = ["jupyter", "nbconvert",
                "--to", "notebook", "--execute", execproc_timeout]
        if six.PY2:
            args += ["--ExecutePreprocessor.kernel_name=python2"]
        elif six.PY3:
            args += ["--ExecutePreprocessor.kernel_name=python3"]
        args += ["--output", fout.name, path]
        subprocess.check_call(args)

        fout.seek(0)
        nb = nbformat.read(fout, nbformat.current_nbformat)
        os.remove(fout.name)
```

tests/units/scripts/test_run_tsfresh.py
MaxBenChrist: Did you try this with xdist? (the parallel test runner plugin for pytest)

I am not sure what happens then

tests/integrations/test_notebooks.py
delyanr: I believe I'm not explaining it that well. Basically, NT does not allow you to delete a file which is currently being used by a process (used also means opened for read-only, etc). Not only that, but NT does not allow the same process to delete the file without first explicitly closing it. Therefore, the `close()` method must be first invoked - this happens automatically at the end of the with-statement. Now, the `remove()` method can only be called after the `close()` method, so it must be outside the context manager. Hope this explains it.

tests/units/scripts/test_run_tsfresh.py
delyanr: I have not tried xdist. From quick googling around I'm not sure I understand how to make it work. Could you please let me know how, and I will be happy to test? Thanks

delyanr: But also notice in the `setUp()` method in the same `test_run_tsfresh.py` file where I'm already setting the `self.curr_dir` property to the working directory of the underlying process. I don't believe this is controversial at all.

MaxBenChrist: xdist is the parallelization plugin of pytest. You can install it by `pip install pytest-xdist`. 

I was worried that only the mainprocess calls `setUp()` and that this would create problems if the child processes call `tearDown`. However, it seems that each process calls these methods individually (see e.g. https://github.com/pytest-dev/pytest-xdist/issues/15)

So, from the xdist point of view this change is fine

tests/integrations/test_notebooks.py
MaxBenChrist: Wow. Thank you for the detailed explanation, I haven't used windows in like 10-15 years, so these problems are completely new to me. 😃 

It blows my mind that the `NamedTemporaryFile` method is broken under windows NT. I googled around and your solution seem to be the prefered way to fix this.

Just out of curiosity, does this improve in python3?

PR #462, opened 2018-11-16
Use numpy for faster performance of Benjamini Hochberg

delyanr:
This is the first of a series of performance improvements to make `tsfresh` work faster. I will not change anything structurally and will try to keep the code as close to the original as possible.

Changelog:
* Vectorize the `benjamini_hochberg_test()` function using the `numpy` package. (We already have it as dependency anyway.)


MaxBenChrist:
thanks for the contribution @delyanr.  Looking forward to future prs :)

PR #469, opened 2018-12-03
Use numpy for better performance of significance tests

delyanr:
Changelog:
* Changed all **pandas.Series** methods in the significance tests to **numpy** methods which are faster (see below).

Benchmark example:

```
%timeit x.isnull().any()
573 µs ± 20.7 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

%timeit numpy.isnan(x.values).any()
15.8 µs ± 418 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

### DIFF
--- a/tsfresh/feature_selection/significance_tests.py
+++ b/tsfresh/feature_selection/significance_tests.py
@@ -69,10 +69,10 @@ def target_binary_feature_binary_test(x, y):
     y0, y1 = np.unique(y.values)
 
     # Calculate contingency table
-    n_y1_x0 = np.sum(y.values[x.values == x0] == y1)
-    n_y0_x0 = len(y.values[x.values == x0]) - n_y1_x0
-    n_y1_x1 = np.sum(y.values[x.values == x1] == y1)
-    n_y0_x1 = len(y.values[x.values == x1]) - n_y1_x1
+    n_y1_x0 = np.sum(y[x == x0] == y1)
+    n_y0_x0 = len(y[x == x0]) - n_y1_x0
+    n_y1_x1 = np.sum(y[x == x1] == y1)
+    n_y0_x1 = len(y[x == x1]) - n_y1_x1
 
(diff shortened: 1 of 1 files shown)


MaxBenChrist:
Hey @delyanr, thanks for the great next pr. 

From my experience, speed is not an issue in the tsfresh feature selection, feature extraction is a total different story. Was the feature selection too slow for your application?

Your improvements are sane, but they make the code a little less readable. Also, there could be bugs if x and y are not sorted in the same way. See the following example

```
 import pandas as pd
x = pd.Series([1, 2])
x = x[::-1] # revert x
y = pd.Series([10, 20])
# correct
print(x[y == 10])
# wrong
x.values[y.values == 10]
```
So, I can't merge your pr as it creates possible bugs. 

How can we process from here? You can add an unit test that shows that your implementation has this flaw (so, it should fail). And then afterwards we can fix it by sorting x and y.

delyanr:
Re speed: I definitely want to improve everything possible, since my dataset is massive and I need to re-run it often (due to changing features, new time series, etc.). Nevertheless feature selection is indeed generally fast, and this will be my last commit to it.

You make a very good point above (somehow I thought sorting will not be an issue). So I suggest we leave the **indexing** based on the current code (using pandas), but still incorporate the **numpy** methods for everything else, e.g. `np.isnan(x.values).any()` instead of `x.isnull().any()`. I believe if we were to sort x and y beforehand that will defeat the speed up.

Let me know if you think that works and I can change the commit.

MaxBenChrist:
> You make a very good point above (somehow I thought sorting will not be an issue).

I create some nasty bugs in the past because I was so used of pandas aligning the indices automatically when working with np.arrays, so I got sensible to this. 

> Let me know if you think that works and I can change the commit.

sure, we can replace the pandas with the numpy methods

MaxBenChrist:
Another great pr, thanks @delyanr

PR #470, opened 2018-12-03
Improve rolling using numpy concatenate

delyanr:
Changelog:
* Added new `_roll(a, shift)` method which uses `numpy.concatenate()` to reduce the overhead of `numpy.roll()` limiting the use to 1D arrays only.

Benchmark example:

```
%timeit numpy.roll(x, shift=5)
27.2 µs ± 713 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

%timeit _roll(x, shift=5)
10.6 µs ± 309 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

### DIFF
--- a/tests/units/feature_extraction/test_feature_calculations.py
+++ b/tests/units/feature_extraction/test_feature_calculations.py
@@ -74,9 +74,8 @@ class FeatureCalculationTestCase(TestCase):
                              f(pd.Series(input_to_f), *args, **kwargs), result))
 
     def test__roll(self):
-        np.random.seed(42)
         x = np.random.normal(size=30)
-        for shift in [1, 10, 30, 50, 150]:
+        for shift in [0, 1, 10, 11, 30, 31, 50, 51, 150, 151]:
             np.testing.assert_array_equal(_roll(x,  shift), np.roll(x,  shift))
             np.testing.assert_array_equal(_roll(x, -shift), np.roll(x, -shift))
 
--- a/tests/units/feature_extraction/test_feature_calculations.py
+++ b/tests/units/feature_extraction/test_feature_calculations.py
@@ -8,6 +8,7 @@ from builtins import range
 from random import shuffle
 from unittest import TestCase
 from tsfresh.feature_extraction.feature_calculators import *
+from tsfresh.feature_extraction.feature_calculators import _roll
 from tsfresh.feature_extraction.feature_calculators import _get_length_sequences_where
 from tsfresh.feature_extraction.feature_calculators import _estimate_friedrich_coefficients
 from tsfresh.feature_extraction.feature_calculators import _aggregate_on_chunks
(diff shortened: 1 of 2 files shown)


tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -31,6 +31,22 @@
 # todo: make sure '_' works in parameter names in all cases, add a warning if not
 
 
+def _roll(a, shift):
```
MaxBenChrist: can you please add a few unit tests for this method?

MaxBenChrist: Also, you designed this function with a certain goal in mind (I want to replace `np.roll` as its not fast enough). This reason needs to go into the docstring! This will help other people greatly.

Finally, it would be nice to know what the shift parameter does, like what happens if I use -1 or 1, in which direction is the method shifting the input.

MaxBenChrist: for the unit tests you could just extract some random arrays and show that it is equal to numpys.roll

delyanr:
Hi @MaxBenChrist - as per your previous comments, I improved the docstring and added a unittest. Hopefully this looks good now. Thanks! 👍

tests/units/feature_extraction/test_feature_calculations.py
```diff
@@ -72,6 +73,13 @@ def assertEqualPandasSeriesWrapper(self, f, input_to_f, result, *args, **kwargs)
                          msg="Not equal for pandas.Series: %s != %s" % (
                              f(pd.Series(input_to_f), *args, **kwargs), result))
 
+    def test__roll(self):
+        np.random.seed(42)
```
MaxBenChrist: no need for a seed ;)

tests/units/feature_extraction/test_feature_calculations.py
```diff
@@ -72,6 +73,13 @@ def assertEqualPandasSeriesWrapper(self, f, input_to_f, result, *args, **kwargs)
...
+        x = np.random.normal(size=30)
+        for shift in [1, 10, 30, 50, 150]:
```
MaxBenChrist: would also add odd and negative shift parameters, maybe use [-7, -1, 0, 1, 12, 511]

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -31,6 +31,50 @@
...
+    """
+    Roll 1D array elements. Improves the performance of numpy.roll() by reducing the overhead introduced from the
```
MaxBenChrist: great documentation!!

MaxBenChrist:
nearly there, if you remove the seed and change the shift range in the unit tests I will merge this

delyanr:
Hi @MaxBenChrist. I incorporated your latest comments. Note that the test tries both the positive and negative shifts based on the list, so no further action was required there.

MaxBenChrist:
great, thanks @delyanr !

PR #471, opened 2018-12-06
Update import statements to take train_test_split from model_selectio…

georgewambold:
…n instead of cross_validation


MaxBenChrist:
indeed, sklearn.cross_validation was depricated in 0.19.0, however we support scikit-learn down to 0.17.0. I can only merge your pr if I ramp up the dependency of scikit learn in tsfresh

georgewambold:
done 👍

PR #474, opened 2018-12-29
__init.py of feature_extraction modified

ayushpatidar:


MaxBenChrist:
Thanks for the pr, I am not sure why this change should be necessary, we are not exposing the feature_calculators to the user

Issue #486, opened 2019-02-05
Upgrade pyscaffold build dependency or remove it


rth:
Currently tsfresh requires pyscaffold < 2.6 that is not compatible with newer setuptools versions. 

As mentioned by @xhochy in https://github.com/conda-forge/tsfresh-feedstock/pull/2#issuecomment-459332997 this prevents from packaging v0.11.1 for conda forge,

> No, the problem is that these old pyscaffold versions don't work together with newer setuptools releases. But the newer setuptools are needed by other packages in the dependency chain. 

More generally this may be an issue for other installs as well, it's just that pip will not check global consistency of dependencies.

It would be nice to either upgrade pyscaffold build dependency or remove it.

TBH, I don't see the added value of pyscaffold in this case: tsfresh is a pure python package and so `setup.py` would be fairly straightforward with most configuration included in setup.cfg same as currently in any case. Not being compatible with latest setuptools is a larger concern IMO. Though I may be missing something.

xhochy:
Have a look at https://github.com/xhochy/fletcher/blob/master/setup.py for a `setup.py` that was written with recent standards. @rth you might want to submit a PR to get rid of `pyscaffold`, then we can get this back on `conda-forge` again.

rth:
Thanks for the feedback @xhochy , made a PR in #487

MaxBenChrist:
Agree with kicking out pyscaffold. We set tsfresh up with it and then it just stayed in.

xhochy:
@MaxBenChrist Can you cut a release so we get the version on `conda-forge` uptodate again?

MaxBenChrist:
Sure sure, will do a release tomorrow. Experience tells me that you should never deploy/release on a Friday night...

PR #487, opened 2019-02-11
MAINT Remove pyscaffold build dependency

rth:
Closes https://github.com/blue-yonder/tsfresh/issues/486

This removes the pyscaffold build dependency and generally updates the `setup.py`. The  `README.rst` is also no longer necessary I think as PyPi can now render markdown.

This adds `setuptools_scm` as a setup depdendency though as was (indirectly) suggested in #486


rth:
Currently  Travis CI fails with,
```
if [ $TRAVIS_PULL_REQUEST == false ] && ! [ $TRAVIS_BRANCH == 'master' ]; then py.test tests/units -n2; else python setup.py test; fi
/home/travis/build/blue-yonder/tsfresh/.eggs/setuptools_scm-3.2.0-py3.5.egg/setuptools_scm/git.py:68: UserWarning: "/home/travis/build/blue-yonder/tsfresh" is shallow and may cause errors
  warnings.warn('"%s" is shallow and may cause errors' % (wd.path,))
running test
error: error in setup.cfg: command 'test' has no such option 'addopts'
```

the first warning is due to https://github.com/pypa/setuptools_scm/issues/93


For the second I'm going to replace,
```
python setup.py test
```
by
```
pytest tests
```
directly, as the former is somewhat deprecated as far as I know and doesn't allow using any of the useful pytest options.

rth:
> Please copy all content from setup.cfg to into the `setup()` call [in `setup.py`]

Why? This is supported by setuptools since 2016 and is independent of the use of `pyscaffold` (https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files) . As mentioned by setuptools docs,

> This approach not only allows automation scenarios but also reduces boilerplate code in some cases.

rth:
Otherwise the CI failures appear to be the same as those on master..

MaxBenChrist:
Sorry, have been extremely busy the last weeks. Thanks for helping with this. 

The error looks unrelated to this pr, could be a change in a newer dask/pandas version

MaxBenChrist:
I will try to look into the failing unit test later

MaxBenChrist:
Thanks for taking care of this 👍

PR #516, opened 2019-03-22
Fix binder requirements

jamesmyatt:
Binder wasn't installing the package source. This version seems to work better, and avoids the need to build the binder requirements.txt file separately.

Please squash when merging! I've been editing directly in github so far. Otherwise I can squash it myself.


binder/requirements.txt
```diff
@@ -1,22 +1,3 @@
...
-requests>=2.9.1
-scikit-learn>=0.17.1
-scipy>=0.17.0
-seaborn>=0.7.1
-six>=1.10.0
-statsmodels>=0.6.1
-tqdm>=4.10.0
+-e .  # relative to cwd
+-r ../requirements.txt  # relative to here
+-r ../test-requirements.txt  # relative to here
```
MaxBenChrist: Isn't the order skewed? I think you should install requirements first, then test-requirements and finally the package.

further for the package, shouldn't you have -e ../. as well?

MaxBenChrist:
Thanks @jamesmyatt, do you by change know where to find the binder build logs?

MaxBenChrist:
Thx @jamesmyatt, works like a charm 👍

binder/requirements.txt
jamesmyatt: On the last point, no; pip is broken like that: https://github.com/pypa/pip/issues/3772

jamesmyatt: On the order, it doesn't really matter, but I think you want any specific version in the requirements.txt files to override the generic ones in the package.

PR #521, opened 2019-03-26
switch to scalar sum when features are not all independent

rhysimu:
np.sum sums all elements and presents a scalar C. Scalar C is then applied across the full range of K to form T. Tested it to make sure it didn't affect the use of T to compare p-values and determine relevance.


MaxBenChrist:
Sorry for the late answer, was quite busy last weeks.

It is not a good sign that you could make that change without adapting any unit tests. We have to add some tests for the feature selection

PR #526, opened 2019-04-09
fill up argument `max_timeshift` when calling function `roll_time_series`

ZJUguquan:
fill up argument `max_timeshift` when calling function `roll_time_series`


MaxBenChrist:
Thx @ZJUguquan !

PR #535, opened 2019-05-09
Change return type for sum_values to float

rwedge:


MaxBenChrist:
actually, the return type should be numerical as np.sum over integers returns int, see

```Python
np.sum([1, 3]).__class__ == numpy.int64
```

thanks @rwedge

Issue #537, opened 2019-05-16
cross_val_score works with FeatureAugmenter, but not with RelevantFeatureAugmenter


thanhd10:
Hi, I'm new to tsfresh and wanted to implement the RelevantFeatureAugmenter in a sklearn.pipeline to cross-validate, but encountered the problem, that it doesn't work with RelevantFeatureAugmenter, but with FeatureAugmenter. I reconstructed the problem with the robot_failure dataset:

~~~~ python
from tsfresh.examples import load_robot_execution_failures
from tsfresh.transformers import RelevantFeatureAugmenter, FeatureAugmenter, PerColumnImputer, FeatureSelector
from sklearn.svm import SVC
from sklearn.model_selection import cross_val_score
from sklearn.pipeline import make_pipeline
import pandas as pd

x, y = load_robot_execution_failures()
pipe = make_pipeline(RelevantFeatureAugmenter(column_id='id', column_sort='time'),
                     SVC())
pipe.set_params(featureaugmenter__timeseries_container=x)

xfit = pd.DataFrame(index=y.index)
score = cross_val_score(pipe, xfit, y, cv=4)
~~~~
When I run the program, I get the following error message:
~~~~ python
File "/anaconda3/lib/python3.7/site-packages/tsfresh/transformers/relevant_feature_augmenter.py", line 260, in fit
    X_augmented = self.feature_extractor.transform(X_tmp)
  File "/anaconda3/lib/python3.7/site-packages/tsfresh/transformers/feature_augmenter.py", line 185, in transform
    timeseries_container_X = restrict_input_to_index(self.timeseries_container, self.column_id, X.index)
  File "/anaconda3/lib/python3.7/site-packages/tsfresh/utilities/dataframe_functions.py", line 203, in restrict_input_to_index
    df_or_dict_restricted = df_or_dict[df_or_dict[column_id].isin(index)]
  File "/anaconda3/lib/python3.7/site-packages/pandas/core/frame.py", line 2688, in __getitem__
    return self._getitem_column(key)
  File "/anaconda3/lib/python3.7/site-packages/pandas/core/frame.py", line 2695, in _getitem_column
    return self._get_item_cache(key)
  File "/anaconda3/lib/python3.7/site-packages/pandas/core/generic.py", line 2489, in _get_item_cache
    values = self._data.get(item)
  File "/anaconda3/lib/python3.7/site-packages/pandas/core/internals.py", line 4124, in get
[...]

PR #540, opened 2019-05-22
Refactor RelevantFeatureAugmenter to enable cloning

pnb:
Addresses #537 by moving side effects in RelevantFeatureAugmenter from the constructor to the `fit()` function, which allows scikit-learn to clone the transformer. Transformer cloning is necessary for cross-validation and grid search functionality in sklearn.

This enables the following simple example to run:

```python
import pandas as pd
from sklearn.pipeline import Pipeline
from sklearn import model_selection
from sklearn.ensemble import RandomForestClassifier
from tsfresh.examples import load_robot_execution_failures
from tsfresh.transformers import RelevantFeatureAugmenter

df_ts, y = load_robot_execution_failures()
X = pd.DataFrame(index=y.index)

augmenter = RelevantFeatureAugmenter(column_id='id', column_sort='time', timeseries_container=df_ts)
pipeline = Pipeline([('augmenter', augmenter),
                     ('classifier', RandomForestClassifier(random_state=1))])

print(model_selection.cross_val_score(pipeline, X, y, cv=4))
```

It should produce something like `[0.30434783 1.         1.         1.        ]`

### DIFF
--- a/tsfresh/transformers/relevant_feature_augmenter.py
+++ b/tsfresh/transformers/relevant_feature_augmenter.py
@@ -184,30 +184,28 @@ class RelevantFeatureAugmenter(BaseEstimator, TransformerMixin):
                     else regression.
         :type ml_task: str
         """
-
-        self.feature_extractor = FeatureAugmenter(column_id=column_id, column_sort=column_sort, column_kind=column_kind,
-                                                  column_value=column_value,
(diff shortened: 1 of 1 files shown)


MaxBenChrist:
Sorry for the late response and thanks for the pr @pnb

Issue #485, opened 2019-01-28
TypeError: Cannot cast array data from dtype('float64') to dtype('<U32') according to the rule 'safe'`


matej14086:
1.OS:  Ubuntu 18.04.1 LTS
2. tsfresh version = 0.11.0
    numpy==1.16.0

I keep getting following error:
```
  File "<ipython-input-122-1f1001a26d12>", line 1, in <module>
    tsfresh.extract_features(df, column_id = "ID")

  File "/home/matej/anaconda3/lib/python3.6/site-packages/tsfresh/feature_extraction/extraction.py", line 152, in extract_features
    distributor=distributor)

  File "/home/matej/anaconda3/lib/python3.6/site-packages/tsfresh/feature_extraction/extraction.py", line 233, in _do_extraction
    function_kwargs=kwargs)

  File "/home/matej/anaconda3/lib/python3.6/site-packages/tsfresh/utilities/distribution.py", line 149, in map_reduce
    result = list(itertools.chain.from_iterable(result))

  File "/home/matej/anaconda3/lib/python3.6/site-packages/tqdm/_tqdm.py", line 1002, in __iter__
    for obj in iterable:

  File "/home/matej/anaconda3/lib/python3.6/multiprocessing/pool.py", line 735, in next
    raise value

TypeError: Cannot cast array data from dtype('float64') to dtype('<U32') according to the rule 'safe'`
```

when I run:
```
import tsfresh
import pandas as pd
import numpy as np

a = np.arange(10)
df = pd.DataFrame(a,columns = ["value"])
ids = [1,1,1,1,1,5,5,5,5,5]
df['ID'] = ids

tsfresh.extract_features(df, column_id = "ID")
```
Same error occurs when I try to run it with real data.

earthgecko:
Hi @matej14086 

I have tried producing your reported issue with the versions you reported and I have not been able to reproduce the issue from the code snippet above.  I tested with Python 3.6.4 and Python 3.5 and the code runs fine with ```tsfresh==0.11.0``` and also ```tsfresh==0.11.1``` along with numpy 1.14.1 through to numpy 1.16.0

The error suggests that numpy is complaining the data is not numeric but string.  Could you try and run this is a normal python terminal (not ipython) and just check that you get the same result.  Perhaps there is something unseen in a copy and paste.  If you still get errors, could you also report what version of pandas you are using as well.

soccerdroid:
Had the same issue

hantoine:
[...]

PR #499, opened 2019-02-23
Drop support for legacy Python 2.7

hugovk:
Fixes #497.

Adds `python_requires` to help pip.

Also upgrade syntax with https://github.com/asottile/pyupgrade.

PR #562, opened 2019-07-31
Update pandas version to latest

Rictus:
Tried to reproduce [this](https://github.com/blue-yonder/tsfresh/issues/485) error with `pandas==0.24.0` : Got the `TypeError`. 
On `pandas==0.25.0`, no errors. 

After running tests, I understand the story is a bit more complicated with python 3.5.2 and 2.7

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,10 +14,10 @@ install:
 jobs:
   include:
     - stage: Run unit tests and deploy
-      env: NUMPY="latest", PANDAS="0.23.4", SCIKITLEARN="latest", DASK="latest", DISTRIBUTED="latest", SCIPY="latest"
+      env: NUMPY="latest", PANDAS="latest", SCIKITLEARN="latest", DASK="latest", DISTRIBUTED="latest", SCIPY="latest"
       before_script:
         - pip install --upgrade numpy
-        - pip install pandas===0.23.4
+        - pip install --upgrade pandas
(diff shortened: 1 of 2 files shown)


nphilou:
`pandas==0.25.x` only supports Python 3.5.3 and higher (from pandas release notes).

#499 will hopefully solve this problem.

dbarbier:
Please use versioned dependencies in `requirements.txt` only when strictly needed, they do more harm than good.
Due to current pandas dependency, one is assured to not have pandas 0.24.x installed along tsfresh, which is good.  But it is very difficult to install tsfresh 0.12.0 with pandas 0.25.x (with conda or pip), although this works perfectly fine.
On the other hand, if tsfresh did not specify any pandas version, users could easily use tsfresh 0.12.0 with pandas 0.25.x.  And if they had pandas 0.24 installed, that would indeed cause trouble, but at least this is easily fixable.

In my opinion, versioned dependencies should only be used if your package depends on some features which have been added in some specific version (this can be seen as an API version number).  So please, in this PR, do not depend on pandas >= 0.25, but on the minimum required pandas version (could be 0.20.3 or 0.17 or...).

nils-braun:
All the changes of this PR are now in #594 plus the additional fix for the tests.
I am therefore closing this PR and thanking @Rictus for the work (which is of course preserved)!

PR #567, opened 2019-08-16
Improve performance with Numpy's sum function

koho:
The `change_quantiles` function need to sum the bool index array. Currently it's using the builtin `sum` function. However, NumPys `np.sum` will be faster on arrays.

I ran some test on it.
```python
import timeit
stmt = """change_quantiles(np.random.rand(300000) * 1000, 0.1, 0.2, False, 'mean')"""
print(timeit.timeit(stmt=stmt, setup='import numpy as np;from tsfresh.feature_extraction.feature_calculators import change_quantiles', number=1))
```
With the builtin `sum` function it takes 0.5s, while NumPys `np.sum` takes only 0.02s.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,10 +5,19 @@ Changelog
 tsfresh uses `Semantic Versioning <http://semver.org/>`_
 
 
-Unreleased
-==========
+Version 0.12.0
+==============
+
+- fixed bugs
+    - wrong calculation of friedrich coefficients
+    - feature selection selected too many features
+    - an ignored max_timeshift parameter in roll_time_series
+- add deprecation warning for python 2
+- added support for index based features
+- new feature calculator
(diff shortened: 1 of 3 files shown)


MaxBenChrist:
Hi @koho, can you merge the current master so we see that all tests pass? Then we can merge this into master. 👍

dbarbier:
Hello @koho there is another instance of `sum` which could be replaced by `np.sum` in `index_mass_quantile`, maybe you could fix it too?

dbarbier:
Thanks @koho for your quick reply.  @MaxBenChrist These changes are really trivial, could you please review this PR?  These changes provide a nice speedup with `default_fc_parameters=EfficientFCParameters()`.

nils-braun:
Nice, thanks for the fix and sorry for the very long time to merge it!

PR #568, opened 2019-08-17
Drop python 2.7 support - add travis for python 3.7

Froskekongen:
Updated https://github.com/blue-yonder/tsfresh/pull/499. Here, I also update python versions to 3.6 and 3.7.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,21 @@ jobs:
       python: 3.5.2
 
     - stage: Run unit tests and deploy
+      env: NUMPY="latest", PANDAS="0.23.4", SCIKITLEARN="latest", DASK="latest", DISTRIBUTED="latest", SCIPY="latest"
+      before_script:
+        - pip install --upgrade numpy
+        - pip install pandas===0.23.4
+        - pip install --upgrade scikit-learn
+        - pip install --upgrade dask
+        - pip install --upgrade distributed
+        - pip install --upgrade scipy
+        - pip list
+      script:
+        - sed -i 's/\-n auto/\-n 2/g' setup.cfg
+        - "if [ $TRAVIS_PULL_REQUEST == false ] && ! [ $TRAVIS_BRANCH == 'master' ]; then pytest tests/units -n2; else pytest tests; fi"
+      python: 3.6
+
+    - stage: Run unit tests and deploy
       env: NUMPY="1.10.4", PANDAS="0.20.3", SCIKITLEARN="0.18.0", DASK="0.15.2", DISTRIBUTED="1.18.3", SCIPY="1.2.0"
       before_script:
         - pip install numpy==1.10.4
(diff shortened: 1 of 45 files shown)


Froskekongen:
@MaxBenChrist: I continued the work in https://github.com/blue-yonder/tsfresh/pull/499 to remove python 2.7 support. I tried adding pandas 0.25 at the same time, but two tests fails, so it should be fixed in a subsequent PR.

Froskekongen:
@MaxBenChrist: Any reason for holding back this PR?

tomzx:
I'd also like to see this PR being merged so that we can move forward and support pandas=>0.25. 

@Froskekongen It's already a big PR, but it would be nice if it also included relaxing the constraint on pandas, at best by adding `!=0.24.0`, although I wouldn't mind not supporting old versions of pandas (<=0.25.0).

.travis.yml
```diff
@@ -18,17 +16,15 @@ jobs:
...
         - pip install --upgrade dask
         - pip install --upgrade distributed
         - pip install --upgrade scipy
         - pip list
       script:
         - sed -i 's/\-n auto/\-n 2/g' setup.cfg
-        - "if [ $TRAVIS_PULL_REQUEST == false ] && ! [ $TRAVIS_BRANCH == 'master' ]; then py.test tests/units -n2; else python setup.py test; fi"
-      python: 2.7
-      after_success:
-        - coveralls
```
nils-braun: do we still have a coverall somewhere now?

nils-braun:
So, first of all thanks to @Froskekongen and @hugovk  for this PR(s). I (also) think it is a good idea to drop 2.7 support.
I think we can and should separate the two parts - pandas and python version - if possible so it is easier to revert if needed. 
I am adding some review comments, but most of them are minor things.

requirements.txt
```diff
@@ -1,14 +1,11 @@
...
-scikit-learn>=0.19.0,<0.21.0; python_version <= '2.7'
-scikit-learn>=0.19.0; python_version > '3'
-future>=0.16.0
-six>=1.10.0
+scikit-learn>=0.19.0
 tqdm>=4.10.0
-ipaddress>=1.0.18; python_version <= '2.7'
 dask>=0.15.2
 distributed>=1.18.3
+future>=0.11
```
nils-braun: Do we still need it?
The only occurrence I still see is in `test_selection.py` and there we can also remove it, or?

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -108,7 +106,7 @@ def _get_length_sequences_where(x):
 
 
 def _estimate_friedrich_coefficients(x, m, r):
-    """
+    r"""
```
nils-braun: Just out of curiosity: why did you choose this and not to add "\\" as in the other cases?

tsfresh/feature_extraction/extraction.py
```diff
@@ -5,8 +5,6 @@
 This module contains the main function to interact with tsfresh: extract features
 """
 
-from __future__ import absolute_import, division
-
 import logging
 import sys
 import warnings
```
nils-braun: There is still an import of "six" and a check for the python version in this file. This should also be removed.

Froskekongen: Done.

tsfresh/feature_extraction/feature_calculators.py
Froskekongen: It doesn't seem like this is done in a uniform way throughout the codebase.

requirements.txt
Froskekongen: Agreed.

.travis.yml
Froskekongen: I added this now.

nils-braun:
Nice Nice Nice.
We finally get rid of it :-)

PR #570, opened 2019-08-26
Replace Benjamini-Hochberg implementation with statsmodels implementation

pnb:
**NOTE: This PR may have significant effects on results (probably for the better).**

I noticed surprising changes in the number of features that were selected as relevant, given only small changes to my datasets. I traced the problem down to the Benjamini-Hochberg implementation, which appears to have a couple problems. This PR fixes it.

I tested the B-H implementation with this script:
```python
from tsfresh.feature_selection.benjamini_hochberg_test import benjamini_hochberg_test
import pandas as pd


# https://www.real-statistics.com/hypothesis-testing/familywise-error/benjamini-hochberg-and-benjamini-yekutieli-tests/
example1 = pd.DataFrame({'p_value': [.00356, .01042, .01208, .02155, .03329, .11542]})
print(benjamini_hochberg_test(example1, False, .05))

# https://www.statisticshowto.datasciencecentral.com/benjamini-hochberg-procedure/
# This webpage uses < instead of <= as in (1) of Benjamini & Yekutieli, 2001 and (1) of Benjamini &
# Hochberg, 1995; hence, results differ for one case where p exactly equals the adjusted threshold
example2 = pd.DataFrame({'p_value': [.001, .008, .039, .041, .042, .060, .074, .205, *[1] * 17]})
print(benjamini_hochberg_test(example2, False, .25))

# Nothing selected
example3 = pd.DataFrame({'p_value': [.04, .05, .06]})
print(benjamini_hochberg_test(example3, False, .05))

# Exactly one selected
example4 = pd.DataFrame({'p_value': [.001, .9, .9, .9]})
print(benjamini_hochberg_test(example4, False, .05))

# Everything selected with .01 threshold
example5 = pd.DataFrame({'p_value': [.001, .002, .003, .004, .005]})
print(benjamini_hochberg_test(example5, False, .01))
```

Output did not match expectations with the existing B-H implementation. There were many features that should have been selected that were not. For example, in the first case, five of the six features should be selected, but none were. In the last example, all should be selected, but none were.

Original output:
```
   p_value  relevant
0  0.00356     False
1  0.01042     False
2  0.01208     False
3  0.02155     False
4  0.03329     False
5  0.11542     False
    p_value  relevant
0     0.001      True
1     0.008     False
2     0.039     False
3     0.041     False
4     0.042     False
5     0.060     False
6     0.074     False
7     0.205     False
22    1.000     False
21    1.000     False
20    1.000     False
19    1.000     False
18    1.000     False
17    1.000     False
16    1.000     False
12    1.000     False
14    1.000     False
13    1.000     False
23    1.000     False
11    1.000     False
10    1.000     False
9     1.000     False
8     1.000     False
15    1.000     False
24    1.000     False
   p_value  relevant
0     0.04     False
1     0.05     False
2     0.06     False
   p_value  relevant
0    0.001      True
1    0.900     False
2    0.900     False
3    0.900     False
   p_value  relevant
0    0.001     False
1    0.002     False
2    0.003     False
3    0.004     False
4    0.005     False
```

I found two issues.

1. The vector of comparison values was not calculated correctly with `independent=False` (it was not a vector)
2. `k_max` was the first non-selected feature, rather than the max of selected features

I compared code from `statsmodels.stats.multitest` and the source papers mentioned in example code above, and changed the implementation accordingly.

The new output matches expectations:
```
   p_value  relevant
0  0.00356      True
1  0.01042      True
2  0.01208      True
3  0.02155      True
4  0.03329      True
5  0.11542     False
    p_value  relevant
0     0.001      True
1     0.008      True
2     0.039      True
3     0.041      True
4     0.042      True
5     0.060      True
6     0.074     False
7     0.205     False
22    1.000     False
21    1.000     False
20    1.000     False
19    1.000     False
18    1.000     False
17    1.000     False
16    1.000     False
12    1.000     False
14    1.000     False
13    1.000     False
23    1.000     False
11    1.000     False
10    1.000     False
9     1.000     False
8     1.000     False
15    1.000     False
24    1.000     False
   p_value  relevant
0     0.04     False
1     0.05     False
2     0.06     False
   p_value  relevant
0    0.001      True
1    0.900     False
2    0.900     False
3    0.900     False
   p_value  relevant
0    0.001      True
1    0.002      True
2    0.003      True
3    0.004      True
4    0.005      True
```


nils-braun:
@MaxBenChrist This is a PR for you - I do not have enough insight into this. Could you comment if you have time?

kempa-liehr:
What about using the statsmodel implementation?

https://www.statsmodels.org/stable/generated/statsmodels.stats.multitest.multipletests.html

pnb:
The statsmodel implementation supports Benjamini-Yekutieli as well as Benjamini-Hochberg, so it should be fine to replace `hypothesis_independent=True` with the statsmodels B-H implementation and `...=False` with the B-Y implementation (assuming I have the difference correct and not reversed).

I can probably update this PR with that soon.

pnb:
@kempa-liehr I made that switch and updated the pull request. I also removed the unit test for p-value correction since that is covered by statsmodels.

This works fine on my datasets, and is probably preferable (in my view) to a custom implementation since the statsmodels package is widely utilized and tested, and also because it makes future flexibility in the p-value correction method quite easy. Instead of `hypotheses_independent`, there could be a string parameter that statsmodels supports to specify a bunch of different correction options.

FWIW, I think a lot of the issues I was having were not due to the existing implementation, but because of the default `hypotheses_independent = True` setting. Apparently that can be quite conservative, and eliminate all of the features even when many of them are useful. This [Stack Overflow discussion](https://stats.stackexchange.com/questions/63441/what-are-the-practical-differences-between-the-benjamini-hochberg-1995-and-t) and the [slides it links to](http://www.stat.cmu.edu/~genovese/talks/hannover1-04.pdf) (slide 37) seem to suggest B-H is quite robust and B-Y might be too conservative.

kempa-liehr:
@pnb, thanks for implementing the switch to the statsmodel implementations.

Please have a look at my review comments. 

Also, thanks for the interesting links. We might need to update the documentation to discuss the benefits and consequences of using ```hypotheses_independent=True``` vs. ```hypotheses_independent=False```. I agree, that in a typical machine learning pipeline, which combines the tsfresh features with another robust ensemble or regularisation method the setting ```hypotheses_independent=True``` works quite well. However, in the context of time-series feature extraction and the application of FDR control the more conservative choice  ```hypotheses_independent=False``` is the better default.

pnb:
@kempa-liehr I added the test back in and modified it. One of the tests failed after switching to statsmodels, because of a bug in the original B-Y implementation. Specifically, the B-Y (independent = False) test with p-values = [.1, .1., .05] failed because the original B-Y implementation was looking for the first False in the sorted array, rather than the last. The expected result in the unit test was [False, False, False], but should have been [True, True, True]. Note that my original fix was also incorrect, so probably another good reason to go with statsmodels...

I checked this case with some various R packages and manual calculation as well, to verify that the statsmodels result is indeed correct. The code is below, if you're curious.

My apologies if I did not address the review comments properly -- I'm not super experienced with GitHub and I'm not sure if I properly found them.

```R
p = c(.1, .1, .05)
p = sort(p)

library(sgof)
str(BY(p, alpha = .20))  # Does not support negative correlation (less conservative)

library(fda.usc)
pvalue.FDR(p, dep = -1)  # dep = -1 to allow negative correlation

library(stats)
p.adjust(p, method='BY')  # Allows for negative correlation (matches fda.usc result)

# B-Y
cm = sum(c(1 / 3, 2 / 3, 3 / 3))  # To allow negative correlation
adj_p = c(1 / 3, 2 / 3, 3 / 3) * .20 / cm
adj_p
p <= adj_p  # The original TSFRESH B-Y implementation looks for the first False instead of the last
```

kempa-liehr:
Hi png,
Thanks for the pull-request. I agree, that it will impact the feature selection and should be included into the next major commit. 
@nils-braun, could you please exclude this commit from the planned minor commit?

nils-braun:
@kempa-liehr As you have reviewed this PR, I can also now just skip the patch release and go for the minor already :-)
( your definition of major/minor is not my definition of major/minor ;-) )

PR #572, opened 2019-09-05
Add set_property import

d-chambers:
Add the `set_property` import so users know which module it comes from for defining custom features.


MaxBenChrist:
thx @d-chambers

Issue #542, opened 2019-06-02
Almost useless feature called mean_change


thautwarm:
https://github.com/blue-yonder/tsfresh/blob/00884fb2a7ed3a364ebe27f140be8bf425fcde86/tsfresh/feature_extraction/feature_calculators.py#L528

The result is equivalent to (`x_last - x_first) / (n - 1)`.

If you do want this as a feature, `(x[-1] - x[0])/(len(x) - 1)` instead.

nils-braun:
The optimized function is included, thanks to @blu3r4y 
It might still be a useless feature, but you never know and that is why we have a feature selection ;-)

PR #574, opened 2019-10-11
Optimize mean_change (fixes issue #542) and correct documentation

blu3r4y:
This uses a simpler formula for calculating `mean_change` that was already proposed in #542. The proof for this simplification follows. Moreover, I fixed the documentation, because we actually divide by `n - 1` because a series with `n` elements only has `n - 1` differences. I also added a few test cases to check some corner cases.

image


dbarbier:
Likewise `mean_second_derivative_central` could return `(x[-1] - x[-2] - x[1] + x[0]) / (2 * (len(x) - 2))`

nils-braun:
Makes sense to do it like this! Thanks.
@dbarbier do you want to include the fix you mentioned as another PR?

dbarbier:
Okay, will do, thanks for merging these PRs.

Issue #557, opened 2019-07-14
UserWarning: linear_trend_timewise requires the data to have a index of type <class 'pandas.core.indexes.datetimes.DatetimeIndex'>. Results will not be calculated


forhonourlx:
Hi,

I am suffering this warning, could somebody know how to solve this problem?
/home/simon/anaconda3/lib/python3.6/site-packages/tsfresh/feature_extraction/extraction.py:380: UserWarning: linear_trend_timewise requires the data to have a index of type <class 'pandas.core.indexes.datetimes.DatetimeIndex'>. Results will not be calculated

Thanks.

advahad:
Having the same issue, any luck?

xenmind:
Pretty sure this is explained here: https://github.com/blue-yonder/tsfresh/blob/master/notebooks/feature_extraction_with_Datetime_index.ipynb A feature set needs date/time index which you aren't providing.

forhonourlx:
@xenmind Thank you for your help.
I am following this example [https://github.com/blue-yonder/tsfresh/blob/master/notebooks/timeseries_forecasting_google_stock.ipynb](url)

Codes:
`ser_singular = pd.Series(data_np[i_,:,ii_], index=self.date_pd)
                print('instrument:',i_,',idx:',ii_,',ser_singular:\n',ser_singular)

                df_shift, y = make_forecasting_frame(ser_singular, kind="price", max_timeshift=tsfresh_rollback, rolling_direction=1)
                print('instrument:',i_,',idx:',ii_,',df_shift:\n',df_shift.head())
                #print('y:\n',y.head())
                X = tsfresh.extract_features(df_shift, column_id="id", column_sort="time", column_value="value", impute_function=impute, show_warnings=False)
                print('instrument:',i_,',idx:',ii_,',features:\n',X.head())`

Then got:
/home/simon/anaconda3/lib/python3.6/site-packages/tsfresh/feature_extraction/extraction.py:380: UserWarning: linear_trend_timewise requires the data to have a index of type <class 'pandas.core.indexes.datetimes.DatetimeIndex'>. Results will not be calculated
  "not be calculated".format(function_name, index_type)

Is there anything to deal when tsfresh.utilities.dataframe_functions.make_forecasting_frame() ？

forhonourlx:
instrument: 0 ,idx: 5 ,ser_singular:
 Date
2013-09-15 21:16:00     0.000000
...
2013-09-20 20:57:00    48.338407
[...]

PR #576, opened 2019-10-16
Preserve index when column_value and column_kind not provided

jeffzi:
Hello,

This PR solves #557.  The support for  index based features introduced in tsfresh 0.12.0 only works when `column_kind` and `column_value` are passed to `extract_features()`. 

Here is an example of the issue:

```python
import pandas as pd
from tsfresh.feature_extraction import extract_features
from tsfresh.feature_extraction.settings import TimeBasedFCParameters  

df = pd.DataFrame({"id": ["a", "a", "a", "a", "b", "b", "b", "b"], 
                   "value": [1, 2, 3, 1, 3, 1, 0, 8],
                   "kind": ["temperature", "temperature", "pressure", "pressure",
                            "temperature", "temperature", "pressure", "pressure"]})

# Without pd.DatetimeIndex -> empty (as expected)
extracted = extract_features(df, column_id="id", column_value='value', column_kind='kind',
                             default_fc_parameters=TimeBasedFCParameters())
extracted.info()
#> <class 'pandas.core.frame.DataFrame'>
#> Index: 0 entries
#> Empty DataFrame
```

```python
# With pd.DatetimeIndex -> works
time_index = pd.DatetimeIndex(
    ['2019-03-01 10:04:00', '2019-03-01 10:50:00', '2019-03-02 00:00:00', '2019-03-02 09:04:59',
     '2019-03-02 23:54:12', '2019-03-03 08:13:04', '2019-03-04 08:00:00', '2019-03-04 08:01:00']
)
df = df.set_index(time_index).sort_index()
extracted = extract_features(df, column_id="id", column_value='value', column_kind='kind',
                             default_fc_parameters=TimeBasedFCParameters())
extracted.info()
#> <class 'pandas.core.frame.DataFrame'>
#> Index: 2 entries, a to b
#> Data columns (total 10 columns):
#> pressure__linear_trend_timewise__attr_"intercept"       2 non-null float64
#> pressure__linear_trend_timewise__attr_"pvalue"          2 non-null float64
#> pressure__linear_trend_timewise__attr_"rvalue"          2 non-null float64
#> pressure__linear_trend_timewise__attr_"slope"           2 non-null float64
#> pressure__linear_trend_timewise__attr_"stderr"          2 non-null float64
#> temperature__linear_trend_timewise__attr_"intercept"    2 non-null float64
#> temperature__linear_trend_timewise__attr_"pvalue"       2 non-null float64
#> temperature__linear_trend_timewise__attr_"rvalue"       2 non-null float64
#> temperature__linear_trend_timewise__attr_"slope"        2 non-null float64
#> temperature__linear_trend_timewise__attr_"stderr"       2 non-null float64
#> dtypes: float64(10)
#> memory usage: 176.0+ bytes
```

```python
# With pd.DatetimeIndex but without specifying column_value and column_kind 
# -> empty without this PR
extracted = (df
             .groupby('kind')
             .apply(lambda g : extract_features(g, column_id="id",
                          default_fc_parameters=TimeBasedFCParameters()))
            )
extracted.info()
#> <class 'pandas.core.frame.DataFrame'>
#> MultiIndex: 0 entries
#> Empty DataFrame
```

<sup>Created on 2019-10-16 by the [reprexpy package](https://github.com/crew102/reprexpy)</sup>

The bug comes from the index not being preserved when melting the dataframe in `_normalize_input_to_internal_representation`, specifically when `column_kind` and `column_value` are None.

### DIFF
--- a/tsfresh/utilities/dataframe_functions.py
+++ b/tsfresh/utilities/dataframe_functions.py
@@ -321,8 +321,13 @@ def _normalize_input_to_internal_representation(timeseries_container, column_id,
             column_value = "_values"
             column_sort = "_sort"
             sort = range(len(timeseries_container))
-            timeseries_container = pd.melt(timeseries_container, id_vars=[column_id],
+
+            timeseries_container.index.name = 'index'
+            timeseries_container = pd.melt(timeseries_container.reset_index(),
+                                           id_vars=['index', column_id],
                                            value_name=column_value, var_name=column_kind)
+            timeseries_container = timeseries_container.set_index('index')
+
             timeseries_container[column_sort] = np.tile(sort, (len(timeseries_container) // len(sort)))
 
     # Check kind column
--- a/tsfresh/utilities/dataframe_functions.py
+++ b/tsfresh/utilities/dataframe_functions.py
@@ -309,26 +309,22 @@ def _normalize_input_to_internal_representation(timeseries_container, column_id,
 
     if column_kind is None and column_value is None:
         if column_sort is not None:
-            column_kind = "_variables"
-            column_value = "_values"
             sort = timeseries_container[column_sort].values
-            timeseries_container = pd.melt(timeseries_container.drop(column_sort, axis=1),
-                                           id_vars=[column_id],
-                                           value_name=column_value, var_name=column_kind)
-            timeseries_container[column_sort] = np.tile(sort, (len(timeseries_container) // len(sort)))
+            timeseries_container = timeseries_container.drop(column_sort, axis=1)
         else:
-            column_kind = "_variables"
-            column_value = "_values"
-            column_sort = "_sort"
             sort = range(len(timeseries_container))
+            column_sort = "_sort"
 
-            timeseries_container.index.name = 'index'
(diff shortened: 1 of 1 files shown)


nils-braun:
Thanks for the fix.
Without further checking (sorry, not much time now, will do tomorrow): don't you end up with the additional index column?

jeffzi:
The index doesn't appear in the results of `extract_features`.

The line below sets the index after the `melt`, therefore the index is not in the column list.
```python
timeseries_container = timeseries_container.set_index('index')
```

nils-braun:
Ah sorry sorry. I forgot that `drop=True` is the standard in `set_index` (thought it would be drop=False).
Then this is fine!

nils-braun:
Do we also want that for the other meld in the lines above? (if you want you can also pull the two common lines with the meld an the tile out of the `if` branches to reduce code duplication)

jeffzi:
Yes I think the other melt should be modified as well. I didn't test the column_sort argument before, my bad !

I've deduplicated the code as you suggested. Let me know if that's ok :)

nils-braun:
Looks good as far as I can tell!

PR #578, opened 2019-10-21
Fix off-by-one error in longest strike features (fixes #577)

### LINKED ISSUE #577: Off-by-one error in longest strike features
The two features `longest_strike_below_mean` and `longest_strike_above_mean` are implemented in a "less-than" and "greater-than" fashion, which yields incorrect results.

Here is a MWE where µ = 3 and the longest strike below and above is 2 - `tsfresh` yields 3.

```
assert fc.longest_strike_below_mean([1, 2, 3, 4, 5]) == 2
assert fc.longest_strike_above_mean([1, 2, 3, 4, 5]) == 2
```

blu3r4y:
Use less and greater operators instead of less-or-equal and greater-or-equal operators in `longest_strike_below_mean` and `longest_strike_above_mean` to fix issue #577. I additionally added a few more test cases on this corner case.


MaxBenChrist:
Thanks for taking care of this!

PR #582, opened 2019-11-12
add @set_property("input", "pd.Series") when needed

dbarbier:
This ensures that _do_extraction_on_chunk will pass a Series argument
and not a Numpy array, and avoids a Series->ndarray->Series conversion.


nils-braun:
Nice!

Issue #529, opened 2019-04-16
get_moment function integer overflow on windows


albertgt:
1. Operating system
Windows 10 64-bit
Version	10.0.17134 Build 17134

2. tsfresh version  0.11.2

Code snippet which reproduces the problem/bug:

**in tsfresh/feature_extraction/feature_calculators.py**
_in function def get_moment(y, moment):_
> return y.dot(np.arange(len(y))**moment) / y.sum()

_the ** operator on windows incorrectly calculates:
assuming len(y) is 4096 and moment is 3_
> np.arange(4096)**3
array([         0,          1,          8, ..., -150884379, -100614152,
        -50319361], dtype=int32)

_but on linux and mac osx, the same code correctly calculates:_
> np.arange(4096)**3
array([          0,           1,           8, ..., 68568592357,
       68618862584, 68669157375])

_i believe the fix is to explicitly use numpy.float_power instead of the ** operator.
on all OS's this correctly calculates:_
> array([0.00000000e+00, 1.00000000e+00, 8.00000000e+00, ...,
       6.85685924e+10, 6.86188626e+10, 6.86691574e+10])

MaxBenChrist:
Wow. I can reproduce the behaviour under windows. However, I think it makes sense to calculate everything inside the numpy world. 

May I ask, how did you found this? Were you tracking differences in the calculated features between unix and windows?

albertgt:
Yes, my good friend a data scientist was using tsfresh to develop a model in aws sagemaker and it worked great there (I’m guessing sagemaker is amazon ec2 Linux) but then as we worked together to productionize it, we tested on our windows laptops and got different answers than sagemaker. He narrowed the issue down to the get_moment function and as we debugged it we found the integer overflow. When we uploaded the code to an ec2 Linux server for further testing we found it worked fine just like sagemaker. I tried it on a Mac and it worked fine too. But every windows os we used had the integer overflow issue.

nils-braun:
This was fixed thanks to @dbarbier !

PR #584, opened 2019-11-12
ensure that get_moment() uses float computations

dbarbier:
According to #529, np.arange(4096)**3 overflows on Windows 10 because
dtype is set to int32.  Specify dtype=float to avoid overflows.


nils-braun:
Thanks for the fix, looks good.

PR #587, opened 2019-11-16
Optimize mean_second_derivative_central

dbarbier:
This is similar to #542, mean_second_derivative_central has a simpler formula.
Also fix its docstring as had been done in #574.


nils-braun:
:+1:

PR #588, opened 2019-11-17
Avoid division by zero in energy_ratio_by_chunks

### LINKED ISSUE #547: Issue with energy_ratio_by_chunks
There is an error in feature_extraction/feature_calculators.py

When computing the energy of each segment, and in the special case that the time series array is a constant 0.0, full_series_energy would have a value of 0.0 causing a division by zero error.

This can be resolved easily by setting full_series_energy=1.0 when the value is 0.0

dbarbier:
Fix #547.

### DIFF
--- a/tests/units/feature_extraction/test_feature_calculations.py
+++ b/tests/units/feature_extraction/test_feature_calculations.py
@@ -1035,6 +1035,14 @@ class FeatureCalculationTestCase(TestCase):
             sum = sum + dat
         self.assertAlmostEqual(sum, 1.0)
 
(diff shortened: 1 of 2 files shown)


nils-braun:
I have basically no idea on the mathematics, so this might be a stupid question: why not set the result to NaN in this case? Because setting it to 1 seems like a bit arbitrary choice, or?

dbarbier:
1 is arbitrary, yes, but since x is null, res_data only contains zeros whichever nonnull value of full_series_energy is chosen; but I agree with you than NaN is a better alternative, will fix it.

PR #591, opened 2019-11-17
Early return from change_quantiles if ql >= qh

dbarbier:
Since 'raise' keyword was missing from this block:

      if ql >= qh:
          ValueError("ql={} should be lower than qh={}".format(ql, qh))

this test did nothing and 0 was returned after some computations.

Modify ComprehensiveFCParameters["change_quantiles"] to only consider parameters with ql < qh.


nils-braun:
Nice catch

PR #592, opened 2019-11-17
Compute quantiles and Friedrich coefficients only once

dbarbier:
`change_quantiles()` is called with several ql and qh arguments. Quantile computation is costly, it is faster to compute all quantiles only once.
For each data chunk, this commit computes all quantiles which are needed, and put them in a `_quantiles` argument, which is used as a cache variable.  If it is set, it is used to speed-up computations.  Otherwise, everything is computed as before. This cache variable is not passed to convert_to_output_format.

Likewise for Friedrich coefficients.

This is a draft pull request, these changes may be controversial but on the other hand provide some nice speed-up.

### DIFF
--- a/tests/units/feature_extraction/test_settings.py
+++ b/tests/units/feature_extraction/test_settings.py
@@ -163,7 +163,7 @@ class TestEfficientFCParameters(TestCase):
         """
         rfs = IndexBasedFCParameters()
         all_feature_calculators = [name for name, func in feature_calculators.__dict__.items()
-                                   if getattr(func, "input", None) == "pd.Series"]
+                                   if getattr(func, "fctype", False) and getattr(func, "input", None) == "pd.Series"]
 
         for calculator in all_feature_calculators:
             self.assertIn(calculator, rfs,
--- a/tsfresh/feature_extraction/extraction.py
+++ b/tsfresh/feature_extraction/extraction.py
@@ -5,6 +5,7 @@
 This module contains the main function to interact with tsfresh: extract features
 """
 
+import copy
 import logging
 import warnings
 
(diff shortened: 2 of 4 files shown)


dbarbier:
<details><summary>Benchmark script (click to display)</summary>

```python
import tsfresh
import pandas as pd
import numpy as np
from time import time

np.random.seed(42)
df = pd.DataFrame(np.concatenate([np.repeat(range(500),40)[:,np.newaxis],
                                  np.tile(range(40),500)[:,np.newaxis],
                                  np.random.randn(20000, 1)], axis=1), columns=["id", "time", "x"])

now = time()
ts = tsfresh.extract_features(df, column_id="id",
    default_fc_parameters=tsfresh.feature_extraction.EfficientFCParameters())
print(time() - now)
```
</details>

On my laptop, `extract_features` runs in 33s with master and 21s with this PR (and 41s with 0.12.0).

dbarbier:
I rebased this branch and above timings are now: 31s with v0.13.0 and 19s with this PR.

This looks great, but I am not sure if this is the best solution, this is why I woulk like to discuss it.  With default parameters, `change_quantiles` method is called 60 times with different parameters on each data chunk.  This method calls `pandas.qcut` with ql and qh parameters, which has to compute quantiles on x and thus sort it.  So the same data are sorted 60 times, and quantiles are computed with different parameters.

This PR proposes to compute all quantiles only once, store them in some cache variable, and use this variable if not None.  This has to be done by `_do_extraction_on_chunk`, before calling `change_quantiles`.  But I am not very fond of the current proposal (hence the draft mode), cache computations should be put in `feature_calculators.py` along with the method they cache.

Another point to note is that those cache computations are done sequentially; if you have more than 60 CPUs, current PR may be slower.  For Friedrich coefficients, the situation is worse, `_estimate_friedrich_coefficients` is called only 5 times with default parameters, if I am not mistaken.  So with more than 6 CPUs, speedup could be negative.  But I still think that this has to be considered, it is always better to perform less computations.

Thoughts?

dbarbier:
Okay, looks better now.

dbarbier:
After more thinking, this PR does not make sense, closing it; caching can already be achieved by changing estimators' fctype to "combiner".

PR #593, opened 2019-11-17
Fix cache in friedrich_coefficients and agg_linear_trend

dbarbier:
There was a logic error in both functions, innermost dictionary did only contain a single element.

Moreover in agg_linear_trend, _aggregate_on_chunks was called even if result is cached.

### DIFF
--- a/tsfresh/feature_extraction/feature_calculators.py
+++ b/tsfresh/feature_extraction/feature_calculators.py
@@ -19,6 +19,7 @@ from __future__ import absolute_import, division
 import itertools
 import warnings
 from builtins import range
+from collections import defaultdict
 
 import numpy as np
 import pandas as pd
@@ -1746,7 +1747,7 @@ def friedrich_coefficients(x, param):
     :return: the different feature values
     :return type: pandas.Series
     """
-    calculated = {}  # calculated is dictionary storing the calculated coefficients {m: {r: friedrich_coefficients}}
+    calculated = defaultdict(dict)  # calculated is dictionary storing the calculated coefficients {m: {r: friedrich_coefficients}}
     res = {}  # res is a dictionary containg the results {"m_10__r_2__coeff_3": 15.43}
 
     for parameter_combination in param:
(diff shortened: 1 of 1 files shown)


tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -1757,11 +1758,8 @@ def friedrich_coefficients(x, param):
         assert coeff >= 0, "Coefficients must be positive or zero. Found {}".format(coeff)
 
         # calculate the current friedrich coefficients if they do not exist yet
-        if m not in calculated:
-            calculated[m] = {r: _estimate_friedrich_coefficients(x, m, r)}
-        else:
-            if r not in calculated[m]:
-                calculated[m] = {r: _estimate_friedrich_coefficients(x, m, r)}
+        if m not in calculated or r not in calculated[m]:
+            calculated[m].update({r: _estimate_friedrich_coefficients(x, m, r)})
```
dbarbier: Instead of dict of dicts, an alternative would be to use a single dict with tuples `(m, r)` as keys, let me know if you prefer this solution.

nils-braun: I am fine with the dict of dicts, but maybe you can write 
`calculated[m][r] = ...`
as I think it is more clear. Maybe you have a different opinion.

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -1840,13 +1838,13 @@ def agg_linear_trend(x, param):
         chunk_len = parameter_combination["chunk_len"]
         f_agg = parameter_combination["f_agg"]
 
-        aggregate_result = _aggregate_on_chunks(x, f_agg, chunk_len)
         if f_agg not in calculated_agg or chunk_len not in calculated_agg[f_agg]:
             if chunk_len >= len(x):
-                calculated_agg[f_agg] = {chunk_len: np.NaN}
+                calculated_agg[f_agg].update({chunk_len: np.NaN})
```
nils-braun: same here

tsfresh/feature_extraction/feature_calculators.py
dbarbier: Agreed, much more readable, thanks.

Issue #528, opened 2019-04-13
tsfresh seems to be incompatible with latest versions of pandas


cchorn:
Oh no, you encountered a problem while using *tsfesh*.

We, the maintainers, are happy to help you. When opening an issue, please provide the following information to us:

1. Your operating system
Ubuntu 18.04.2 LTS

2. The version of *tsfresh* that you are using
latest version
3. The data on which the problem occurred (please do not upload 1000s of time series but try to boil the problem down to a small group or even a singular one)
4. A minimal code snippet which reproduces the problem/bug
5. Any reported errors or traceback

error when trying to install:

tsfresh 0.11.2 has requirement pandas<=0.23.4,>=0.20.3, but you'll have pandas 0.24.2 which is incompatible.

For questions, you can also use our [gitter chatroom](https://gitter.im/tsfresh/)

AStupidBear:
Is there any specific reason for tsfresh not to use the latest stable pandas?

jnj16180340:
```
pandas>=0.20.3,<=0.23.4 # pandas dropna is buggy in 0.24.0, see https://github.com/blue-yonder/tsfresh/issues/485 and https://github.com/pandas-dev/pandas/issues/25087
```

Looks like it's fixed in `0.25.0` https://github.com/pandas-dev/pandas/issues/25087#issuecomment-506855641

lobpcg:
```
ERROR: tsfresh 0.12.0 has requirement pandas<=0.23.4,>=0.20.3, but you'll have pandas 0.25.0 which is incompatible.
```

jpzhangvincent:
Same question! Can we merge this pull request? https://github.com/blue-yonder/tsfresh/pull/562

dbarbier:
Hi, new conda packages are available, this issue can be closed.

PR #594, opened 2019-11-18
Feature/528 allow newest pandas version

nils-braun:
This builds on top of #562 and fixes the two unit tests.

It also does not require the newest version of pandas, but only to not be 0.24.*

PS: we should really fix the .travis.yaml, I have the strong feeling we are not doing this very optimally... I especially do not like the specification of the versions. Will tackle this later.

@dbarbier Are you fine with this?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ jobs:
       env: NUMPY="latest", PANDAS="latest", SCIKITLEARN="latest", DASK="latest", DISTRIBUTED="latest", SCIPY="latest"
       before_script:
         - pip install --upgrade numpy
-        - pip install --upgrade pandas
+        - pip install pandas===0.23.4 # pandas installation fails with newer version
         - pip install scikit-learn==0.20.3 # last scikit-learn version to support python 2.7
         - pip install --upgrade dask
         - pip install --upgrade distributed
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,7 +59,7 @@ jobs:
       after_success:
         - coveralls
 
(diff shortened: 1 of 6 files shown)


nils-braun:
This should fix #528

dbarbier:
> @dbarbier Are you fine with this?

@nils-braun Sure, thanks a lot!

dbarbier:
@nils-braun Failure is caused by a change in pandas 0.25.3, pd.melt now calls np.ravel, and quite surprisingly `np.ravel(['index', 1])` returns `array(['index', '1'], dtype='<U5')`, and pd.melt complains that there is no column named `str(1)`.  One might argue that this is a pandas bug, but on the other hand I am not sure whether mixing character and numeric column names is well supported, thus I suggest to use a numeric name for index if columns have numeric names.

nils-braun:
Yes, thanks for pointing out. I did see it yesterday evening but could not fix it before going to bed.

I will create an issue on the pandas repo nevertheless, but I guess you are right: we should fix it also from our side.

nils-braun:
(hopefully this will all go away once https://github.com/pandas-dev/pandas/pull/28859 is finished)

Froskekongen:
Would it be possible to do a new release once this PR is merged?

dbarbier:
Hi @nils-braun do you have any idea why travis builds are failing?

dbarbier:
Got it, pandas 0.25 requires python >= 3.5.3 https://pandas.pydata.org/pandas-docs/stable/install.html#python-version-support
I can confirm that everything works fine with python 3.5.3.

Rictus:
Thank you for this work!

PR #602, opened 2019-11-24
Refactor .travis.yml

### LINKED ISSUE #595: Optimize travis build
I think our .travis file has grown a bit historically :-)

Once python 2 support is dropped I think it can be cleaned up a bit. Also I think there should be a better way to test different dependency versions.

dbarbier:
Only build coverage with Python 3.7 and latest package versions,
this allows to run test stages in parallel.

Add python 3.8 but allow it to fail, some dependencies are not
yet ready.

This is a proposal in order to fix #595, obviously I did not check neither coverage nor deployment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,7 @@ jobs:
       script: *test
 
     - stage: Test coverage
-      if: env(PYTEST_ADDOPTS) = "-c setup.cfg" AND fork = false
+      python: 3.7
       script: coveralls
 
     - stage: Deploy
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,14 +10,15 @@ install:
   - pip install -r requirements.txt -r test-requirements.txt
   - pip install -U .
   - pip freeze
-  - "[ $NUMPY = latest ] && pip install --upgrade numpy || pip install numpy==$NUMPY"
-  - "[ $PANDAS = latest ] && pip install --upgrade pandas || pip install pandas==$PANDAS"
-  - "[ $SCIKITLEARN = latest ] && pip install --upgrade scikit-learn || pip install scikit-learn==$SCIKITLEARN"
-  - "[ $DASK = latest ] && pip install --upgrade dask || pip install dask==$DASK"
-  - "[ $DISTRIBUTED = latest ] && pip install --upgrade distributed || pip install distributed==$DISTRIBUTED"
+  # install is run during all stages; those packages are needed only to run tests
+  - "[ \"$NUMPY\" = latest ] && pip install --upgrade numpy || [ -z \"$NUMPY\" ] || pip install numpy==$NUMPY"
+  - "[ \"$PANDAS\" = latest ] && pip install --upgrade pandas || [ -z \"$PANDAS\" ] || pip install pandas==$PANDAS"
+  - "[ \"$SCIKITLEARN\" = latest ] && pip install --upgrade scikit-learn || [ -z \"$SCIKITLEARN\" ] || pip install scikit-learn==$SCIKITLEARN"
+  - "[ \"$DASK\" = latest ] && pip install --upgrade dask || [ -z \"$DASK\" ] || pip install dask==$DASK"
+  - "[ \"$DISTRIBUTED\" = latest ] && pip install --upgrade distributed || [ -z \"$DISTRIBUTED\" ] || pip install distributed==$DISTRIBUTED"
   # need to downgrade tornado manually
-  - "[ $SCIPY = latest ] || pip install tornado==4.5.3"
-  - "[ $SCIPY = latest ] && pip install --upgrade scipy || pip install scipy==$SCIPY"
+  - "[ \"$SCIPY\" = latest ] || [ -z \"$SCIPY\" ] || pip install tornado==4.5.3"
+  - "[ \"$SCIPY\" = latest ] && pip install --upgrade scipy || [ -z \"$SCIPY\" ] || pip install scipy==$SCIPY"
   - pip list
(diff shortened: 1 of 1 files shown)


nils-braun:
Thanks for looking into this @dbarbier !
I should have told you I have a un-pushed branch with some changes already :-D But you were faster, so thank you very much! 
I have some comments though ;-)

.travis.yml
```diff
@@ -9,107 +10,59 @@ install:
   - pip install -r requirements.txt -r test-requirements.txt
   - pip install -U .
   - pip freeze
+before_script:
```
nils-braun: any reason to note move this to the installation step?

.travis.yml
```diff
@@ -9,107 +10,59 @@ install:
...
+  - "[ $DISTRIBUTED = latest ] && pip install --upgrade distributed || pip install distributed==$DISTRIBUTED"
+  # need to downgrade tornado manually
+  - "[ $SCIPY = latest ] || pip install tornado==4.5.3"
+  - "[ $SCIPY = latest ] && pip install --upgrade scipy || pip install scipy==$SCIPY"
+  - pip list
+  - sed -i -e 's/-n auto/-n 2/g' setup.cfg
+  - sed -e '/^\s*--cov tsfresh/d' setup.cfg > setup-nocov.cfg
+env:
+  global:
+    - PYTEST_ADDOPTS="-c setup-nocov.cfg"
```
nils-braun: Maybe you want to add a comment why you do this here (and to the line above)

.travis.yml
```diff
@@ -9,107 +10,59 @@ install:
...
-        - pip install tornado==4.5.3
-        - pip install scipy==1.2.0
-        - pip list
-      script:
-        - sed -i 's/\-n auto/\-n 2/g' setup.cfg
-        - "if [ $TRAVIS_PULL_REQUEST == false ] && ! [ $TRAVIS_BRANCH == 'master' ]; then pytest tests/units -n2; else pytest tests; fi"
-      python: 3.6
+    - stage: Test coverage
+      if: env(PYTEST_ADDOPTS) = "-c setup.cfg" AND fork = false
+      script: coveralls
```
nils-braun: Maybe I am just too stupid, but why is it not possible to do a "normal" python version matrix (as e.g. https://docs.travis-ci.com/user/languages/python/) with the additional `env` variables, just define the test script once (as the script) and have the coveralls as an `after_success` as described here: 
https://github.com/z4r/python-coveralls?
Also, why did you disable it on forks? Sorry for the maybe naive questions

.travis.yml
```diff
@@ -9,107 +10,59 @@ install:
...
-        - pip install dask==0.16.1
-        - pip install distributed==1.18.3
-        # need to downgrade tornado manually
-        - pip install tornado==4.5.3
-        - pip install scipy==1.2.0
-        - pip list
-      script:
-        - sed -i 's/\-n auto/\-n 2/g' setup.cfg
-        - "if [ $TRAVIS_PULL_REQUEST == false ] && ! [ $TRAVIS_BRANCH == 'master' ]; then pytest tests/units -n2; else pytest tests; fi"
+    - stage: Deploy
```
nils-braun: Again there is probably a good reason, by why not just have this as the deploy option (not as a stage)?

dbarbier:
> Thanks for looking into this @dbarbier !
> I should have told you I have a un-pushed branch with some changes already :-D But you were faster, so thank you very much!

No problem, this is mostly a proposal, feel free to close this PR if you have another solution.

.travis.yml
dbarbier: No, I tried to make minimal changes, and kept this part.  I will move it to install.

.travis.yml
dbarbier: Ok

.travis.yml
dbarbier: The problem is that environment variables depend on Python version; my feeling is that this script would be less readable with a matrix, hence I did not try, but I may be wrong.

About forks, I did not want to make noise when making experiments in my repository, I am not sure whether this is a good idea of not.

.travis.yml
dbarbier: Okay, I am not a travis expert, so please take explanations below with a grain of salt.

If you look at current travis, there are 6 different stages.  As soon as a stage fail, others are cancelled, as can be seen at https://travis-ci.org/blue-yonder/tsfresh/builds/616016632 or https://travis-ci.org/blue-yonder/tsfresh/builds/614728433

When something goes wrong, it is important to gather as much information about all builds as possible, so this behavior is IMO very annoying.

I tried to understand why you did it that way, and thought that you wanted to avoid job concurrency.  Also I do not fully understand how coverage works; for instance if you look at https://travis-ci.org/blue-yonder/tsfresh/builds/616693314/ config says that coveralls is called by the build with python 3.7 and latest dependencies.  But if you look at all logs, you see at the end no-data-collected warning in all builds but 2343.4, which is for python 3.5.3 with oldest dependencies.  I was really puzzled, and tried something different which seems to work as expected ;)
So unless you have an explanation for current logs, I prefer this solution with stages.

.travis.yml
dbarbier: Maybe a simpler solution would be to run all tests without coverage during the build phase, and to run a single job with coverage afterwards.

.travis.yml
dbarbier: I forgot to mention that by moving deployment to a different stage, this ensures that all builds succeeded, and not only python 3.7 with latest dependencies.

nils-braun: :+1:

dbarbier:
Superseded by #605

PR #605, opened 2019-11-26
Refactor .travis.yml part II

nils-braun:
This PR builds on top of #602  with just a bit of extra refactoring. All the hard work was done by @dbarbier! 

I though it is easier to actually to the changes instead of describing it one by one :-) This is how I imagined it with the script just defined once (I also added some comments).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,7 @@ jobs:
       script: *test
 
     - stage: Test coverage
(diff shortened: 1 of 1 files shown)


nils-braun:
@dbarbier what do you think?

dbarbier:
Looks good to me.  The only untested bit is deployment, but I have no idea on how to test it.

nils-braun:
Well, I will test it on the next release ;-)

PR #606, opened 2019-11-26
Add github actions

nils-braun:
This PR is the base for testing github actions.

As a first test, I added a stylecheck action to see its impact.


nils-braun:
It seems I first need to fix some more pep-8 issues :-D

PR #607, opened 2019-11-26
Feature/pep8

nils-braun:
I know. This is the PR basically everyone hates.

But, for a modern python package we should also comply to pep8, so I added a github action to check exactly this + some settings (line length = 120).

I also pep8-ified everything, which means we have a large PR but *no changes except whitespace and alike*.
Tests will hopefully show that :-)


dbarbier:
Another option is to use https://github.com/psf/black, it makes life easier IMHO.

nils-braun:
I guess in the end it boils down to a matter of taste and I (personally) do not like the black decisions that much :man_shrugging: 
But of course, if people like it - I have no strong objections :-)

nils-braun:
My suggestion: lets start with that (which is "easier" to digest) and see how people like it.
And then move to black maybe later.

dbarbier:
Feel free to merge it when it is ready, I will take care of rebasing my branches.

MaxBenChrist:
I used flake8 and pylint in the past. Pylint is super aggressive. Flake8 was nice. We could add it to the travis build?

nils-braun:
This PR adds a github action (which is the CI tool of github) to do a pycodestyle check (which is exactly pep8 with some additional 120 column size setting).

nils-braun:
@MaxBenChrist Any more thoughts or can I merge?
I want to merge this as quickly as possible because it affects so many other PRs...

PR #609, opened 2019-11-26
Improve test coverage

dbarbier:


tsfresh/utilities/distribution.py
```diff
@@ -319,7 +319,9 @@ def distribute(self, func, partitioned_chunks, kwargs):
         :return: The result of the calculation as a list - each item should be the result of the application of func
             to a single element.
         """
-
+        if isinstance(partitioned_chunks, Iterable):
+            # since dask 2.0.0 client map no longer accepts iterables
+            partitioned_chunks = list(partitioned_chunks)
```
MaxBenChrist: Doesn't this casting (`partitioned_chunks = list(partitioned_chunks)`) creates a risk to blow up the memory?

Let us say we have data that is 6 GB big. Now, partitioned_chunks gets casted as list, so we have another 6 GB, 12 used in total. Now, we push that data into the workers and suddenly we use 18 GB of memory in total.

On the other hand, the `client.map` will break on the iterable, right?

dbarbier: True, but this is the exact same fix you applied at #550, and I like consistency ;)

dbarbier: In fact I was wrong, client.map internally converts iterables into a list, so there is no change.
What you want is a queue of fixed size, but it is harder to implement.

nils-braun:
@MaxBenChrist lgtm

PR #612, opened 2019-11-29
Add "autolag" parameter to augmented_dickey_fuller()

fujimotos:
The result of Augumented Dickey-Fuller test deeply depends on how we
choose the lag parameter. StatsModels' adfuller() function provides
several criteria for that selection (AIC, BIC and t-stats etc).

This makes it a parameter, and enables users to tweak the lag selection
according to their preference.

Signed-off-by: Fujimoto Seiji <fujimoto@clear-code.com>

### DIFF
--- a/tests/units/feature_extraction/test_feature_calculations.py
+++ b/tests/units/feature_extraction/test_feature_calculations.py
@@ -301,6 +301,11 @@ class FeatureCalculationTestCase(TestCase):
         for index, val in res_value_error:
             self.assertIsNaN(val)
 
+        # Should return NaN if "attr" is unknown
+        res_attr_error = augmented_dickey_fuller(x=x, param=[{"autolag": "AIC", "attr": ""}])
+        for index, val in res_attr_error:
+            self.assertIsNaN(val)
+
     def test_abs_energy(self):
         self.assertEqualOnAllArrayTypes(abs_energy, [1, 1, 1], 3)
         self.assertEqualOnAllArrayTypes(abs_energy, [1, 2, 3], 14)
--- a/tests/units/feature_extraction/test_feature_calculations.py
+++ b/tests/units/feature_extraction/test_feature_calculations.py
@@ -251,14 +251,22 @@ class FeatureCalculationTestCase(TestCase):
         # H0 is true
         np.random.seed(seed=42)
         x = np.cumsum(np.random.uniform(size=100))
-        param = [{"attr": "teststat"}, {"attr": "pvalue"}, {"attr": "usedlag"}]
-        expected_index = ['attr_"teststat"', 'attr_"pvalue"', 'attr_"usedlag"']
+        param = [
+            {"autolag": "BIC", "attr": "teststat"},
+            {"autolag": "BIC", "attr": "pvalue"},
+            {"autolag": "BIC", "attr": "usedlag"}
+        ]
+        expected_index = [
+            'autolag_"BIC"__attr_"teststat"',
+            'autolag_"BIC"__attr_"pvalue"',
+            'autolag_"BIC"__attr_"usedlag"',
+        ]
 
         res = augmented_dickey_fuller(x=x, param=param)
         res = pd.Series(dict(res))
         self.assertCountEqual(list(res.index), expected_index)
-        self.assertGreater(res['attr_"pvalue"'], 0.10)
-        self.assertEqual(res['attr_"usedlag"'], 0)
+        self.assertGreater(res['autolag_"BIC"__attr_"pvalue"'], 0.10)
+        self.assertEqual(res['autolag_"BIC"__attr_"usedlag"'], 0)
 
         # H0 should be rejected for AR(1) model with x_{t} = 1/2 x_{t-1} + e_{t}
         np.random.seed(seed=42)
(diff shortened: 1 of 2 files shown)


fujimotos:
For what's it worth, this parameter is also very useful for making the
computation of `augmented_dickey_fuller()` less resource-intensive.

In particular, the current implementation's memory usage grows very rapidly
as the input data size grows (In fact, we're seeing some OOM errors while
experimenting with a real-life data set).

We can confirm that the resource usage shrinks dramatically by setting
the `autolag` parameter to "None", which makes adfuller() use a fixed lag
parameter.

Here is a graph from our experiment. The Y axis is the memory usage and 
X axis is the input data size (number of entries). The red line is the current
implementation (which uses AIC) and the blue line is when we set `None`
to the "autolag" parameter.

augmented-dickey-fuller

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -427,27 +427,42 @@ def augmented_dickey_fuller(x, param):
...
-                  res[0] if config["attr"] == "teststat"
-             else res[1] if config["attr"] == "pvalue"
-             else res[2] if config["attr"] == "usedlag" else np.NaN)
-            for config in param]
 
+    def compute_adf(autolag):
+        try:
+            return adfuller(x, autolag=autolag)
+        except LinAlgError:
+            return (np.NaN, np.NaN, np.NaN)
```
nils-braun: Is there a reason why you introduced brackets here?

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -427,27 +427,42 @@ def augmented_dickey_fuller(x, param):
 
     :param x: the time series to calculate the feature of
     :type x: numpy.ndarray
-    :param param: contains dictionaries {"attr": x} with x str, either "teststat", "pvalue" or "usedlag"
+    :param param: contains dictionaries {"attr": x, "autolag": y} with x str, either "teststat", "pvalue" or "usedlag"
+                  and with y str, either of "AIC", "BIC", "t-stats" or None.
```
nils-braun: Why can it be "none"? Why would I want to have this?
If this is really a usual case, you might want to document it here :-)

tests/units/feature_extraction/test_feature_calculations.py
```diff
@@ -284,6 +300,11 @@ def test_augmented_dickey_fuller(self):
         for index, val in res_value_error:
             self.assertIsNaN(val)
 
+        # Should return NaN if "attr" is unknown
+        res_attr_error = augmented_dickey_fuller(x=x, param=[{"autolag": "AIC", "attr": ""}])
+        for index, val in res_attr_error:
```
nils-braun: Wouldn't it be better to throw some warning?'if we do not know the setting?

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -427,27 +427,42 @@ def augmented_dickey_fuller(x, param):
...
+            return (np.NaN, np.NaN, np.NaN)
+        except MissingDataError: # is thrown for e.g. inf or nan in the data
+            return (np.NaN, np.NaN, np.NaN)
+
+    cache = {}
+    res = []
+    for config in param:
+        autolag = config.get("autolag", "AIC")
+        if autolag not in cache:
+            cache[autolag] = compute_adf(autolag)
```
nils-braun: You could move the last three lines also into the helper function and just call it here - would look a bit cleaner in my opinion

tsfresh/feature_extraction/feature_calculators.py
fujimotos: No particular reason. I removed these brackets in e77b909
so that it looks consistent with other codes.

tsfresh/feature_extraction/feature_calculators.py
fujimotos: I added a reference to StatsModels' documentation there in e77b909.

> Why can it be "none"? Why would I want to have this?

This is corresponding to StatsModels' `adfuller()` function. 

https://www.statsmodels.org/stable/generated/statsmodels.tsa.stattools.adfuller.html

This function accepts `None` as the `autolag` parameter, which means we use the
constant `12*(nobs/100)^{1/4}` as the maximum lag length. This is a sensible choice
of the parameter and especially useful when we want to avoid CPU/memory
intensive fitting process.

tests/units/feature_extraction/test_feature_calculations.py
fujimotos: This just preserves the old behaviour. If it is preferable to raise
errors (instead of return NaN) in such case, I'll fix it so.

tsfresh/feature_extraction/feature_calculators.py
fujimotos: > You could move the last three lines also into the helper function and just call it here - would look a bit cleaner in my opinion

OK. I fixed the helper function to cache the result automatically.

fujimotos:
@nils-braun Thanks for your review! I posted a update e77b909
that should fix the issues you pointed out.

tests/units/feature_extraction/test_feature_calculations.py
nils-braun: Right, correct!

nils-braun:
Thanks, looks good now!

nils-braun:
@fujimotos -san: So far you only changed the code of the calculator, not the settings. This means by default only the AIC is calculated (as it is now). Do you think it would be worth it to add the others (or only some of them) to the default settings? Because otherwise I fear no one will use this feature...

fujimotos:
@nils-braun 

> Do you think it would be worth it to add the others (or only some of them) to the default settings?

I don't think it's worth doing.

The selection of AIC, BIC or t-stats is for those who are really into the Dickey
Fuller test. For average users, it's hard to argue that providing results for
BIC and t-stats is something worth the additional computing time.

> Because otherwise I fear no one will use this feature...

If I do, I will tweak the parameter in `EfficientFCParameters`.

A great thing about this patch is that we can cut the computation cost
of `augmented_dickey_fuller()` by 80-90% by setting `autolag` to None.

This should be a noticable improvement, since `augumented_dickery_fuller()`
ranks second in term of the computation cost (see the graph below).

Runtime among EfficientFCParameters

fujimotos:
Rebased to the master HEAD and resolve conflicts.

nils-braun:
Ok, very nice. I will merge this now.
Maybe we can discuss together with @MaxBenChrist if we should set the default to "None" because it is so much faster.

fujimotos:
Thank you!

PR #624, opened 2020-02-02
Upgrading numpy version

nils-braun:
There seems to be API incompatibilities

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 requests>=2.9.1
-numpy>=1.12.0
+numpy>=1.13.1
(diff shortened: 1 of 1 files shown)


dbarbier:
@nils-braun Build failures seem to be caused by stasmodels, not numpy. You should handle it like other packages in `.travis.yml`, I guess that this will fix build failures with older packages.  Morever in statsmodels 0.11.0, they changed `AR` class, and `calculated_AR` variable in `feature_calculators.py` can no longer be used for all `k` values, you need to instantiate one `AR` class for each `k`.

PR #625, opened 2020-02-03
Add a versioned dependency on statsmodels for travis CI

dbarbier:
New statsmodels 0.11.0 package seems to cause compatibility issues with older packages:

    RuntimeError: module compiled against API version 0xc but this version of numpy is 0xa

We depend on statsmodels 0.9.0 because this is the first version available on conda
for Python 3.5, 3.6 and 3.7, but users can use 0.8.0 as well.

Fix `ar_coefficient method` with newest statsmodels, `AR` class is now immutable.

Also latest dask requires fsspec in order to read data from disk.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,8 +28,6 @@ install:
   # need to downgrade tornado manually
   - "[ \"$SCIPY\" = latest ] || [ -z \"$SCIPY\" ] || pip install tornado==4.5.3"
   - "[ \"$SCIPY\" = latest ] && pip install --upgrade scipy || [ -z \"$SCIPY\" ] || pip install scipy==$SCIPY"
-  # recent dask requires fsspec to load data from disk
-  - "[ \"$DASK\" = latest ] && pip install fsspec || true"
   # Print out the pip versions for debugging
   - pip list
   # Only use two cores
@@ -94,4 +92,3 @@ jobs:
     - python: 3.8
   # Make sure to not wait for Python 3.8
   fast_finish: true
-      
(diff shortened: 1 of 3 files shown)


dbarbier:
Supersedes #624

nils-braun:
As always, thanks for fixing to fast! I have seen that the numpy version was not the problem but I did not have time to look into it. Kudos!

.travis.yml
```diff
@@ -22,11 +22,14 @@ install:
...
   - "[ \"$PANDAS\" = latest ] && pip install --upgrade pandas || [ -z \"$PANDAS\" ] || pip install pandas==$PANDAS"
   - "[ \"$SCIKITLEARN\" = latest ] && pip install --upgrade scikit-learn || [ -z \"$SCIKITLEARN\" ] || pip install scikit-learn==$SCIKITLEARN"
+  - "[ \"$STATSMODELS\" = latest ] && pip install --upgrade statsmodels || [ -z \"$STATSMODELS\" ] || pip install statsmodels==$STATSMODELS"
   - "[ \"$DASK\" = latest ] && pip install --upgrade dask || [ -z \"$DASK\" ] || pip install dask==$DASK"
   - "[ \"$DISTRIBUTED\" = latest ] && pip install --upgrade distributed || [ -z \"$DISTRIBUTED\" ] || pip install distributed==$DISTRIBUTED"
   # need to downgrade tornado manually
   - "[ \"$SCIPY\" = latest ] || [ -z \"$SCIPY\" ] || pip install tornado==4.5.3"
   - "[ \"$SCIPY\" = latest ] && pip install --upgrade scipy || [ -z \"$SCIPY\" ] || pip install scipy==$SCIPY"
+  # recent dask requires fsspec to load data from disk
+  - "[ \"$DASK\" = latest ] && pip install fsspec || true"
```
nils-braun: I would actually prefer to have this in the test-requirements.txt, don't you think?

dbarbier: Yes, this will work since fsspec is provided as a conda package and its only requirement is Python >= 3.5.  But I am not sure this is a good idea, you do not need fsspec if you have an older dask.
Anyway I have no strong feeling, feel free to change it if you prefer.

nils-braun: Thinking about it, we actually rely on dask[dataframe] instead of dask, and this will also include ffspec automatically. I like this solution more - because currently people installing our package might have problems with that also (not only in the tests).

PR #626, opened 2020-02-04
Fix the kernel and the naming of a notebook

nils-braun:


notebooks/feature_extraction_with_datetime_index.ipynb
```diff
@@ -571,9 +571,9 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "umcu_env",
```
MaxBenChrist: do you know what umcu_env is?

PR #630, opened 2020-02-10
added security parameter for dask

jontis:
from distributed.security import Security
sec = Security(tls_ca_file='./dask-ca.pem', tls_client_cert='./dask-cert.pem', tls_client_key='./dask-key.pem', require_encryption=True)
Distributor = ClusterDaskDistributor(address="tls://ip_adress_or_url:8786", security=sec)

### DIFF
--- a/tsfresh/utilities/distribution.py
+++ b/tsfresh/utilities/distribution.py
@@ -296,7 +296,7 @@ class ClusterDaskDistributor(DistributorBaseClass):
     Distributor using a dask cluster, meaning that the calculation is spread over a cluster
     """
 
-    def __init__(self, address):
+    def __init__(self, address, security=None):
         """
         Sets up a distributor that connects to a Dask Scheduler to distribute the calculaton of the features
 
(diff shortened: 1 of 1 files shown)


MaxBenChrist:
I think it would make more sense to expose all [distributed.Client](https://distributed.dask.org/en/latest/api.html?highlight=client#distributed.Client) arguments as kwargs.

nils-braun:
@jontis Do you want to add the `kwargs`?

jontis:
> @jontis Do you want to add the `kwargs`?

Not sure of the best way to do that in the most transparent way. 
I'm leaning towards for clarity
def __init__(self, address=None, **kwargs)

from distributed import Client
self.client = Client(address=address, **kwargs)

But it could as well be
def __init__(self, **kwargs)

from distributed import Client
self.client = Client(**kwargs)

Opinions?
@MaxBenChrist

nils-braun:
Yes, I would also keep the address argument separate

PR #632, opened 2020-02-12
Add count_above and count_below features

danielazevedo:
I created two simple features, which I needed for my project. These features return the percentage of values in a time series that are above/below a certain threshold.

### DIFF
--- a/tests/units/feature_extraction/test_feature_calculations.py
+++ b/tests/units/feature_extraction/test_feature_calculations.py
@@ -1160,6 +1160,29 @@ class FeatureCalculationTestCase(TestCase):
         res = change_quantiles(np.random.rand(10000) * 1000, 0.1, 0.2, False, 'mean')
         self.assertAlmostEqual(res, -0.9443846621365727)
 
+    def test_count_above(self):
+        self.assertEqualPandasSeriesWrapper(count_above, [1] * 10, 1, t=1)
+        self.assertEqualPandasSeriesWrapper(count_above, list(range(10)), 1, t=0)
+        self.assertEqualPandasSeriesWrapper(count_above, list(range(10)), 0.5, t=5)
+        self.assertEqualPandasSeriesWrapper(count_above, [0.1, 0.2, 0.3] * 3, 2/3, t=0.2)
+        self.assertEqualPandasSeriesWrapper(count_above, [np.NaN, 0, 1] * 3, 2/3, t=0)
+        self.assertEqualPandasSeriesWrapper(count_above, [np.NINF, 0, 1] * 3, 2/3, t=0)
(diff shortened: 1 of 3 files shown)


MaxBenChrist:
Hi @danielazevedo, welcome to tsfresh.

Can you add some unit tests for those features?

tsfresh/feature_extraction/settings.py
```diff
@@ -146,7 +146,9 @@ def __init__(self):
             "energy_ratio_by_chunks": [{"num_segments": 10, "segment_focus": i} for i in range(10)],
             "ratio_beyond_r_sigma": [{"r": x} for x in [0.5, 1, 1.5, 2, 2.5, 3, 5, 6, 7, 10]],
             "linear_trend_timewise": [{"attr": "pvalue"}, {"attr": "rvalue"}, {"attr": "intercept"},
-                                      {"attr": "slope"}, {"attr": "stderr"}]
+                                      {"attr": "slope"}, {"attr": "stderr"}],
+            "count_above": [{"t": t} for t in [-1, 0, 1]],
```
MaxBenChrist: I don't think we should give defaults for this feature, as the usefullness highly depends on the value range for your time series.

danielazevedo: Indeed, so you are suggesting I should remove them from the ComprehensiveFCParameters?

danielazevedo:
Sure, I've added some unit test for the count_above and count_below features

tsfresh/feature_extraction/settings.py
MaxBenChrist: Yes, I would remove it

tests/units/feature_extraction/test_feature_calculations.py
```diff
@@ -1160,6 +1160,30 @@ def test_change_quantiles(self):
         res = change_quantiles(np.random.rand(10000) * 1000, 0.1, 0.2, False, 'mean')
         self.assertAlmostEqual(res, -0.9443846621365727)
 
+    def test_count_above(self):
```
MaxBenChrist: looks good!

tsfresh/feature_extraction/settings.py
danielazevedo: By removing the features from the ComprehensiveFCParameters, it throws an error during the tests execution. The _test_default_calculates_all_features_ test fails, as it checks if, by default, the ComprehensiveFCParameters object contains all the existing features.

nils-braun: If any, then I would go only with the 0 as default parameter

danielazevedo: I added the 0 as the default parameter

nils-braun:
There are some very strange things going on with our tests... In principle the test runs through, so no problem here

Issue #635, opened 2020-02-26
Parameters not sorted alphabetically in feature name during feature extraction


trueckel:
First of all, great work with tsfresh. Really nicely done!

Now to the issue:
OS: Windows 10 Pro x64, but doesn't matter here.
tsfresh ver: 0.13.0

The function 'extract_features' returns a dataframe in which the column names represent the feature names, also containing the parameters used to calculate that feature.
The order of the parameter names within the feature name however are not sorted alphabetically.
This surprised me, because you implemented a utility function 'convert_to_output_format' in the string_manipulation module which converts parameter dictionaries to a string, in which the parameters are in fact sorted alphabetically.

As it is:
Signal1__fft_coefficient__coeff_77__attr_"angle"

As it should be:
Signal1__fft_coefficient__attr_"angle"__coeff_77

In my use case same features happen to have different names because the parameters in the string are sorted differently.

nils-braun:
Yes, good point. That is definitely something that should be fixed. 
Do you want to do a pull request or should we tackle the issue? We are always very happy for every contributor!

trueckel:
I would like to do a pull request and fix it, my first pull request though.
So started with a fork and later do the pull request right?

nils-braun:
Yes! Welcome :-)

PR #641, opened 2020-03-10
Fix parameter sorting

trueckel:
This shall fix issue #635


nils-braun:
Thank you very much for the PR, but our tests also rely on the extracted parameter names. Would you mind also fixing them?

nils-braun:
@trueckel Should I fix the tests or do you need help in any way for this?

trueckel:
@nils-braun I am a bit clueless regarding the tests, so I would prefer if you would do it.

nils-braun:
Superseeded by #656 with added tests.

PR #656, opened 2020-04-05
Fix parameter sorting

nils-braun:
Superseeds #641 and adds tests.
Thanks to @trueckel for the implementation.

### DIFF
--- a/tsfresh/feature_extraction/feature_calculators.py
+++ b/tsfresh/feature_extraction/feature_calculators.py
@@ -12,6 +12,9 @@ There are two types of features:
 They are specified using the "fctype" parameter of each feature calculator, which is added using the
 set_property function. Only functions in this python module, which have a parameter called  "fctype" are
 seen by tsfresh as a feature calculator. Others will not be calculated.
+
+Feature calculators of type combiner should return the concatenated parameters sorted
+alphabetically ascending.
 """
 
 import itertools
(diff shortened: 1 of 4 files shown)


nils-braun:
@MaxBenChrist what do you think? The feature @trueckel implemented is actually very reasonable, but it breaks our API (and maybe the code of many people).
Need your OK here :-)

MaxBenChrist:
> @MaxBenChrist what do you think? The feature @trueckel implemented is actually very reasonable, but it breaks our API (and maybe the code of many people).
> Need your OK here :-)

I am not seeing why this breaks our API?

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -12,6 +12,9 @@
 They are specified using the "fctype" parameter of each feature calculator, which is added using the
 set_property function. Only functions in this python module, which have a parameter called  "fctype" are
 seen by tsfresh as a feature calculator. Others will not be calculated.
+
+Feature calculators of type combiner should return the concatenated parameters sorted
```
MaxBenChrist, line 16: This is garantueed by https://github.com/blue-yonder/tsfresh/blame/54de174f593290a0c798ccad2ad7cad493c9628e/tsfresh/utilities/string_manipulation.py#L70, isn't?

nils-braun: Well unfortunately no. Only if it is a non-combiner...

nils-braun:
@MaxBenChrist I will merge it nevertheless. I does break our API if people really use the extracted feature names in the code. We would need to make sure to communicate this clearly in the net version I guess?

PR #666, opened 2020-04-13
Add add_sub_time_series_index

### LINKED ISSUE #633: Generate sub-series from timeseries, to extract all possible features
Using tsfresh ( for example the roll function): 
Is it possible to generate features as a matrix X, containing in each row, all the possible features of every sub-series of the time-series data ? 

image

nils-braun:
This fixes #633.


nils-braun:
@MaxBenChrist This adds another function to the utilities, to chunk the data into equally-sized packages.
I ask for your review just to check if you are still alive :-D

Issue #659, opened 2020-04-08
Roll_time_series


OJL96:
Hello,

thank for for the tsfresh package, it has been most useful.

So im trying to extract features from my dataset, before doing so i would like to using a rolling method to increase my data entries by getting the mean of  (im guessing thats the purpose of it).

so my dataset looks like this:
TIME            VALUE  CLASS  uniqueID
 0.150050 | 0.368338   |   0    |     0
 0.150100 | 0.364448   |   0    |     1
 0.150150 | 0.360232   |   0    |     2
 0.150200 |0.355811    |   0    |     3
 0.150250 |0.351306    |   0    |     4
         ...       ...    ...       ...
 0.299799 | 1.917936   |   6  |  539455
 0.299849 | 1.916304   |   6  |  539456
 0.299900 | 1.914653   |   6  |  539457
 0.299950 | 1.913025   |   6  |  539458
 0.300000 | 1.911464   |   6  |  539459

and after executing:
data["uniqueID"] = len(data)
y = data.CLASS
rolledData = roll_time_series(data, 
                              column_id="CLASS",
                              column_sort="TIME",
                              column_kind=None,
                              rolling_direction=100)

i get the following user warning: UserWarning: Your time stamps are not uniformly sampled, which makes rolling nonsensical in some domains. 

it never finishes the function and displays the new data 

any help is much appreciated.

nils-braun:
Hi, thanks for the issue.

So maybe to start from the beginning: I think I did not really understand your data.
* why do you want to roll? Rolling produces a prediction dataset out of a normal time series. You can find some information here: https://tsfresh.readthedocs.io/en/latest/text/forecasting.html From your "y = data.CLASS" I understand however, you have a classification problem? Or what is the purpose you want to achieve? 
* what do you mean by "getting the mean"?
* you are using `data.CLASS` both as `y` as well as an the `column_id`. Are you sure that makes sense? Check out the nomenclature here: https://tsfresh.readthedocs.io/en/latest/text/data_formats.html
* what do you mean it never ends? Is it just too slow? The warning is not a reason why it should not end (but it is a warning, which is for your information). Check out about warnings: https://docs.python.org/3/library/warnings.html which are normally not a reason why it should not finish. Could you try with less data? You have quite a large data sample.

nils-braun:
I guess #668 and the additional parameter for the number of jobs will help you. If not, please reopen the issue!

Issue #646, opened 2020-03-24
Add multiprocessing capability to rolling


demontamer:
Oh no, you encountered a problem while using *tsfesh*.
I notice the above functions do not take njobs or distributor as args. They are quite slow for even moderately large datasets of size 5000?

nils-braun:
Yes, `roll_time_series` is not implemented in a multiprocessing ready way. Probably it is possible, but we have never thought about it. If you have any good idea on how to do it, we are happy to hear it (or even a pull request :-))
The `extract_relevant_features` function has a parameter for `njobs`, but not for the distributor, that is correct. I will fix this. Until that, you can also just use the normal `extract_features` function with the `select_features` function, which is basically the same.

nils-braun:
I added the parameter in #648

nils-braun:
Now also `roll_time_series` has multiprocessing capabilities.

Issue #638, opened 2020-02-27
tsfresh.utilities.dataframe_functions.make_forecasting_frame, add option to remove ids with lesser lags


jontis:
tsfresh.utilities.dataframe_functions.make_forecasting_frame conveniently builds a dataframe with lags from a time series and takes the parameter max_timeshift for number of lags to include. A few of the id's in the end will get less than max_timeshift lags. Running feature extraction on those will thus give very different results than if they had max_timeshift lags. I propose to include an option to filter out those ids that did not get max_timeshift lags.

nils-braun:
Yes, very good proposal!

nils-braun:
Would you like to do a PR for this?

nils-braun:
A function for that was introduced in #668. You can now specify the `min_timeshift` to do exactly this.

PR #668, opened 2020-04-16
Better/Correct/Faster rolling

nils-braun:
`tsfresh` is used for time series forecasting a lot . However, our rolling function is a common source for problems.
This PR tries to solve many of them:
1. it speeds up the process by making the rolling multiprocessing ready. Same as the feature extraction, it will also work with a distributor (if passed) and works with multiprocessing out of the box
2. it corrects/improves the documentation, hopefully it is clearer now
3. it simplifies the code for the rolling logic, so people that want to have a look into the code can read it more clearly what is going on
4. it brings back the new identifiers, which got lost unfortunately at some point. Instead of just writing `<timeshift>` as the new identifier, we now (again) write `id=<id>,timeshift=<timeshift>` as the former has problems if there is more than one id.

This fixes (maybe) #659, #646, #638 (well, at least partly), #673.

@MaxBenChrist what do you think. I need your input here.

### DIFF
--- a/docs/text/forecasting.rst
+++ b/docs/text/forecasting.rst
@@ -32,7 +32,7 @@ The following image illustrates the process:
 
 
 Another example can be found in streaming data, e.g. in Industry 4.0 applications.
-Here you typically get one new data row at a time and use this to for example predict machine failures. To train your model,
+Here you typically get one new data row at a time and use this to, for example, predict machine failures. To train your model,
 you could act as if you would stream the data, by feeding your classifier the data after one time step,
 the data after the first two time steps etc.
 
@@ -45,14 +45,7 @@ Let's walk through an example to see how it works:
 The rolling mechanism
 ---------------------
 
-The rolling mechanism takes a time series :math:`x` with its data rows :math:`[x_1, x_2, x_3, ..., x_n]`
-and creates :math:`n` new time series :math:`\hat x^k`, each of them with a different consecutive part
-of :math:`x`:
-
-.. math::
-    \hat x^k = [x_k, x_{k-1}, x_{k-2}, ..., x_1]
-
-To see what this does in real-world applications, we look into the following example flat DataFrame in tsfresh format
+We look into the following example flat DataFrame in tsfresh format
 
 +----+------+----+----+
 | id | time | x  | y  |
@@ -88,6 +81,7 @@ If you want to follow along, here is the python code to generate this data:
 Now, we can use :func:`tsfresh.utilities.dataframe_functions.roll_time_series` to get consecutive sub-time series.
 You could think of having a window sliding over your time series data and extracting out every data you can see through this window.
 There are three parameters to tune the window:
+
 * `rolling_direction`: if you want to slide in positive (increasing sort) or negative (decreasing sort) direction. Default is positive.
 * `max_timeshift` defines, how large the window size will grow. This means the extracted time series will have at maximum `max_timeshift + 1` steps in the past (or future). Default is infinite.
 * `min_timeshift` defines the minimal size. Defaults to 0.
--- a/docs/text/forecasting.rst
+++ b/docs/text/forecasting.rst
@@ -20,8 +20,7 @@ The rolling utilities implemented in `tsfresh` help you in this process of resha
 Please note that "time" does not necessarily mean clock time here.
 The "sort" column of a DataFrame in the supported :ref:`data-formats-label` gives a sequential state to the
 individual measurements.
-In the case of time series this can be the *time* dimension while in the case of spectra the
-order is given by the *wavelength* or *frequency* dimensions.
+In the case of time series this can be the *time* dimension while in other cases, this can be a location, a frequency. etc.
 
 The following image illustrates the process:
 
@@ -64,7 +63,7 @@ We look into the following example flat DataFrame in tsfresh format
 +----+------+----+----+
 
 where you have measured the values from two sensors x and y for two different entities (id 1 and 2) in 4 or 2 time
-steps (t1 to t9).
+steps (1 to 9).
 
 If you want to follow along, here is the python code to generate this data:
 
@@ -82,9 +81,9 @@ Now, we can use :func:`tsfresh.utilities.dataframe_functions.roll_time_series` t
 You could think of having a window sliding over your time series data and extracting out every data you can see through this window.
 There are three parameters to tune the window:
 
-* `rolling_direction`: if you want to slide in positive (increasing sort) or negative (decreasing sort) direction. Default is positive.
-* `max_timeshift` defines, how large the window size will grow. This means the extracted time series will have at maximum `max_timeshift + 1` steps in the past (or future). Default is infinite.
-* `min_timeshift` defines the minimal size. Defaults to 0.
+* `max_timeshift` defines, how large the window size will grow. This means the extracted time series will have at maximum `max_timeshift + 1` steps in the past (or future).
+* `min_timeshift` defines the minimal size.
+* Advanced: `rolling_direction`: if you want to slide in positive (increasing sort) or negative (decreasing sort) direction. You barely need negative direction, so you probably not want to change the default.
 
 The column parameters are the same as in the usual :ref:`data-formats-label`.
 
--- a/docs/text/forecasting.rst
+++ b/docs/text/forecasting.rst
@@ -51,11 +51,11 @@ We look into the following example flat DataFrame in tsfresh format
 +====+======+====+====+
 | 1  |  1   | 1  | 5  |
 +----+------+----+----+
-| 1  |  2   | 2	 | 6  |
+| 1  |  2   | 2  | 6  |
 +----+------+----+----+
-| 1  |  3   | 3	 | 7  |
+| 1  |  3   | 3  | 7  |
 +----+------+----+----+
-| 1  |  4   | 4	 | 8  |
+| 1  |  4   | 4  | 8  |
 +----+------+----+----+
 | 2  |  8   | 10 | 12 |
 +----+------+----+----+
@@ -63,7 +63,7 @@ We look into the following example flat DataFrame in tsfresh format
 +----+------+----+----+
 
 where you have measured the values from two sensors x and y for two different entities (id 1 and 2) in 4 or 2 time
-steps (1 to 9).
+steps (1, 2, 3, 4, 8, 9).
 
 If you want to follow along, here is the python code to generate this data:
 
@@ -174,6 +174,8 @@ You will end up with features generated for each of the parts above, which you c
 +------------------+----------------+-----------------------------+-----+
 | id=2,timeshift=9 |          221.0 |                         1.0 | ... |
 +------------------+----------------+-----------------------------+-----+
+
+The features for e.g. ``id=1,timeshift=3`` are extracted using the data up to and including ``t=3`` (so ``t=1``, ``t=2`` and ``t=3``).
 
 If you want to train for a forecasting, `tsfresh` also offers the function :func:`tsfresh.utilities.dataframe_functions.make_forecasting_frame`, which will also help you match the target vector properly.
 This process is also visualized by the following figure.
--- a/docs/text/forecasting.rst
+++ b/docs/text/forecasting.rst
@@ -1,22 +1,27 @@
 .. _forecasting-label:
 
-Time series forecasting
-=======================
+Time series forecasting/Rolling
+===============================
 
 Features that are extracted with *tsfresh* can be used for many different tasks, such as time series classification,
 compression or forecasting.
 This section explains how one can use the features for time series forecasting tasks.
 
-The "sort" column of a DataFrame in the supported :ref:`data-formats-label` gives a sequential state to the
-individual measurements. In the case of time series this can be the *time* dimension while in the case of spectra the
-order is given by the *wavelength* or *frequency* dimensions.
-We can exploit this sequence to generate more input data out of a single time series, by *rolling* over the data.
-
 Lets say you have the price of a certain stock, e.g. Apple, for 100 time steps.
 Now, you want to build a feature-based model to forecast future prices of the Apple stock.
-So you will have to extract features in every time step of the original time series while looking at
-a certain number of past values.
-A rolling mechanism will give you the sub time series of last *m* time steps to construct the features.
+You could remove the last price value (of today) and extract features from the time series until today to predict the price of today.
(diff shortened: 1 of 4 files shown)


tsfresh/utilities/dataframe_functions.py
```diff
@@ -440,57 +528,76 @@ def roll_time_series(df_or_dict, column_id, column_sort, column_kind, rolling_di
...
 
-    if np.isnan(max_timeshift):
+    if np.isnan(prediction_steps):
         raise ValueError("Somehow the maximum length of your time series is NaN (Does your time series container have "
                          "only one row?). Can not perform rolling.")
+    max_timeshift = max_timeshift or prediction_steps
+
+    # Todo: not default for columns_sort to be None
+    if column_sort is None:
+        df["sort"] = range(df.shape[0])
```
MaxBenChrist, line 564: don't you have to set `columns_sort` to `sort` here?

tsfresh/utilities/dataframe_functions.py
```diff
@@ -440,57 +528,76 @@ def roll_time_series(df_or_dict, column_id, column_sort, column_kind, rolling_di
...
 
     if rolling_direction > 0:
-        range_of_shifts = range(max_timeshift, -1, -1)
+        range_of_shifts = range(1, prediction_steps + 1)
```
MaxBenChrist, line 567: both branches are identical, so maybe omit the if block?

tsfresh/utilities/dataframe_functions.py
```diff
@@ -372,7 +375,55 @@ def _normalize_input_to_internal_representation(timeseries_container, column_id,
...
+    """
+    Internal helper function for roll_time_series.
+    This function has the task to extract the rolled forecast data frame of the number `time_shift`.
+    This means it is `time_shift` in the future - starting counting from the first row
+    for each id (and kind).
+    This means we cut out the data until `time_shift`.
+    The first row we cut out is either 0 or given by the maximal allowed length of `max_timeshift`.
+    for a negative rolling direction it is reversed
+    """
+    def _f(x):
```
MaxBenChrist, line 389: I would expose `time_shift`, `max_timeshift`, etc. as parameters of _f, makes it also easier to test that method.

could rename `time_shift` to start here as well

tsfresh/utilities/dataframe_functions.py
```diff
@@ -372,7 +375,55 @@ def _normalize_input_to_internal_representation(timeseries_container, column_id,
     return timeseries_container, column_id, column_kind, column_value
 
 
-def roll_time_series(df_or_dict, column_id, column_sort, column_kind, rolling_direction, max_timeshift=None):
+def _roll_out_time_series(time_shift, grouped_data, rolling_direction, max_timeshift, min_timeshift,
```
MaxBenChrist, line 378: we should align the spelling of the arguments, e.g. time_shift versus max_timeshift, maybe remove the `time` from those argument names?

tsfresh/utilities/dataframe_functions.py
```diff
@@ -372,7 +375,55 @@ def _normalize_input_to_internal_representation(timeseries_container, column_id,
...
+            shift_from = max(shift_until - max_timeshift - 1, 0)
+
+            df_temp = x.iloc[shift_from:shift_until] if shift_until <= len(x) else None
+        else:
+            shift_from = max(time_shift - 1, 0)
+            shift_until = shift_from + max_timeshift + 1
+
+            df_temp = x.iloc[shift_from:shift_until]
+
+        if df_temp is None or len(df_temp) < min_timeshift + 1:
```
MaxBenChrist, line 401: is +1 really needed?

tsfresh/utilities/dataframe_functions.py
```diff
@@ -372,7 +375,55 @@ def _normalize_input_to_internal_representation(timeseries_container, column_id,
...
+    Internal helper function for roll_time_series.
+    This function has the task to extract the rolled forecast data frame of the number `time_shift`.
+    This means it is `time_shift` in the future - starting counting from the first row
+    for each id (and kind).
+    This means we cut out the data until `time_shift`.
+    The first row we cut out is either 0 or given by the maximal allowed length of `max_timeshift`.
+    for a negative rolling direction it is reversed
+    """
+    def _f(x):
+        if rolling_direction > 0:
```
MaxBenChrist, line 390: I am a little confused, in this branch, we shift backwards in time, but the docstring below says that 

> :param rolling_direction: The sign decides, if to roll backwards or forwards in "time"

so, for negative sign, we should roll back in time and not for a positive sign as implemented here?

docs/text/forecasting.rst
```diff
@@ -26,290 +31,203 @@ The following image illustrates the process:
...
+      "time": [1, 2, 3, 4, 8, 9],
+      "x": [1, 2, 3, 4, 10, 11],
+      "y": [5, 6, 7, 8, 12, 13],
+   })
 
+Now, we can use :func:`tsfresh.utilities.dataframe_functions.roll_time_series` to get consecutive sub-time series.
+You could think of having a window sliding over your time series data and extracting out every data you can see through this window.
+There are three parameters to tune the window:
+
+* `rolling_direction`: if you want to slide in positive (increasing sort) or negative (decreasing sort) direction. Default is positive.
```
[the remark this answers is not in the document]
nils-braun: That is actually a good point. I basically kept it for consistency (and somehow backward compatibility). In principle, foreward and backward rolling does the same, it is just a matter of nomenclature (and what happens on the borders). I would keep the functionality, just because we always had it. But honestly, I also do not see a lot of use cases...

docs/text/forecasting.rst
```diff
@@ -26,290 +31,203 @@ The following image illustrates the process:
...
+.. code:: python
+
+   from tsfresh.utilities.dataframe_functions import roll_time_series
+   df_rolled = roll_time_series(df, column_id="id", column_sort="time")
+
+The new data set consists only of values from the old data set, but with new indices.
+If you group by index, you will end up with the following parts:
+
++-----------------+-------+---+----+
+|id               | time  | x |  y |
```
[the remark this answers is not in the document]
nils-braun: I mean, we do have the id in the dataframe included in the new id. But I see your point.
The reason I did not want to include it, is that people should not mix it up with the time series values, e.g. if you "just" feed this into `extract_features` afterwards you would always need to remove the old `id` column again.

tsfresh/utilities/dataframe_functions.py
```diff
@@ -440,57 +528,76 @@ def roll_time_series(df_or_dict, column_id, column_sort, column_kind, rolling_di
...
-        differences = df.groupby(grouper)[column_sort].apply(
-            lambda x: x.values[:-1] - x.values[1:])
-        # Write all of them into one big list
-        differences = sum(map(list, differences), [])
-        # Test if all differences are the same
-        if differences and min(differences) != max(differences):
-            warnings.warn("Your time stamps are not uniformly sampled, which makes rolling "
-                          "nonsensical in some domains.")
+            differences = df.groupby(grouper)[column_sort].apply(
+                lambda x: x.values[:-1] - x.values[1:])
```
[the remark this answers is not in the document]
nils-braun: I could (actually, I copied this from the code before - but still a valid comment) - but this would change the types and would therefore a bit more `Series` to `list` conversion stuff. I would prefer to keep it for now.

tsfresh/utilities/dataframe_functions.py
```diff
@@ -440,57 +528,76 @@ def roll_time_series(df_or_dict, column_id, column_sort, column_kind, rolling_di
...
+            # Write all of them into one big list
+            differences = sum(map(list, differences), [])
+            # Test if all differences are the same
+            if differences and min(differences) != max(differences):
+                warnings.warn("Your time stamps are not uniformly sampled, which makes rolling "
```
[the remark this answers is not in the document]
nils-braun: Moved to its own issue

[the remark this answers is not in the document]
nils-braun: Therefore closing it here.

tsfresh/utilities/dataframe_functions.py
```diff
@@ -440,57 +528,76 @@ def roll_time_series(df_or_dict, column_id, column_sort, column_kind, rolling_di
...
 
     # Roll the data frames if requested
     rolling_direction = np.sign(rolling_direction)
 
     grouped_data = df.groupby(grouper)
-    max_timeshift = max_timeshift or grouped_data.count().max().max()
+    prediction_steps = grouped_data.count().max().max()
 
-    if np.isnan(max_timeshift):
+    if np.isnan(prediction_steps):
```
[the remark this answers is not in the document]
nils-braun: No, unfortunately not:
```
>>>d f = pd.DataFrame(columns=["id", "value"])
>>> df.groupby("id").count().max().max()
nan
```
Actually, you have introduced this check 3 years ago :-D 
https://github.com/blue-yonder/tsfresh/pull/173/files

But yes, it should say "zero rows", not "one row"

tsfresh/utilities/dataframe_functions.py
```diff
@@ -440,57 +528,76 @@ def roll_time_series(df_or_dict, column_id, column_sort, column_kind, rolling_di
...
         raise ValueError("Somehow the maximum length of your time series is NaN (Does your time series container have "
                          "only one row?). Can not perform rolling.")
+    max_timeshift = max_timeshift or prediction_steps
+
+    # Todo: not default for columns_sort to be None
```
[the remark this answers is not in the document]
nils-braun: Don't know, you added it in #249 ;-)

tsfresh/utilities/dataframe_functions.py
nils-braun: I handled this case in the `_roll_out_time_series` separately. Because the sort column is an increasing number throughout all `id`s and `kind`s in this case, the index build from the old `id` and the `sort` column does not make much sense. Example:

You fill in data like
```
id  value
1  a
1  b
1  c
2  d
2  e
```
and the added sort column will be
```
id  value sort
1  a        1
1  b        2
1  c        3
2  d        4
2  e         5
```

If I would use that value, we would end up with some ids names `id=2,timeshift=5`, which would not make much sense (in my opinion). Rather, they are now named `id=2,timeshift=0` and `id=2,timeshift=1` .

tsfresh/utilities/dataframe_functions.py
nils-braun: The way I see it: If you request a minimal time shift of 3, you want to have all the data which includes at least 3 time steps in the past.
In our nomenclature, that would be "t, t-1, t-2, t-3", so this would give us a length of 4.

tsfresh/utilities/dataframe_functions.py
nils-braun: No, we do actually shift forward in time.
Basically the only difference between forward and backward is the way we handle the borders and how we name the sub-time-series.

Example: if we have a window size of 3 and we shift in positive direction, it will at some point include the times [t4, t5, t6]. We would name this "t6" in this case.. If we shift backwards, it will also include this range, but we would name it "t4" in this case. 

If you image a window shifting over your time series and you move it forwards (positive), you would first only include the very first row, then the first and second, then the first second and third... So the right edge of your window increases with the `time_shift` parameters.

tsfresh/utilities/dataframe_functions.py
nils-braun: The reason I did not expose them, is that I can then just call `groupby.apply(_f)` and not
`groupby.apply(lambda x: _f(x, param, param, param)`.

Please note, that the `time_shift` parameter is not the start point!

tsfresh/utilities/dataframe_functions.py
nils-braun: Ah damn it. I have seen our confusion.
In my nomenclature, the direction is the direction of the shift of the window.
In yours, it is the data seen at each step. I will add some more docstrings here.

tsfresh/utilities/dataframe_functions.py
nils-braun: I am ok with aligning, but I would want to keep the "time". As you have said, rolling barely makes sense in non-time domains.

nils-braun:
@MaxBenChrist Thanks for this super review!
I included all your comments or replied to them. I also increased documentation again.
Could you have another look?

MaxBenChrist:
> @MaxBenChrist Thanks for this super review!
> I included all your comments or replied to them. I also increased documentation again.
> Could you have another look?

tsfresh/utilities/dataframe_functions.py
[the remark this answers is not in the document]
MaxBenChrist: Then I would move that check upwards and add a check if df is empty instead of checking if the prediction_step is nan - in favor of explicitly.

tsfresh/utilities/dataframe_functions.py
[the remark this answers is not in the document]
MaxBenChrist: Well, I don't remember what I intended. Can you remove the todo?

MaxBenChrist:
Good work nils!

MaxBenChrist:
closing this wa unintended

tsfresh/utilities/dataframe_functions.py
```diff
@@ -372,48 +375,167 @@ def _normalize_input_to_internal_representation(timeseries_container, column_id,
...
+    Each cut-out seen by the window is a new time series with a new identifier.
 
     A few remarks:
 
      * This method will create new IDs!
-     * The sign of rolling defines the direction of time rolling, a positive value means we are going back in time
-     * It is possible to shift time series of different lengths but
+     * The sign of rolling defines the direction of time rolling, a positive value means we are shifting
+       the cut-out window foreward in time. The name of each new sub time series is given by the last time point.
+       This means, the time series named `id=4,timeshift=5` with a `max_timeshift` of 3 includes the data
```
MaxBenChrist, line 470: includes the data of the times 3,2 and 1, so it can be used to forecast time point for id=4

nils-braun: No :-D
That is the point. The `id` is the old `id`, the `timeshift` is the time parameter.
So it includes the times 3, 4 and 5 and can be used to predict 6.

nils-braun:
Thanks for the review @MaxBenChrist. As you said, it is still too complicated and should be rewritten with the "window"" in mind and without a negative rolling direction. I will leave this for the next PR.

PR #689, opened 2020-05-15
Created benford_correlation(x) function

gubertoli:
Hi,

I have been using tsfresh for about 2 months, one attribute that came up to my mind in my cases is to use the deviation from Newcomb-Benford distribution, specifically the correlation between a data array and the expected distribution.

Newcomb-Benford distribution considers a log distribution of the first digits from data.

This analysis has been used for fraud and anomaly detection. More uses - https://scholar.google.com/scholar?q=benford+law+time+series

_This is my first pull request to a major package, please if anything is wrong, please let me know._

### DIFF
--- a/tests/units/feature_extraction/test_feature_calculations.py
+++ b/tests/units/feature_extraction/test_feature_calculations.py
@@ -1258,9 +1258,9 @@ class FeatureCalculationTestCase(TestCase):
         # A list containing NaN
         list_with_nan = [1.354, 0.058, 0.055, 0.99, 3.15, np.nan, 0.3, 2.3, 0, 0.59, 0.74]
 
-        self.assertAlmostEqual(benford_correlation, random_list, 0.39458056)
-        self.assertAlmostEqual(benford_correlation, fibonacci, 0.998)
-        self.assertAlmostEqual(benford_correlation, list_with_nan, 0.10357511)
+        self.assertAlmostEqual(benford_correlation(random_list), 0.39458056)
+        self.assertAlmostEqual(benford_correlation(fibonacci_list), 0.998003988)
+        self.assertAlmostEqual(benford_correlation(list_with_nan), 0.10357511)
         self.assertIsNaN(benford_correlation(equal_list))
 
 
--- a/tests/units/feature_extraction/test_feature_calculations.py
+++ b/tests/units/feature_extraction/test_feature_calculations.py
@@ -23,10 +23,10 @@ class FeatureCalculationTestCase(TestCase):
                          msg="Not equal for lists: {} != {}".format(f(input_to_f, *args, **kwargs), result))
         self.assertEqual(f(np.array(input_to_f), *args, **kwargs), result,
                          msg="Not equal for numpy.arrays: {} != {}".format(
-                         f(np.array(input_to_f), *args, **kwargs), result))
+                             f(np.array(input_to_f), *args, **kwargs), result))
         self.assertEqual(f(pd.Series(input_to_f), *args, **kwargs), result,
                          msg="Not equal for pandas.Series: {} != {}".format(
-                         f(pd.Series(input_to_f), *args, **kwargs), result))
+                             f(pd.Series(input_to_f), *args, **kwargs), result))
 
     def assertTrueOnAllArrayTypes(self, f, input_to_f, *args, **kwargs):
         self.assertTrue(f(input_to_f, *args, **kwargs), msg="Not true for lists")
@@ -323,8 +323,8 @@ class FeatureCalculationTestCase(TestCase):
         self.assertEqualOnAllArrayTypes(cid_ce, [-4.33, -1.33, 2.67], 5, normalize=False)
 
     def test_lempel_ziv_complexity(self):
-        self.assertAlmostEqualOnAllArrayTypes(lempel_ziv_complexity, [1, 1, 1], 2./3, bins=2)
-        self.assertAlmostEqualOnAllArrayTypes(lempel_ziv_complexity, [1, 1, 1], 2./3, bins=5)
+        self.assertAlmostEqualOnAllArrayTypes(lempel_ziv_complexity, [1, 1, 1], 2. / 3, bins=2)
+        self.assertAlmostEqualOnAllArrayTypes(lempel_ziv_complexity, [1, 1, 1], 2. / 3, bins=5)
 
         self.assertAlmostEqualOnAllArrayTypes(lempel_ziv_complexity, [1, 1, 1, 1, 1, 1, 1],
                                               0.4285714285, bins=2)
@@ -403,7 +403,7 @@ class FeatureCalculationTestCase(TestCase):
         self.assertIsNanOnAllArrayTypes(standard_deviation, [])
 
     def test_variation_coefficient(self):
-        self.assertIsNanOnAllArrayTypes(variation_coefficient, [1, 1, -1, -1],)
+        self.assertIsNanOnAllArrayTypes(variation_coefficient, [1, 1, -1, -1], )
         self.assertAlmostEqualOnAllArrayTypes(variation_coefficient, [1, 2, -3, -1], -7.681145747868608)
         self.assertAlmostEqualOnAllArrayTypes(variation_coefficient, [1, 2, 4, -1], 1.2018504251546631)
         self.assertIsNanOnAllArrayTypes(variation_coefficient, [])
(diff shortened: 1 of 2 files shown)


tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -2083,3 +2083,43 @@ def count_below(x, t):
...
+   :return: the value of this feature
+   :return type: float
+   """
+    if not isinstance(x, (np.ndarray, pd.Series)):
+        x = np.asarray(x)
+
+    x = np.trunc(x)  # retrieve first digit from data
+
+    # benford distribution
+    benford_distribution = np.array([0.301, 0.176, 0.125, 0.097, 0.079, 0.067, 0.058, 0.051, 0.046])
```
nils-braun, line 2120: Using "magical numbers" (= hard-coded number values) in code is not considered good practice for some reasons:
1. complicated to change, e.g. to add more significant digits (because I would need to re-calculate by hand)
2. error-prone (everyone reading your code would need to re-calculate to make sure, it is what the comment says)
3. we do not expect it here, but: if something in the original distribution changes, how can we make sure the correct values are used?

So I would propose the following:
* either calculate the distribution on the fly using its formula `log10(n + 1) - log10(n)`
* or calculate it once, store it into a constant outside of the function and use it here (preferred)

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -2083,3 +2083,43 @@ def count_below(x, t):
...
+
+    data_distribution = np.array([(x == 1).mean(), (x == 2).mean(), (x == 3).mean(), (x == 4).mean(), (x == 5).mean(),
+                                  (x == 6).mean(), (x == 7).mean(), (x == 8).mean(), (x == 9).mean()])
```
nils-braun, line 2123: This can (and should) be replaced by a loop, because it contains a lot of repeated code.
You can replace this with `np.array([(x == n).mean() for n in range(1, 10)])`

Said that, it can be even replaced with some numpy internal function (which we should prefer, as they are in general faster and better tested). What you are doing here, is to calculate how often a certain value occurs and divide it by the total number of data entries.
This can be achieved with 
```python
_, data_distribution = np.unique(x, return_counts=True)
data_distribution /= len(x)
```
(but this only works if you make sure, that only numbers between 0 and 10 are present, see above)

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -2083,3 +2083,43 @@ def count_below(x, t):
...
+
+    return np.corrcoef(benford_distribution, data_distribution)[0, 1]
```
nils-braun, line 2125: Please add a comment on what this line does (because - not your fault - the "corrcoef" is not a very telling name in my opinion) and why you especially take the `[0, 1]` component.

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -2083,3 +2083,43 @@ def count_below(x, t):
...
+
+   :param x: the time series to calculate the feature of
+   :type x: numpy.ndarray
+   :return: the value of this feature
+   :return type: float
+   """
+    if not isinstance(x, (np.ndarray, pd.Series)):
+        x = np.asarray(x)
+
+    x = np.trunc(x)  # retrieve first digit from data
```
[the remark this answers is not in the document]
gubertoli: Once the Benford Law is evidenced on decimal numbers also, I am proposing the following:
> ```python
> >>> x = np.array([1, 0.02, 3990.5, 4.7, 0.51, -0.006, -7.2, -8.9, 9])
> >>> np.array([int(str(np.format_float_scientific(i))[:1]) for i in np.abs(x)])
> array([1, 2, 3, 4, 5, 6, 7, 8, 9])
> ```

tsfresh/feature_extraction/feature_calculators.py
gubertoli: I choose the first suggestion once the values can be out of the 0-10 range.

tests/units/feature_extraction/test_feature_calculations.py
```diff
@@ -1221,26 +1222,47 @@ def test_count_above(self):
         self.assertEqualPandasSeriesWrapper(count_above, [1] * 10, 1, t=1)
         self.assertEqualPandasSeriesWrapper(count_above, list(range(10)), 1, t=0)
         self.assertEqualPandasSeriesWrapper(count_above, list(range(10)), 0.5, t=5)
-        self.assertEqualPandasSeriesWrapper(count_above, [0.1, 0.2, 0.3] * 3, 2/3, t=0.2)
-        self.assertEqualPandasSeriesWrapper(count_above, [np.NaN, 0, 1] * 3, 2/3, t=0)
-        self.assertEqualPandasSeriesWrapper(count_above, [np.NINF, 0, 1] * 3, 2/3, t=0)
+        self.assertEqualPandasSeriesWrapper(count_above, [0.1, 0.2, 0.3] * 3, 2 / 3, t=0.2)
```
nils-braun, line 1225: I would probably not format it like this, but PEP8 does not enforce any of the two possibilities (https://stackoverflow.com/questions/18171560/does-pep-8-require-whitespace-around-operators-in-function-arguments). So I will leave it as it is if you think it is better like this!

tests/units/feature_extraction/test_feature_calculations.py
```diff
@@ -1221,26 +1222,47 @@ def test_count_above(self):
...
+        self.assertEqualPandasSeriesWrapper(count_below, [np.NINF, 0, 1] * 3, 2 / 3, t=0)
+        self.assertEqualPandasSeriesWrapper(count_below, [np.PINF, 0, 1] * 3, 1 / 3, t=0)
         self.assertEqualPandasSeriesWrapper(count_below, [np.NaN, 0, 1] * 3, 0, t=np.NaN)
-        self.assertEqualPandasSeriesWrapper(count_below, [np.NINF, 0, np.PINF] * 3, 1/3, t=np.NINF)
+        self.assertEqualPandasSeriesWrapper(count_below, [np.NINF, 0, np.PINF] * 3, 1 / 3, t=np.NINF)
         self.assertEqualPandasSeriesWrapper(count_below, [np.PINF, 0, 1] * 3, 1, t=np.PINF)
 
+    def test_benford_correlation(self):
+        # A test with list of random values
+        np.random.seed(42)
```
nils-braun, line 1247: nice

tests/units/feature_extraction/test_feature_calculations.py
```diff
@@ -1221,26 +1222,47 @@ def test_count_above(self):
...
+        random_list = np.random.uniform(size=100)
+
+        # Fibonacci series is known to match the Newcomb-Benford's Distribution
+        fibonacci_list = [0, 1]
+        for i in range(2, 200):
+            fibonacci_list.append(fibonacci_list[i - 1] + fibonacci_list[i - 2])
+
+        # A list of equally distributed digits (returns NaN)
```
nils-braun, line 1255: Is this expected to return nan? If so, why?

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -2083,3 +2083,44 @@ def count_below(x, t):
...
+   :return: the value of this feature
+   :return type: float
+   """
+    x = np.asarray(x)
+
+    # retrieve first digit from data
+    x = np.array([int(str(np.format_float_scientific(i))[:1]) for i in np.abs(np.nan_to_num(x))])
+
+    # benford distribution
+    benford_distribution = np.array([np.log10(1+1/n) for n in range(1, 10)])
```
nils-braun, line 2120: Funnily enough, I would add spaces here (around the "+") :-D

tests/units/feature_extraction/test_feature_calculations.py
gubertoli: This change was automatically made by the IDE (PyCharm CE 2020.1)

tests/units/feature_extraction/test_feature_calculations.py
gubertoli: The correlation between two variables x and y is:



And the covariance between them:



Besides covariance being zero for a series with equal values,  and  are the variances for each series and it will be zero for the series that contains the same values (no variance et all), so the correlation calculus will have a 0/0 (NaN)

tsfresh/feature_extraction/feature_calculators.py
gubertoli: Just to be consistent and I think that a space will support readability, I inserted it.

nils-braun:
Thanks again @gubertoli for your work! Appreciate the changes!

gubertoli:
> Thanks again @gubertoli for your work! Appreciate the changes!

Thank you @nils-braun for the patience and all the processual knowledge shared during this PR.

Issue #699, opened 2020-05-20
Sorting of created id column from rolling is awkward


TKlerx:
Hi there,

it has been some time since last working with tsfresh and I am impressed what you implemented in the mean time.
I am using the rolling feature and the newly created id column is not sorted as it should be.

1. Your operating system: Win10x64
2. The version of *tsfresh* that you are using: 0.16.0

Here is a minimal example:
```
import pandas as pd
from tsfresh.utilities.dataframe_functions import roll_time_series
from tsfresh import extract_features
df = pd.DataFrame({
   "initial_id": [1]*12,
   "time": [1, 2, 3, 4,5,6,7,8,9,10,20,30],
   "x": [1, 2, 3, 4,5,6,7,8,9,10,20,30]
})
window_size = 2
df_rolled = roll_time_series(df, column_id="id", column_sort="time", min_timeshift=window_size-1, max_timeshift=window_size-1)
df_rolled = df_rolled.drop('initial_id', axis = 1)
df_features = extract_features(df_rolled, column_id="id", column_sort="time", n_jobs = 8)
```
After sorting, the first id is `id=1,timeshift=10` and not `id=1,timeshift=1`.

I guess the issue is that the sort column is added as a string and if it is an integer, a '1' is not padded with some leading 0s.
This is not a real bug, but it is somehow awkward, because 10 is greater than 2 when treating 10 and 2 as integers instead of strings.

nils-braun:
Welcome back, @TKlerx :-)

That is a very valid point. Unfortunately, we have to store multiple information (the old id and the timeshift) in a single column.
Would it be better if the column is a tuple (old id, timeshift) with the numerical values still present? We somehow would loose the "interpretability" a bit though...

TKlerx:
Wow, that was a quick answer. I found that my example was not sufficient and wanted to edit a bit.
Actually, the problem arises when passing `df_rolled` to `extract_features` because I need to get rid of the old id and time, but only use the new id (id=<old_id>, timeshift) of type string.

I thought about a bigger change to be able to pass more than one sorting column to the `extract_features` method.

nils-braun:
Making the sorting handle more than one columns is easy, the problem is that we would need to have the "column_id" handle more than one column.
In principle, that is not so hard - in practice it is of course all over the place.

Thanks for the heads-up on the memory consumption.
If you want, I can have a look into this - but I am happy for contributions :-)

TKlerx:
[...]

PR #700, opened 2020-05-21
Changed constructed id in roll_time_series from string to tuple

TKlerx:
As discussed in #699 I changed the construction of the new id in roll_time_series from string to tuple. 
Also wrote a testcase to test whether the sorting really works as intended on `column_sort` if the column is integer containing integers of different length (number of digits).


TKlerx:
Do you know why this fails for pandas v0.20.3?
The failures do not make sense to me.

nils-braun:
@TKlerx I can only guess, but I think in this older version of pandas the logic to have a multi dimensional cell was not well supported.
What I think it does: the dataframe has actually 4 columns (id, kind, time, value) and just a single row (at least when it fails). The apply should create a single entity, which is a 2-tuple. pandas tries to match this 2-tuple with the 4 columns (which is obviously wrong). Do not ask me why it does so...

But, it can be fixed :-)
By moving the `[column_id]` to the dataframe before applying, you turn it into a series and no confusion is possible:
```python
df_temp["id"] = df_temp[column_id].apply(lambda row: (row, timeshift_value))
```
I just did a quick test with this old pandas version and it seems to work properly.

nils-braun:
Looks good to me @TKlerx! Thanks for working on this.

Before we merge this, we would need to fix the documentation for the rolling/forecasting (both the docstrings, although the should be mostly fine already and the sphinx docu). Would you like to work on this or shall I propose a change?

TKlerx:
Should be done, I guess. Could you please check that I did not miss anything?

nils-braun:
Very good! Thanks for your contribution.

HaveF:
hi, @nils-braun 
Do we have any plan to release this feature?
I see this on latest doc, but it seems  not release on code.

nils-braun:
Yes, it is in the latest master version that you can install via pip install -e git+https://github.com/blue-yonder/tsfresh.

You are right however, that it is not (yet) in a properly released version. We are still discussing on some internal details on some other topics (see the currently open PRs) which I would love to have fixed before we do the next release. 
Is installation from source (or as I have shown above) no option for you?

HaveF:
@nils-braun Thanks. Just a bit curious the out of sync about the doc and code. Git version is fine. Thanks again.

nils-braun:
Ok! We (and many other projects) do have multiple versions of the documentation. The "latest" always shows the current main branch - here it is included (both in code and docu). Then there is "stable", which is the last released version (0.16.0 in our case). Here, the docu still shows the old format (because, as you correctly stated, it is not in the last release). And then there is of course documentation for every released version.

If you want to only get the documentation of your installed tsfresh version, you should make sure that you choose the correct version of the docs also.

Issue #703, opened 2020-05-25
Feature extraction progress bar fills fast, but some part in the code takes a long time


TKlerx:
1. OS: Linux (16 cores)
2. version: current master branch
3. a rolled time series
```
<class 'pandas.core.frame.DataFrame'>
Int64Index: 627750 entries, 0 to 627749
Data columns (total 9 columns):
 #   Column      Non-Null Count   Dtype  
---  ------      --------------   -----  
 0   ride_id     627750 non-null  uint8  
 1   tick        627750 non-null  uint16 
 2   imu_ax_FFT  627750 non-null  float32
 3   imu_ay_FFT  627750 non-null  float32
 4   imu_az_FFT  627750 non-null  float32
 5   imu_gx_FFT  627750 non-null  float32
 6   imu_gy_FFT  627750 non-null  float32
 7   imu_gz_FFT  627750 non-null  float32
 8   id          627750 non-null  object 
dtypes: float32(6), object(1), uint16(1), uint8(1)
memory usage: 68.2 MB
```

When I run the feature extraction on this rolled dataframe, it takes about 10 minutes until the progress bar is filled to 100% (with using all 16 cores on the system).
Then, it takes a very long time (approx. 20 - 30 minutes. maybe even more) until the feature extraction completes (with just utilizing a single core).

If I read the code correctly, it must happen in this part of extraction.py (after ``_do_extraction_on_chunk``):
```python
    # Return a dataframe in the typical form (id as index and feature names as columns)
    result = pd.DataFrame(result)
    if "value" in result.columns:
        result["value"] = result["value"].astype(float)
    if len(result) != 0:
        result = result.pivot("id", "variable", "value")
        result.index = result.index.astype(df[column_id].dtype)
```
I did not do any profiling, but it must be the creation of the dataframe or the pivotization.
Do you have any idea how to speed this up?
If it is the pivot, could this be done on the workers?

A second question (I don't want to open another issue for this):
Did you consider switching from native pandas to modin to parallelize up e.g. ``concat``s and other pandas operations (https://github.com/modin-project/modin)? I know this would be a bigger change :)

nils-braun:
These are all very valid points you raise.
I think it is probably the pivoting.  I do not know exactly how many ids you have, but I guess the resulting dataframe will have many columns. pivoting is really slow - we have a super old issue to fix this (https://github.com/blue-yonder/tsfresh/issues/417) but so far no one has tackled this.

[...]

Issue #417, opened 2018-08-23
get rid of pivoting on results chunks


MaxBenChrist:
In 
https://github.com/blue-yonder/tsfresh/blob/ba8e3db386257f3156352deb8e8ff2ae3b379400/tsfresh/feature_extraction/extraction.py#L240

we use a pivot command to aggregate chunk results.

I want to profile this command and see if its a performance bottleneck. Maybe we can parallelize it. I am not sure if its not bloating up memory for bigger datasets

nils-braun:
Fixed in #705

PR #705, opened 2020-05-26
Replace pivoting by a simpler function

nils-braun:
Following the discussion in #703, I implemented a custom logic for collecting the results instead of the `pivot` we currently have. This also solves #417.
Actually, implementing the logic was quite simple and we should have done that way back already...

timing

The orange line is the new implementation (which is now linear, which is expected), the blue one is the current one. For very small number of columns it makes no difference, but for larger result dataframes (larger=wider) it is quite visible. I did this with a time series length of 10 and 100 ids.

@TKlerx If you have some spare time, would you mind checking if it also improves the timing for you?

It still has a call to sort, because I do not want to break the API, but sorting on the index should be quite fast. I also tried to think about memory consumption and I actually think this solution should be better - but I am not sure.


dbarbier:
Looks good to me, and since you are now manually crafting `return_df_dict`, you could modify `_do_extraction_on_chunk` to return a list of tuples `(variable, value, id)` instead of a list of dicts to lower memory consumption.
BTW I never understood why `_do_extraction_on_chunk` defines a generator and returns `list(_f())` instead of directly building a list.

tsfresh/feature_extraction/extraction.py
```diff
@@ -313,16 +314,21 @@ def _do_extraction(df, column_id, column_value, column_kind,
                                     function_kwargs=kwargs)
     distributor.close()
 
-    # Return a dataframe in the typical form (id as index and feature names as columns)
-    result = pd.DataFrame(result)
-    if "value" in result.columns:
-        result["value"] = result["value"].astype(float)
+    return_df_dict = defaultdict(dict)
```
MaxBenChrist, line 317: I would write a method for the conversion that now happens in lines 317-323 and test that method.

MaxBenChrist:
> which is now linear, which is expected)

why is it expected to have linear runtime? We do one pass over all values in our code but we don't know the runtime of the pandas dataframe constructor (I guess the `pd.DataFrame` sorts the data).

I am wondering why this new implementation is soo much faster. To do a pivot in any case you need to sort your values by feature type and then sort by id. This was contained in the pivot and is now contained in the `pd.DataFrame` constructor call

nils-braun:
Quick answer concerning your first question: I expected the total runtime of feature extraction to be dominated by the feature calculation itself. This should scale exactly linearly with the number of columns, as they are independent.

Everything on top of the actual feature extraction (e.g. the pivoting) is not linear with he columns (as you have correctly pointed out). However if I see this in the total processing time, this means this pivoting has a large impact in the total time, which I think is bad (and therefore unexpected)

nils-braun:
@MaxBenChrist 
I did some more studies, here are the results. This time I just compared the last final bit of the extract features - with or without pivoting.
I created results with 1000 features and some number of ids (x axis in the plot) and measured how long only the pivoting compared to the new method takes.

pivot_time_only

As you can see, the new method is slower for very small dataframes (but: here the difference is very small in absolute numbers) and gets much better for large number of ids. (it is a log-log plot).
Please note: I did not scale the number of features (= columns), but the number of ids (= rows)!

Now that is unexpected, so I dig a bit deaper.
Here you can see a profile plot of the pivot function:

pivot

compared to the new method

no_pivot

on the same data. As you can see, most of the time is actually not spend in the call to `pivot()` itself (top picture, right column, 5.8s -> comparable to the new method in total), but in the creation of the dataframe that is needed before! It seems that calling the constructor with a very long list of dictionaries (1000*1000 in this case) is quite slow.

So you are correct, @MaxBenChrist, the call to `pivot` and the new implementation do roughly the same thing, so there run time is also comparable. But the old implementation also needed a call to the dataframe constructor, which is not needed now. Therefore it is much faster.
Sorry also to @TKlerx for being wrong on that - it *is* the dataframe constructor, not the pivot to blame here.

@dbarbier : the reason why we have a yield not a list append is somehow historical. But as you have this on mind for so long already: would you like to do a PR to fix this?

tsfresh/feature_extraction/extraction.py
nils-braun: Seems good, will do so.

nils-braun:
I moved to tuples instead of dictionaries, created a new function, some unittests and documentation.
From my point of view, this is now ready to merge. @MaxBenChrist or anyone else, do you have any final comments or feedback?

tsfresh/utilities/dataframe_functions.py
```diff
@@ -787,3 +788,25 @@ def _add_id_column(df_chunk):
...
+    all b's as columns and all c's as values.
+
+    It basically does a pd.pivot(first entry, second entry, third entry),
+    but optimized for non-pandas input (= python list of tuples).
+    """
+    return_df_dict = defaultdict(dict)
+    for result_entry in list_of_tuples:
+        # the result_entry is a tuple with the id, the variable and the value
+        # we turn it into a mapping variable -> id -> value
+        chunk_id, variable, value = result_entry
```
dbarbier, line 808: You could write
```
for chunk_id, variable, value in list_of_tuples:
```

dbarbier:
> @dbarbier : the reason why we have a yield not a list append is somehow historical. But as you have this on mind for so long already: would you like to do a PR to fix this?

Okay, will do after this PR is merged.

tsfresh/utilities/dataframe_functions.py
nils-braun: That is absolutely correct. Thanks for the comment!

TKlerx:
I tried with my data set (300k ids, with 30 rows for each id) and with the new feature it is way faster (but still not really fast).
I guess the problem is the high number of chunks, as every id is in a single chunk (if I am not mistaken).
I will think about whether there is a possibility to somehow gain more speed and will open a new issue with my thoughts if I am somewhat successful.

tests/units/utilities/test_dataframe_functions.py
```diff
@@ -1089,3 +1089,47 @@ def test_dict_input(self):
...
+
+        self.assertEqual(return_df.loc["a", "b"], 1)
+        self.assertEqual(return_df.loc["a", "c"], 2)
+        self.assertEqual(return_df.loc["A", "b"], 3)
+        self.assertEqual(return_df.loc["A", "c"], 4)
+        self.assertEqual(return_df.loc["X", "Y"], 5)
+
+    def test_long_input(self):
+        input_list = []
+        for i in range(100):
```
MaxBenChrist, line 1124: You could use https://docs.python.org/3/library/itertools.html#itertools.product here to get rid of the loops

tsfresh/feature_extraction/extraction.py
```diff
@@ -313,16 +314,18 @@ def _do_extraction(df, column_id, column_value, column_kind,
...
 
-    if len(result) != 0:
-        result = result.pivot("id", "variable", "value")
-        result.index = result.index.astype(df[column_id].dtype)
+    # All the features are floats
+    return_df = return_df.astype(float)
 
-    return result
+    # copy the type of the index
+    return_df.index = return_df.index.astype(df[column_id].dtype)
```
MaxBenChrist, line 323: for which type of index is that recasting necessary? I guess something like numbers that are saved as strings?

are we testing the types of indecies that need this recasting in some unit tests?

tsfresh/feature_extraction/extraction.py
```diff
@@ -334,10 +337,10 @@ def _do_extraction_on_chunk(chunk, default_fc_parameters, kind_to_fc_parameters)
...
     which is then converted to a numpy array - so a single time series.
 
-    Returned is a list of the extracted features. Each one is a dictionary consisting of
-    { "variable": the feature name in the format <kind>__<feature>__<parameters>,
-      "value": the number value of the feature,
-      "id": the id of the chunk }
+    Returned is a list of the extracted features. Each one is a tuple consisting of
+    { the id of the chunk,
+      the feature name in the format <kind>__<feature>__<parameters>,
+      the number value of the feature, }
```
MaxBenChrist, line 343:
```suggestion
      the numeric value of the feature or np.nan , }
```

tsfresh/utilities/dataframe_functions.py
```diff
@@ -787,3 +788,31 @@ def _add_id_column(df_chunk):
...
+    entries: (a, b, c).
+    Out of this, a pandas dataframe will be created with all a's as index,
+    all b's as columns and all c's as values.
+
+    It basically does a pd.pivot(first entry, second entry, third entry),
+    but optimized for non-pandas input (= python list of tuples).
+    """
+    return_df_dict = defaultdict(dict)
+    for chunk_id, variable, value in list_of_tuples:
+        # we turn it into a mapping column -> index -> value
```
MaxBenChrist, line 807:
```suggestion
        # we turn it into a nested mapping `column -> index -> value`
```

tests/units/utilities/test_dataframe_functions.py
nils-braun: That is true - but I wanted to keep the unittest as "simple" as possible :-) 
I am actually not sure which possibility is easier...

tsfresh/feature_extraction/extraction.py
nils-braun: Good question. If I remember correctly, that was mostly needed because of datetime columns.

I will add an issue to dig into this further, possible either removing this line or adding a unittest.
Thinking now a bit on it, I am not sure why this cast would trigger at all (because the type should already be the same).

nils-braun: Raised in #706

nils-braun:
Thank you everyone for your feedback!

tsfresh/feature_extraction/extraction.py
```diff
@@ -313,16 +314,18 @@ def _do_extraction(df, column_id, column_value, column_kind,
...
-    result = pd.DataFrame(result)
-    if "value" in result.columns:
-        result["value"] = result["value"].astype(float)
+    return_df = pivot_list(result)
 
-    if len(result) != 0:
-        result = result.pivot("id", "variable", "value")
-        result.index = result.index.astype(df[column_id].dtype)
+    # All the features are floats
+    return_df = return_df.astype(float)
```
dbarbier, line 320: astype is a hotspot in your profile, you could remove this line if you add `dtype=float` to DataFrame constructor in `pivot_list`

nils-braun: Nice catch. Actually I thought that this line does basically nothing in most of the cases. But yes, I should have been more careful.
Changed that!

PR #710, opened 2020-06-08
Add benchmarking test

nils-braun:
As we are currently talking a lot about performance, I added some benchmarks.
They are also automatically shown here: https://blue-yonder.github.io/tsfresh/dev/bench/ (which is currently of course still a bit empty)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -59,7 +59,6 @@ packages =
 addopts =
     --cov tsfresh --cov-report term-missing
     --verbose
-    --benchmark-min-rounds=2
     -n auto
 
 [aliases]
--- a/setup.cfg
+++ b/setup.cfg
@@ -59,6 +59,7 @@ packages =
 addopts =
     --cov tsfresh --cov-report term-missing
     --verbose
+    --benchmark-min-rounds=2
     -n auto
 
 [aliases]
--- a/tests/benchmark.py
+++ b/tests/benchmark.py
@@ -1,0 +1,44 @@
+# Testcase for benchmarking tsfresh feature extraction and selection
+import pytest
+import pandas as pd
+import numpy as np
+
+from tsfresh import extract_features, extract_relevant_features
+from tsfresh.feature_extraction.settings import ComprehensiveFCParameters, MinimalFCParameters
+
+
+def create_data(time_series_length, num_ids, random_seed=42):
+    np.random.seed(random_seed)
+
+    df = pd.concat([
+        pd.DataFrame({
+            "id": [i] * time_series_length,
+            "time": range(time_series_length),
+            "value": np.random.randn(time_series_length)
+        })
+        for i in range(num_ids)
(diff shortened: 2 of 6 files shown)


dbarbier:
[Numpy](https://pv.github.io/numpy-bench/) and [pandas](https://pandas.pydata.org/speed/pandas) (both links point to benchmarks) use https://github.com/airspeed-velocity/asv maybe it is worth looking at.

nils-braun:
Thanks for the tip @dbarbier !
I am also not 100% satisfied with this tool. Do you know a simple solution for airspeed-velocity to publish the results?

In principle, one would need to do this on a separate, controlled server (not on CI), but I do not have access to one. pandas for example use their own managed server(s) for that. 
What I would like to have is basically a quick test to check, if we are seeing any reasonable large input - and best also an overview per PR (such as the coveralls comment above). But so far, I have not seen anything like this.

setup.cfg
```diff
@@ -59,6 +59,7 @@ packages =
 addopts =
     --cov tsfresh --cov-report term-missing
     --verbose
+    --benchmark-min-rounds=2
```
MaxBenChrist, line 62: wouldn't it make more sense to move that option into the workflow benchmark.yml? We are not conducting a benchmark if we run the test suite, are we?

tests/benchmark.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def create_data(time_series_length, num_ids, random_seed=42):
+    np.random.seed(random_seed)
+
+    df = pd.concat([
+        pd.DataFrame({
+            "id": [i] * time_series_length,
+            "time": range(time_series_length),
+            "value": np.random.randn(time_series_length)
```
MaxBenChrist, line 17: I am not sure but it could be that some of our features actually profit from normally distributed values (e.g. he wavelet features). Would be interesting to use a real world dataset for this benchmarking.

Or as a short term solution, I would cumsum the values. Otherwise its not really a curve but just pure chaos

MaxBenChrist:
I would also mention this in the changelog.

dbarbier:
> Thanks for the tip @dbarbier ! Also looks fine.
> I am not 100% satisfied with the tool I proposed. Do you know a simple solution for airspeed-velocity to publish the results?

Unfortunately not, maybe some company interested in better performances could sponsor a dedicated server?

nils-braun:
@dbarbier That would be the best thing of course :-)
So far, I think we will stick with the GitHub action and live with the fact, that it is not precise. I am still thinking about building a nicer output, maybe as a comment to the PR. Maybe I find some time to build something.

nils-braun:
As I was not satisfied with any of the already present actions, I wrote my own one. It basically just writes the results as a comment to the PR, but this is so far all I want.
Of course, it would be also nice to compare this to the default branch, but that would need some storage of the results.

Issue #713, opened 2020-06-15
extract_relevant_features() doesn't work if y isn't sorted by index


PavloFesenko:
1. Linux 3.10.0-1062.9.1.el7.x86_64
2. tsfresh 0.16.0
3. Sorry, I can't post the data here due to privacy reasons but I will try to describe the problem as well as I can. I have a target for a regression problem but the function `extract_relevant_features()` doesn't return any relevant features. However, if I sort my Pandas DataFrame by "id" before splitting it in "timeseries" and "y", then everything works. I think the problem might be in the function `select_features()` because if I do step-by-step `extract_features()`, `impute()` and `select_features()` without sorting "id", then the relevant features aren't returned as well. By the way, I couldn't reproduce this issue for the robot failures dataset which is a classification problem so it might be only relevant for regression. If there is maybe a toy dataset for regression, then I could provide a minimal code snippet. Thanks a lot!

nils-braun:
Hi @PavloFesenko! 
Thank you very much for your issue! 
I am not quite sure I can understand the issue, but it might look like some indexing problems. What are the types of data (I guess a dataframe for `df` and a pandas series for `y` or only a NumPy array)?

Here is some code to generate a test regression data sample:
```python
from tsfresh import extract_relevant_features
from tsfresh.examples import load_driftbif

df, y = load_driftbif(100, 10, classification=False, seed=42)

print(df.head())
print(y.head())

X = extract_relevant_features(df, y, column_id="id", column_sort="time", column_kind="dimension", column_value="value")
print(X.head())
```
At least for this data sample, I get some extracted features (360 in this case).
Can you somehow map this to your problem?

PavloFesenko:
Hi @nils-braun!

Thank you for your prompt response! 😊

The issue appears when the column "id" is a string instead of an integer (as is the case in my dataset). I added two lines that change the type of the column "id" and the type of the index for `y` to string. If you try the following code, you would only see 1 extracted feature:

``` python
from tsfresh import extract_relevant_features
from tsfresh.examples import load_driftbif

df, y = load_driftbif(100, 10, classification=False, seed=42)

df['id'] = df['id'].astype('str')
y.index = y.index.astype('str')

X = extract_relevant_features(df, y, column_id="id", column_sort="time", column_kind="dimension", column_value="value")
print(X.head())
```

If you, however, change the task to `classification=True`, you will get 360 features.

Hope this will help to debug the issue. 😉

[...]

PR #715, opened 2020-06-16
Make sure X and y are sorted equally in select_features

nils-braun:
This looks like a quite large bug to me. 

In principle, X and y could have the same index entries (what we check), but we never check if the order is the same! So passing in a `X` with index `[0, 1, 2]` and a `y` with index `[2, 1, 0]` will not raise any problems.

For classification tasks, this is not a problem, as all the significance tests except `target_real_feature_real_test` do not care about the order of the index, as `x` and `y` are re-ordered internally again in all the other tasks. But for `target_real_feature_real_test` the index is stripped before any re-order could happen - which leads to the bug described in #713.

This PR fixes this. I am a bit puzzled why this was never a problem before :-/ 
Therefore I added the exact test from the issue, to see any degradation again.
@MaxBenChrist as a general remark: we should add more tests on `select_features` :-)

### DIFF
--- a/tests/integrations/examples/test_driftbif_simulation.py
+++ b/tests/integrations/examples/test_driftbif_simulation.py
@@ -7,6 +7,7 @@ import unittest
 import pandas as pd
 
 from tsfresh.examples.driftbif_simulation import velocity, load_driftbif, sample_tau
+from tsfresh import extract_relevant_features
 
 
 class DriftBifSimlationTestCase(unittest.TestCase):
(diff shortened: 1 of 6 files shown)


tsfresh/feature_selection/selection.py
```diff
@@ -140,6 +140,11 @@ def select_features(X, y, test_for_binary_target_binary_feature=defaults.TEST_FO
     if isinstance(y, np.ndarray):
         y = pd.Series(y, index=X.index)
 
+    y = y.sort_index()
```
MaxBenChrist, line 143: I would move the sorting inside the calculate_relevance_table method. That way it gets more robust

tests/integrations/examples/test_driftbif_simulation.py
```diff
@@ -63,6 +64,18 @@ def test_dimensionality(self):
         self.assertEqual(v.shape, (Nt, 3),
                          'The returned vector should reflect the dimension of the initial condition.')
 
+    def test_relevant_feature_extraction(self):
```
MaxBenChrist, line 67: Why is this test catching the bug?

nils-braun: Because string identifiers are not sorted in the same order as integers, which makes y and the extracted features having a different order.
Any example where y is not order "normally" would be a good example, but as we do not have any other example with string identifiers I thought it might be good to add one anyways

PR #717, opened 2020-06-18
Changing master to main

nils-braun:
This is not a real "pull" request, as I do not want to merge the changes back to master, but actually would like to propose "main" to be our new default branch.
As @MaxBenChrist, @kempa-liehr and I are the only people that can push to this repository directly anyways, that should not be a large problem and will not cause any problems.
Other repositories are currently doing the same and GitHub will implement a change in the future anyway (and I also think it is good).

Once this PR is approved, I will change the default branch to this "main" branch and drop the former master branch.


MaxBenChrist:
what is the reasoning behind this default branch name change from master to main?

nils-braun:
@kempa-liehr I would love to get your opinion before I do the change :-)

kempa-liehr:
Totally fine with the change.

PR #723, opened 2020-06-25
travis to GitHub actions

nils-braun:
Recently, I am not so satisfied with Travis anymore.
* in many cases, it does not show the results in the pull requests (see e.g. here: https://github.com/blue-yonder/tsfresh/pull/722)
* our master branch was failing because some integration tests were failing due to wrong dependencies. I did not notice.
* It is sometimes quite slow in startup

So I am proposing to change to GitHub actions instead.
This PR introduces four important changes:
1. Download the human activity and robot data into temporary files, so that they get downloaded every time (and the download is actually tested - and not used from cache). This is also needed, as GitHub now runs all tests on the same filesystem (and we run 6 tests in parallel)
2. Update our requirements. This is needed for some functions we have. This is unrelated to the switch from Travis to GitHub, but as I now actually run the tests, the branch would fail otherwise
3. Remove the .travis file and add .github actions.
4. Let notebook tests skip by default and only turn them on via an env variable

We have four actions:
* the style check action (which we had already before)
* action for pull requests, which is basically the same as the .travis file before, but now runs all tests except the notebook tests
* action for the master branch, which runs all tests including the notebook tests
* a deployment action (which is so far untested)

There will be more actions in the future, such as the benchmarking actions we discuss in #710


MaxBenChrist:
> our master branch was failing because some integration tests were failing due to wrong dependencies. I did not notice.

How did this happen? Because travis only checks test suite on merging?

nils-braun:
@MaxBenChrist we configured Travis to not check notebooks on each PR, as it takes too long. However, it should check the default branch, but there it did not send out emails correctly (it seems).
Additionally, the GitHub integration did not really work out in many cases, and a non-working test result was not reported back to the GitHub PR.

Issue #724, opened 2020-06-25
Description not matching function name


emanuelef:
Hi,
just noticed a possible mismatch in the documentation, is it possible that the descriptions for 
percentage_of_reoccurring_datapoints_to_all_datapoints  and
percentage_of_reoccurring_values_to_all_values
have been swapped ?

https://github.com/blue-yonder/tsfresh/blob/f8a952e4ea9d80189e2a6d6c43b119727047c4c4/tsfresh/feature_extraction/feature_calculators.py#L866

https://github.com/blue-yonder/tsfresh/blob/f8a952e4ea9d80189e2a6d6c43b119727047c4c4/tsfresh/feature_extraction/feature_calculators.py#L894

nils-braun:
Hi @emanuelef !
Thanks for the issue :-) You are right, there is something unclear there.
In my opinion, I think the descriptions match the code:

```
   ...
        len(different values occurring more than once) / len(different values)
    This means the percentage is normalized to the number of unique values,
    ...
np.sum(counts > 1) / float(counts.shape[0])
```

and

```
    ...
        # of data points occurring more than once / # of all data points
    This means the ratio is normalized to the number of data points in the time series
    ...
reoccuring_values / x.size
```

However, one could argue that the function names themselves are swapped! 
Would you like to do a pull request to change that?

emanuelef:
Sure, I guess also tests needs to be fixed then:
https://github.com/blue-yonder/tsfresh/blob/f8a952e4ea9d80189e2a6d6c43b119727047c4c4/tests/units/feature_extraction/test_feature_calculations.py#L518

I also wanted to ask why one has two decorators:
@set_property("fctype", "simple")
@set_property("input", "pd.Series")
but the other one misses "input"

Thanks

emanuelef:
Another question is about the name of the test for percentage_of_reoccurring_values_to_all_values
which is currently test_ratio_of_doubled_values

emanuelef:
Last question is about creating the PR, I cannot create a branch here, ok to create the PR from a fork ?

nils-braun:
Let me try to answer:
* Yes, forking is the usual way (not only for tsfresh, but basically for every project on GitHub)
* Feel free to change every naming that makes sense to you! We can discuss the details on the pull request :-)
* The input parameter controls the type conversation. I need to look up the details, but it looks like it is not needed. You can leave it as it is for now

nils-braun:
[...]

PR #725, opened 2020-06-27
swapped function names for percentage of reoccuring

emanuelef:
As per issue #724 (https://github.com/blue-yonder/tsfresh/issues/724).

percentage_of_reoccurring_datapoints_to_all_datapoints is actually percentage_of_reoccurring_values_to_all_values implementation and vice versa .

### DIFF
--- a/tsfresh/feature_extraction/feature_calculators.py
+++ b/tsfresh/feature_extraction/feature_calculators.py
@@ -871,7 +871,7 @@ def percentage_of_reoccurring_values_to_all_values(x):
         len(different values occurring more than once) / len(different values)
 
     This means the percentage is normalized to the number of unique values,
-    in contrast to the percentage_of_reoccurring_values_to_all_values.
+    in contrast to the percentage_of_reoccurring_datapoints_to_all_datapoints.
 
     :param x: the time series to calculate the feature of
     :type x: numpy.ndarray
@@ -899,7 +899,7 @@ def percentage_of_reoccurring_datapoints_to_all_datapoints(x):
         # of data points occurring more than once / # of all data points
 
     This means the ratio is normalized to the number of data points in the time series,
-    in contrast to the percentage_of_reoccurring_datapoints_to_all_datapoints.
+    in contrast to the percentage_of_reoccurring_values_to_all_values.
 
     :param x: the time series to calculate the feature of
     :type x: numpy.ndarray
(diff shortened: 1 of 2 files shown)


tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -891,7 +891,7 @@ def percentage_of_reoccurring_datapoints_to_all_datapoints(x):
 
 @set_property("fctype", "simple")
 @set_property("input", "pd.Series")
-def percentage_of_reoccurring_values_to_all_values(x):
+def percentage_of_reoccurring_datapoints_to_all_datapoints(x):
     """
     Returns the ratio of unique values, that are present in the time series
     more than once.
```
nils-braun, line 897: Could you also fix the comment further down (And also in the other function?). It still references the old function name

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -863,7 +863,7 @@ def first_location_of_minimum(x):
 
 
 @set_property("fctype", "simple")
-def percentage_of_reoccurring_datapoints_to_all_datapoints(x):
+def percentage_of_reoccurring_values_to_all_values(x):
     """
     Returns the percentage of unique values, that are present in the time series
```
MaxBenChrist, line 868: I think the function descriptions are misleading. I would change it to

```suggestion
    Returns the percentage of values in the time series, that are present 
```

tsfresh/feature_extraction/feature_calculators.py
MaxBenChrist, line 897:
```suggestion
    Returns the fraction of non-unique data points. Non-unique means that they are contained another time in the time series again.
```

MaxBenChrist:
While looking at this feature, I am not sure if it wouldn't make sense to add some rounding before you calculate the number of reocurring values to deal with floating point imprecision?

tsfresh/feature_extraction/feature_calculators.py
emanuelef: Missed that! Fixed now

emanuelef:
Made some changes according to suggestions, I used "percentage" for both.

emanuelef:
> would be great if you could polish the docstrings a little bit

Sure, what should I change ?

nils-braun:
I think it is fine as it is, thanks for fixing!

PR #733, opened 2020-07-18
Include comments from #722

nils-braun:
@hoesler I included your comments from the PR (which I, unfortunately, closed too early...)


hoesler:
Perfect, thanks.

PR #735, opened 2020-07-24
Remove slices again

nils-braun:
This PR is an addon to #707, which introduced the concept of data classes into tsfresh.
Especially this PR handles some performance issues I have seen.

PR #707 introduced the very nice concept of data classes, which I really love.
It makes so many things so much easier - thank you again for this addition @hoesler .

However, there was a very subtle bug in the implementation, which causes a larger
runtime and can even cause a larger memory consumption (instead of a smaller).
As I am not sure if I have overlooked something, let me try to walk you through my thoughts
so that you can tell me if I have forgotten something.

Putting the dict ts adapter aside, the main entity in the adapters is a `_Slice` object
and before the feature extraction the data is "partitioned" into these slices.
The idea is good: it is something like an iterator (storing where to start and where to end)
without actually doing the iteration - leaving the iteration up to the calling function
(e.g. in multiprocessing a different process or on dask even a different server).
We can think of a Slice as Slice = start, end, pointer to (grouped) data.

1. Problem: Runtime.
Once a slice reaches the worker, the worker needs to extract the chunk from the data.
Conceptional, it does something like this:

```python
    df.groupby([... ])[start, end]
```

However, the groupby in pandas is "clever" (for us it is actually the opposite): it won't be
calculated before it is called. So every worker needs to recompute the groupby for every
chunk again. Even worse, it then needs to traverse through all these groups until it has reached
the one with the correct number - which means it has to iterate e.g. through the first few group
for all the chunks, for the next few groups for all but the first chunk etc.

2. Problem: Memory.
The problem with this idea is, how python transports data from the start process to the workers
(at least for multiprocessing, but this is our main use case). It pickles the data.
If you want, you can have a look into this example to see, when the data is pickled:

```python
from multiprocessing import Pool
import os

# A small helper class that prints when it is (un)pickled
class PickleMe:
    def __init__(self, number):
        self.number = number

    def __getstate__(self, *args, **kwargs):
        print(self.number, "is pickled on", os.getpid())
        return self.__dict__

    def __setstate__(self, d):
        self.__dict__ = d
        print(self.number, "is unpickled on", os.getpid())


large_list = [PickleMe(i) for i in range(10)]


with Pool(4) as p:
    p.map(print, large_list)
```

The problem is, that the _Slice object also contains a pointer to the full data sample
(as it has not already extracted the data). Pickle is "stupid" - it will just pickle everything it has.
Including the data, which needs to be copied multiple times for every chunk and also be kept in memory.

You can see the influence in this code snippet:

```python
# use for example the create_data from the benchmark test
df = create_data(50, 5000)
ts = to_tsdata(df, column_id="id", column_sort="time", column_value="value")

i = ts.partition(100)

import pickle
print(len(pickle.dumps(next(i))))
```

The slice object has a size of 12MB, which is even larger than the full data (8MB, because pickling a grouped-by pandas object seems to be more complicated than the df itself). I am not sure if the OS or python is smart enough to see, that it at least only needs to keep a single pickled version of the grouped dataframe - but I actually doubt it.
So we might end up with N+1 copies of the data for N workers.

Solution:
How can we solve that? In this PR I reverted some changes back to what we had - basically iterating through the
groupby and taking the next n elements, wrapping them into a list, and pushing them to the workers.
The size per pickled object is then 159 kB (for the same example as above), so the memory consumption should be small again.
Here are some very preliminary runtime tests, using a large random data sample (100 columns, length 1000, 500 ids)
using only MinimalFCParameters (I did not want to wait that long...)

slice: 19s (current main)
no slice: 11s (this PR)

Drawbacks:
Nothing comes without drawbacks. In the current implementation using slices the groupby is repeated multiple times
as I mentioned above, however, it is calculated on the workers (instead of the main process).
This is a benefit if you have a very fast calculation (e.g. very little samples and MinimalFCParameters)
because then the workers do not need to wait for the main process.
For example in 100 columns, length of 10, 100 ids and still MinimalFCParameters, one gets

slice: 1.29s
no slice: 1.81s

However, I think that is quite an unusual use-case.

Writing this text I do remember, that 3 years ago when we started the project we did face similar problems
with multiple iterations and calls to groupby. Back then, I think we did not really understand it.

I am sorry for not having mentioned that on your PR already back then, @hoesler.
It took some time for me to fully understand what is going on.
Do not feel discouraged - I still really like the ideas in your PR. This just builds on them.
I hope I covered everything in this PR - please help me find any open points.

Remaining question is, why was there still an improvement seen in tests by users?
Because the melding is and was not needed anymore. That is the real nice optimization implemented in the PR
and I guess this took so much time for many use-cases.

PS: I also sneaked in a small optimization to sort only after grouping as we discussed on the last PR. I did not see any improvements on my benchmarks, but it just *feels* better (and also gives no degregations)


nils-braun:
> # [Codecov](https://codecov.io/gh/blue-yonder/tsfresh/pull/735?src=pr&el=h1) Report
> > Merging [#735](https://codecov.io/gh/blue-yonder/tsfresh/pull/735?src=pr&el=desc) into [main](https://codecov.io/gh/blue-yonder/tsfresh/commit/c23be2b89490dced6afb80bd6a88a4f48f780792&el=desc) will **increase** coverage by `8.63%`.
> > The diff coverage is `93.75%`.
> 
> [Impacted file tree graph](https://codecov.io/gh/blue-yonder/tsfresh/pull/735?src=pr&el=tree)
> 
> ```diff
> @@            Coverage Diff             @@
> ##             main     #735      +/-   ##
> ==========================================
> + Coverage   87.61%   96.25%   +8.63%     
> ==========================================
>   Files          29       18      -11     
>   Lines        1922     1656     -266     
>   Branches      348      317      -31     
> ==========================================
> - Hits         1684     1594      -90     
> + Misses        199       30     -169     
> + Partials       39       32       -7     
> ```
> 
> [Impacted Files](https://codecov.io/gh/blue-yonder/tsfresh/pull/735?src=pr&el=tree)	Coverage Δ	
> [tsfresh/utilities/distribution.py](https://codecov.io/gh/blue-yonder/tsfresh/pull/735/diff?src=pr&el=tree#diff-dHNmcmVzaC91dGlsaXRpZXMvZGlzdHJpYnV0aW9uLnB5)	`91.91% <66.66%> (-1.42%)`	
> [tsfresh/feature_extraction/data.py](https://codecov.io/gh/blue-yonder/tsfresh/pull/735/diff?src=pr&el=tree#diff-dHNmcmVzaC9mZWF0dXJlX2V4dHJhY3Rpb24vZGF0YS5weQ==)	`96.03% <100.00%> (+0.55%)`	
> [tsfresh/feature_extraction/feature_calculators.py](https://codecov.io/gh/blue-yonder/tsfresh/pull/735/diff?src=pr&el=tree#diff-dHNmcmVzaC9mZWF0dXJlX2V4dHJhY3Rpb24vZmVhdHVyZV9jYWxjdWxhdG9ycy5weQ==)	`97.31% <0.00%> (-0.02%)`	
> [tsfresh/examples/__init__.py](https://codecov.io/gh/blue-yonder/tsfresh/pull/735/diff?src=pr&el=tree#diff-dHNmcmVzaC9leGFtcGxlcy9fX2luaXRfXy5weQ==)		
> [tsfresh/transformers/__init__.py](https://codecov.io/gh/blue-yonder/tsfresh/pull/735/diff?src=pr&el=tree#diff-dHNmcmVzaC90cmFuc2Zvcm1lcnMvX19pbml0X18ucHk=)		
> [tsfresh/examples/driftbif_simulation.py](https://codecov.io/gh/blue-yonder/tsfresh/pull/735/diff?src=pr&el=tree#diff-dHNmcmVzaC9leGFtcGxlcy9kcmlmdGJpZl9zaW11bGF0aW9uLnB5)		
> [tsfresh/utilities/profiling.py](https://codecov.io/gh/blue-yonder/tsfresh/pull/735/diff?src=pr&el=tree#diff-dHNmcmVzaC91dGlsaXRpZXMvcHJvZmlsaW5nLnB5)		
> [tsfresh/scripts/measure_execution_time.py](https://codecov.io/gh/blue-yonder/tsfresh/pull/735/diff?src=pr&el=tree#diff-dHNmcmVzaC9zY3JpcHRzL21lYXN1cmVfZXhlY3V0aW9uX3RpbWUucHk=)		
> [tsfresh/feature_selection/__init__.py](https://codecov.io/gh/blue-yonder/tsfresh/pull/735/diff?src=pr&el=tree#diff-dHNmcmVzaC9mZWF0dXJlX3NlbGVjdGlvbi9fX2luaXRfXy5weQ==)		
> [tsfresh/examples/har_dataset.py](https://codecov.io/gh/blue-yonder/tsfresh/pull/735/diff?src=pr&el=tree#diff-dHNmcmVzaC9leGFtcGxlcy9oYXJfZGF0YXNldC5weQ==)		
> ... and [5 more](https://codecov.io/gh/blue-yonder/tsfresh/pull/735/diff?src=pr&el=tree-more)		
> [Continue to review full report at Codecov](https://codecov.io/gh/blue-yonder/tsfresh/pull/735?src=pr&el=continue).
> 
> > **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> > `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> > Powered by [Codecov](https://codecov.io/gh/blue-yonder/tsfresh/pull/735?src=pr&el=footer). Last update [c23be2b...f1634f8](https://codecov.io/gh/blue-yonder/tsfresh/pull/735?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

There seems to be a problem on main - our coverage is not so low...

hoesler:
Wow, you really spent some time with this. Great work! And no worries, I am glad to have contributed a basis for further improvements. Pandas is still a mystery to me sometimes, when it comes lazy operations.
Without thinking about your findings too much, I wonder if you also took different partition sizes into account while comparing different implementation. As I mentioned in #707, the current defaults are probably not optimal any more. For my use case, the _Slice version appeared to be the overall fastest implementation when using a partition size equal to the number of available processors.

nils-braun:
Thanks for your very quick response!
I did not test with different chunksizes so far - good point. I just did a very quick test with the same sample as mentioned above and ended up with 23s with your PR #707 - however with a very large stddev between different runs. 
I assume that due to the much larger size of chunks that needed to be kept in memory I already started to run short on memory (I did this on my Laptop, 2 cores, 8 GB or RAM) or might even have swapped for a very short amount of time.

I might have a reason why it is faster for you to use chunksize = len / workers. In this case, you will have every worker doing exactly a single run over its data (whereas  len / (5 * workers) would mean we have every worker running 5 times over the data). As mentioned above the groupby is actually duplicated work, so having chunksize = len / workers minimized the amount of this duplication. But this is just a very first guess.
If you - by any chance - have the data for your tests back then fro #707 still available I would be very interested in the results of this PR :-) 
Do you think you can somehow find a reason why the implementation in #707 should be faster? Because from the thoughts I described above, I would actually argue that the implementation in this PR should be faster in general. But multiprocessing is strange - so I might be 100% wrong...

nils-braun:
@hoesler I will merge this PR now, as it is already open for quite some time.
If you still have any thoughts on this, I would be more than happy to here them - also after the merge!

PR #736, opened 2020-07-31
Dask Integration

nils-braun:
This PR adds an additional input format for tsfresh - namely dask dataframes.
Since very early days, it was possible to scale the computation with dask - but still keep a pandas dataframe.
Then we added dask bindings to help adding tsfresh to a dask DAG.
With this PR we finally allow users to just pass in a dask dataframe like a normal pandas dataframe to the `extract_features` function. 

I added a new distributor for this, which really just does nothing as all the heavy distribution work is handled by dask itself (we do not need to do anything). Also, I moved the pivoting into the data classes and made it optional (as it is often unwanted for dask).

I also improved the documentation on these things a lot and added guidance on when to use what.


docs/text/data_formats.rst
```diff
@@ -10,14 +10,15 @@ tsfresh offers three different options to specify the time series data to be use
 Irrespective of the input format, tsfresh will always return the calculated features in the same output format
 described below.
 
-All three input format options consist of :class:`pandas.DataFrame` objects. There are four important column types that
+Typicalle, the input format options consist of :class:`pandas.DataFrame` objects.
```
nils-braun, line 13:
```suggestion
Typically, the input format options consist of :class:`pandas.DataFrame` objects.
```

Issue #508, opened 2019-03-07
Dependent versus independent features for the Benjamini Yekutieli procedure.


rhysimu:
I've been trying to get my head around the Benjamini Hochberg procedure used in tsfresh. As I interpret the paper, variable 'C' in feature_selection/benjamini_hochberg_test.py should be a scalar value the same for every index as the sum for i=(1 to m) of 1/i. However, the way it is implemented in the code is as a cumulative sum array, with its value different for every i. 

The paper which is referenced (Benjamini and Yekutieli, 2001) seems a bit confusing in this regard as the index i is used in formula (1) and in the definition of Theorem 3.1. However, when I look at Wikipedia (forgive me - https://en.wikipedia.org/wiki/False_discovery_rate), they use k for the index of the feature and i for the index of the sum, which suggests that the resultant C would be a scalar rather than an array.

Are you able to provide any insight on which of the two is correct, as I get a much greater number of relevant features with C as a cumulative sum array than I do with C as a scalar?

MaxBenChrist:
There are two procedures, depending if the hypotheses (and with it the features) can be assumed to be statistically independent.

You use the cumulative sum if the features are dependent, which they are in most of the cases.

The feature selection in tsfresh will only kick out features that the procedures is sure are irrelevant.

> Are you able to provide any insight on which of the two is correct, as I get a much greater number of relevant features with C as a cumulative sum array than I do with C as a scalar?

You use C as a scaler if the features are thought to be independent. Otherwise use the cumulative sum.

rhysimu:
If the features are independent then C is 1.

If the features are dependent, then, the way I read the procedure, C is solely dependent on m (the number of features) and doesn't apply differently along the array of features. 

For example, if there are 10 features, then the sum for i=1 to m of 1/i is ~2.93. I believe the array C should be [2.93, 2.93, ....., 2.93, 2.93] rather than [1, 1.5, ....., 2.83, 2.93] when determining the max p value for relevance.

Have I misread the paper? or misunderstood the implementation in tsfresh?

MaxBenChrist:
Now I see what you are referring to. This looks like a bug.

Do you want to submit a pr for this?

> Are you able to provide any insight on which of the two is correct, as I get a much greater number of relevant features with C as a cumulative sum array than I do with C as a scalar?

I think you should get more relevant features with C as scalar than using the (wrong) cumulative sum. So, after the fix there should be more relevant features. 

[...]

PR #760, opened 2020-09-27
Fix rejection line calculation in Benjamini Yekutieli example

alxfmpl:
Running the Benjamini Yekutieli visualisation example, I noticed that the BY rejection line calculation does not yield the expected result. This PR aligns the BY rejection line calculation used in this example with the statsmodels implementation.

The problem is similar to #508.


notebooks/advanced/visualize-benjamini-yekutieli-procedure.ipynb
```diff
@@ -708,7 +317,7 @@
     "if HYPOTHESES_INDEPENDENT:\n",
     "    C = [1] * m\n",
     "else:\n",
-    "    C = [sum([1.0 / i for i in range(1, k + 1)]) for k in K]\n",
+    "    C = [sum([1.0 / k for k in K])] * m\n",
```
nils-braun, line 320: @MaxBenChrist Do you think you will have some time to have a look into this?

nils-braun (commented):
@alxfmpl Thank you very much for your contribution! Very happy to see this.
I have asked @MaxBenChrist to maybe review this PR (as he was, if I remember correctly, the original author if this notebook).

nils-braun:
You can ignore the failing benchmark test

notebooks/advanced/visualize-benjamini-yekutieli-procedure.ipynb
MaxBenChrist: Yes, will try to free some time the next days.

nils-braun: @MaxBenChrist Gentle reminder :-)

alxfmpl:
Upon review I noticed my problem/fix description was beating around the bush a little 😉. Basically this PR boils down to this:

With `C = [sum([1.0 / i for i in range(1, k + 1)]) for k in K]` the rejection line in the example looks something like this:

grafik

This is not really what I expected. So after consulting [Wikipedia about the BY procedure](https://en.wikipedia.org/wiki/False_discovery_rate#Benjamini%E2%80%93Yekutieli_procedure), the rather [confusing paper introducing the BY procedure](https://projecteuclid.org/download/pdf_1/euclid.aos/1013699998) and the statsmodels implementation, I figured out that with `C = [sum([1.0 / k for k in K])] * m` the rejection line looks a lot more like what I was expecting:

grafik

Hope this helps.

kempa-liehr:
Thanks, for the clarification :-)

nils-braun:
@kempa-liehr You assigned yourself to this. Are you fine with the changes? Shall we merge?

MaxBenChrist:
Better late than never: The changes are correct and we should merge them. 

some additional thoughts
- I would put the calculation of the rejection line inside the notebook into a function
- you stripped the cells of their output, do we want to keep the notebooks with or without output in the repository? the other notebooks in the tsfresh example folder have their output attached

MaxBenChrist (approved):

alxfmpl:
Hi @MaxBenChrist, thanks for your approval.

Regarding your thoughts:

1.  I'm more than happy to create a function for rejection line calculation if wished.
2. According to the [contributing guidlines](https://github.com/blue-yonder/tsfresh/blob/main/docs/text/how_to_contribute.rst#guidelines) notebook outputs should be removed. This obviously makes the diff really large. I think keeping the old outputs would be a bit misleading through, especially since notebooks including their outputs are rendered on GitHub.

nils-braun:
@alxfmpl Thank you very much for taking the time and implementing the fix and I am very sorry it took so long on our side...

alxfmpl:
@nils-braun No worries, thanks for the merge 👍

PR #763, opened 2020-10-06
Ts forecast example improvement

filipj8:
Adaboost:
image

LinearRegression:
image


nils-braun:
Nice! Thanks for the improvement! Looks very reasonable to me

PR #770, opened 2020-11-01
Numba

thibaultbl:
What do you think about this approach ?

I managed to improve speed, but we loose some code readibility.

I need to give you a more detailled benchmarck, with multiple data size.

Results of benchmarck:

image

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -59,7 +59,7 @@ packages =
 addopts =
     --cov tsfresh --cov-report term-missing
     --verbose
-    -n auto
+    -n 0
 
 [aliases]
 docs = build_sphinx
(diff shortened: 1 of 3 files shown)


nils-braun:
@thibaultbl 
You closed the PR immediately after opening. Was this intentionally? Shall I have a look anyway?
I would be super happy to see numba at work in our repo :-)
One thing to keep in mind is if it plays well with multiprocessing (this used to be something that did make it a lot slower with numba in the past)

thibaultbl:
@nils-braun 
Yes sorry, I think now that I need to check some additional elements before submitting a proper PR.

Indeed, I am not sure how it works will multiprocessing (especially because I used numba.prange in my code), I need to check that as well.

PR #778, opened 2020-12-03
Fix column names in quick_start.rst

hsahovic:


hsahovic:
There seems to be some confusion around `id` 20 vs. 21 for the failure case - it might be worth it to double check what works with the actual dataset.

nils-braun:
Thanks for the change. Did I understand you correctly, did you check the actual dataset (for the id) or do you think it should be checked?

hsahovic:
I have not checked the dataset.
The graph and description both refer to `id` 20, but the code snippet refers to `id` 21.

nils-braun:
Just to comment: I have not forgotten about this PR and sorry for the delay. I hope I can have a look on the weekend

nils-braun:
Ok, 20 is fine - I checked with the data and the picture. Nice catch!

PR #779, opened 2020-12-05
Update relevant_feature_augmenter.py

yairst:
Fix bug in fit_transform() where filter_only_tsfresh_features is True and X is not empty. In this case, X_augmented does not include X.columns (inspect _fit_and_augment() output) resulting in pandas error: 'Passing list-likes to .loc or [] with any missing labels is no longer supported'. The changes in this commit fix this bug without hurting the other cases.
Also, change the description of _fit_and_augment() so that it matches the returned value, X_augmented

### DIFF
--- a/tests/units/transformers/test_relevant_feature_augmenter.py
+++ b/tests/units/transformers/test_relevant_feature_augmenter.py
@@ -81,7 +81,7 @@ class RelevantFeatureAugmenterTestCase(DataTestCase):
         self.assertEqual(sum(["pre_feature" == column for column in transformed_X.columns]), 1)
         self.assertEqual(sum(["pre_feature" == column for column in fit_transformed_X.columns]), 1)
 
-    def filter_only_tsfresh_features_false(self):
+    def test_filter_only_tsfresh_features_false(self):
         """
         The boolean flag `filter_only_tsfresh_features` makes sure that only the time series based features are
         filtered. This unit tests checks that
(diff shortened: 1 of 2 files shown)


tests/units/transformers/test_relevant_feature_augmenter.py
```diff
@@ -72,11 +76,14 @@ def test_evaluate_only_added_features_true(self):
         augmenter.fit(X, y)
         transformed_X = augmenter.transform(X.copy())
 
+        fit_transformed_X = augmenter.fit_transform(X, y)
+
         self.assertEqual(sum(["pre_feature" == column for column in transformed_X.columns]), 1)
+        self.assertEqual(sum(["pre_feature" == column for column in fit_transformed_X.columns]), 1)
 
-    def test_evaluate_only_added_features_false(self):
+    def filter_only_tsfresh_features_false(self):
```
nils-braun, line 84: Is there a reason you removed the `test_` prefix from this function? This means it will not be picked up by most of the unittest frameworks

nils-braun (commented):
Thank you very much for the fix! Looks reasonable to me except for the one small comment I have.

tests/units/transformers/test_relevant_feature_augmenter.py
yairst: deleted accidently. will be fixed

PR #787, opened 2021-01-07
First test implementation of a matrix_profile feature

### LINKED ISSUE #785: Integration of matrixprofile with tsfresh
Moving earlier conversations with @nils-braun , @tylerwmarrs and @MaxBenChrist to GitHub.

**Overview**
Our goal is to create a combiner that 1) calculates the matrix profile (with supplied parameter k) for each time series dimension, then 2) returns Tukey's 5 number summary for each dimension. Note that additional features can be easily added in the future as the matrix profile will already be calculated for the dataset.

**Technical Notes**
The integration will most likely use the matrixprofile.compute functionality. Parameters for compute are listed below - there are several values we'll need to align on, and whether these should be user-supplied or leverage sensible defaults (window size, sample pct, etc).

The package also needs to be hosted on conda to be a non-optional dependency. This workstream is being handled [here](https://github.com/matrix-profile-foundation/matrixprofile/issues/70).

_compute parameters_

        ts (array_like) – The time series to analyze.

        windows (int, array_like) – The window(s) to compute the MatrixProfile. Note that it may be an int for a single matrix profile computation or an array of ints for computing the pan matrix profile.

        query (array_like, optional) – The query to analyze. Note that when computing the PMP the query is ignored!

        sample_pct (float, default 1) – A float between 0 and 1 representing how many samples to compute for the MP or PMP. When it is 1, the exact algorithm is used.

        threshold (float, default 0.98) – The correlation coefficient used as the threshold. It should be between 0 and 1. This is used to compute the upper window size when no window(s) is given.

        n_jobs (int, default = 1) – Number of cpu cores to use.

nils-braun:
Fixes #785 

Sorry it took so long for the first draft...

This is the draft for including matrixprofile in tsfresh. See #785 for more information and still remaining issues.
@vanbenschoten and @tylerwmarrs, I have some questions (also mentioned in the code):

* Currently I calculate using a single call to `compute` the following features:
```
F_x__matrix_profile__feature_"min"__sample_pct_...  0.000000
F_x__matrix_profile__feature_"max"__sample_pct_...  3.255922
F_x__matrix_profile__feature_"mean"__sample_pct...  2.003147
F_x__matrix_profile__feature_"median"__sample_p...  2.134084
F_x__matrix_profile__feature_"25"__sample_pct_1...  1.683442
F_x__matrix_profile__feature_"75"__sample_pct_1...  2.447564
```
  Is this ok as a first draft?
* In principle a user would be able to change all settings of the call to `compute` via tsfreshs setting objects. Currently, I set sample_pct and threshold explicitely to the default parameters of mp, to make it more explicit. Does this make sense?
* I have removed all NaN and Inf from the matrix profile before calculating max min etc. Sounds reasonable?

Of course, we should add tests before merging :-)

### DIFF
--- a/tsfresh/feature_extraction/feature_calculators.py
+++ b/tsfresh/feature_extraction/feature_calculators.py
@@ -19,6 +19,7 @@ alphabetically ascending.
 
 import itertools
 import functools
+from tsfresh.utilities.string_manipulation import convert_to_output_format
 import warnings
 from builtins import range
 from collections import defaultdict
@@ -29,6 +30,7 @@ from numpy.linalg import LinAlgError
 from scipy.signal import cwt, find_peaks_cwt, ricker, welch
 from scipy.stats import linregress
 from statsmodels.tools.sm_exceptions import MissingDataError
+import matrixprofile as mp
 
 with warnings.catch_warnings():
     # Ignore warnings of the patsy package
@@ -2212,3 +2214,67 @@ def benford_correlation(x):
     # np.corrcoef outputs the normalized covariance (correlation) between benford_distribution and data_distribution.
     # In this case returns a 2x2 matrix, the  [0, 1] and [1, 1] are the values between the two arrays
     return np.corrcoef(benford_distribution, data_distribution)[0, 1]
+
+
+@set_property("fctype", "combiner")
+def matrix_profile(x, param):
+    """
+    TODO: Documentation
+
+    :param x: the time series to calculate the feature of
+    :type x: numpy.ndarray
+    :param param: contains dictionaries {"sample_pct": x, "threshold": y, "feature": z}
+                  with sample_pct and threshold being parameters of the matrixprofile
+                  package https://matrixprofile.docs.matrixprofile.org/api.html#matrixprofile-compute
+                  and feature being one of "min", "max", "mean", "median", "25", "75"
+                  and decides which feature of the matrix profile to extract
+    :type param: list
+    :return: the different feature values
+    :return type: pandas.Series
+    """
+    x = np.asarray(x)
+
+    def _calculate_pmp(**kwargs):
+        """Calculate the pan-matrix profile"""
+        try:
+            pmp = mp.compute(x, **kwargs)["pmp"]
+            # TODO: reasonable?
+            return pmp[(~np.isnan(pmp)) & (~np.isinf(pmp))]
+        except Exception:
+            return [np.NaN]
+
+    # The already calculated matrix profiles
+    matrix_profiles = {}
+
+    # The results
+    res = {}
+
+    for kwargs in param:
+        key = convert_to_output_format(kwargs)
+        feature = kwargs.pop('feature')
+
+        # Only calculate the pmp if we have not already done so
+        # The feature calculation can happen afterwards
+        featureless_key = convert_to_output_format(kwargs)
+        if featureless_key not in matrix_profiles:
+            matrix_profiles[featureless_key] = _calculate_pmp(**kwargs)
+
+        pmp = matrix_profiles[featureless_key]
+
+        # TODO: is this what we want?
+        if feature == "min":
+            res[key] = np.min(pmp)
+        elif feature == "max":
+            res[key] = np.max(pmp)
+        elif feature == "mean":
+            res[key] = np.mean(pmp)
+        elif feature == "median":
+            res[key] = np.median(pmp)
+        elif feature == "25":
+            res[key] = np.percentile(pmp, 25)
+        elif feature == "75":
+            res[key] = np.percentile(pmp, 75)
+        else:
+            raise ValueError(f"Unknown feature {feature} for the matrix profile")
+
+    return [(key, value) for key, value in res.items()]
--- a/tsfresh/feature_extraction/settings.py
+++ b/tsfresh/feature_extraction/settings.py
@@ -152,7 +152,9 @@ class ComprehensiveFCParameters(dict):
             "lempel_ziv_complexity": [{"bins": x} for x in [2, 3, 5, 10, 100]],
             "fourier_entropy":  [{"bins": x} for x in [2, 3, 5, 10, 100]],
             "permutation_entropy":  [{"tau": 1, "dimension": x} for x in [3, 4, 5, 6, 7]],
-
+            # TODO: also the window parameter?
+            "matrix_profile": [{"sample_pct": 1, "threshold": 0.98, "feature": f}
+                               for f in ["min", "max", "mean", "median", "25", "75"]]
         })
 
         super().__init__(name_to_param)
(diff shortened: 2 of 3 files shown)


tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -2212,3 +2214,67 @@ def benford_correlation(x):
...
+
+
+@set_property("fctype", "combiner")
+def matrix_profile(x, param):
+    """
+    TODO: Documentation
+
+    :param x: the time series to calculate the feature of
+    :type x: numpy.ndarray
+    :param param: contains dictionaries {"sample_pct": x, "threshold": y, "feature": z}
```
tylerwmarrs, line 2226: Computing the PMP by default can be a bit expensive in some cases as many MatrixProfiles are computed. Are you OK with having a window parameter instead of a threshold here? Another approach could be to always default to the maximum window size given a user threshold to determine what MatrixProfile window to compute:

https://matrixprofile.docs.matrixprofile.org/api.html#matrixprofile-algorithms-maximum-subsequence

To determine the maximum "subsequence" (window size), we still have to compute MatrixProfiles to determine what makes sense. It will probably make a little more sense if you take a look at that code. The algorithm starts with a very wide search and refines the window size later on.

vanbenschoten: I'd recommend the default behavior (that is, no user-defined window is passed) be computation of the maximum subsequence with the current 0.95 correlation threshold. I realize that the threshold can be very domain-specific, but this seems like the best compromise for an "automated" solution that many users will prefer to leverage.

@nils-braun to your point, more advanced users can specify window size(s) by passing variables directly to `compute`

tylerwmarrs: The other item that could be confusing for users is that the sample_pct parameter uses the "anytime" algorithm (approximate solution) under the covers. This may result in different answers, however a faster solution in some cases.

vanbenschoten:
@nils-braun I'll be adding some updated code to this PR later tonight.

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -2212,3 +2214,67 @@ def benford_correlation(x):
...
+                  package https://matrixprofile.docs.matrixprofile.org/api.html#matrixprofile-compute
+                  and feature being one of "min", "max", "mean", "median", "25", "75"
+                  and decides which feature of the matrix profile to extract
+    :type param: list
+    :return: the different feature values
+    :return type: pandas.Series
+    """
+    x = np.asarray(x)
+
+    def _calculate_pmp(**kwargs):
```
vanbenschoten, line 2237: Based on the conversations we've been having, we'll use the matrix profile with the maximum subsequence if no window is specified in kwargs. Otherwise, we'll directly calculate the mp with the user specified window value.

```
if 'windows' in kwargs.keys():
    m_p = mp.compute(x,**kwargs)

else:
    m_p = mp.algorithms.maximum_subsequence(x, include_pmp=True)['pmp'][-1]
```
`

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -2212,3 +2214,67 @@ def benford_correlation(x):
...
+
+        # Only calculate the pmp if we have not already done so
+        # The feature calculation can happen afterwards
+        featureless_key = convert_to_output_format(kwargs)
+        if featureless_key not in matrix_profiles:
+            matrix_profiles[featureless_key] = _calculate_pmp(**kwargs)
+
+        pmp = matrix_profiles[featureless_key]
+
+        # TODO: is this what we want?
```
vanbenschoten, line 2264: Another way we could calculate Tukey's Five Numbers is as follows (h/t to @tylerwmarrs):

```
  if data.size < 1:
      return [np.nan] * 5
  
  x = data[~np.isnan(data)]
  x = np.sort(x)
  n = len(x)
  n4 = np.floor((n + 3) / 2) / 2
  d = np.array([1, n4, (n + 1) / 2, n + 1 - n4, n])
  i1 = (np.floor(d) - 1).astype(int)
  i2 = (np.ceil(d) - 1).astype(int)
  
  return 0.5 * (x[i1] + x[i2])
```

The above way may not actually be faster/easier given that Numpy has already optimized the percentile count, so I'm fine if we leave as-is for now.

tsfresh/feature_extraction/settings.py
```diff
@@ -152,7 +152,9 @@ def __init__(self):
             "lempel_ziv_complexity": [{"bins": x} for x in [2, 3, 5, 10, 100]],
             "fourier_entropy":  [{"bins": x} for x in [2, 3, 5, 10, 100]],
             "permutation_entropy":  [{"tau": 1, "dimension": x} for x in [3, 4, 5, 6, 7]],
-
+            # TODO: also the window parameter?
```
vanbenschoten, line 155: I might actually leave the window parameter out, as that's something the power users can add in on their own.

vanbenschoten (commented):
@nils-braun I'm not sure how you prefer to logistically collaborate on PRs, so let me know if you'd rather use something else than the review functionality :)

tsfresh/feature_extraction/settings.py
vanbenschoten: We could also do the same with "threshold" assuming we go the max subsequence route

tsfresh/feature_extraction/feature_calculators.py
tylerwmarrs: `'windows' in kwargs `means the same thing and is more pythonic

nils-braun:
That is definitely ok, from the comments you two were giving however I had the feeling you two were still discussing the best way to go - but with the new comments you gave I think I might know what to do.

If you like, always feel free to just do the code changes which might be a lot easier than trying to phrase it for me, who has no idea on how matrixprofile works :-)

As tsfresh is a side-project for me (I can only work on it in my free time and not during my working hours), my reaction time can sometimes be a bit long, sorry for that.

vanbenschoten:
@nils-braun no worries! We're in the same boat when it comes to availability, so not a problem at all :) I'll push my recommended code changes to the branch later today - @tylerwmarrs are you good with what I've proposed above, or is there other code I should add to that PR?

tsfresh/feature_extraction/feature_calculators.py
nils-braun: So as I do not think that this part will be the performance bottleneck and the other code (at least in my opinion) is easier to read, I would go with the already present one (as you have mentioned)

nils-braun:
I do not think you can push to exactly this branch, but feel free to fork and open a PR from your repo. That is totally fine and we can then close this one :-)

tylerwmarrs:
> @nils-braun no worries! We're in the same boat when it comes to availability, so not a problem at all :) I'll push my recommended code changes to the branch later today - @tylerwmarrs are you good with what I've proposed above, or is there other code I should add to that PR?

Go ahead with your PR. @nils-braun and I can do code reviews there.

nils-braun:
Closed in favor of #793

PR #793, opened 2021-01-15
Addition of matrix_profile feature

vanbenschoten:
@nils-braun @tylerwmarrs let me know what you think!

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,3 +8,4 @@ scikit-learn>=0.19.2
 tqdm>=4.10.0
 dask[dataframe]>=2.9.0
 distributed>=2.11.0
+matrixprofile>=1.1.6
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,4 +8,4 @@ scikit-learn>=0.19.2
 tqdm>=4.10.0
 dask[dataframe]>=2.9.0
 distributed>=2.11.0
-matrixprofile>=1.1.7<2.0.0
+matrixprofile>=1.1.10<2.0.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,4 +8,4 @@ scikit-learn>=0.19.2
 tqdm>=4.10.0
 dask[dataframe]>=2.9.0
 distributed>=2.11.0
-matrixprofile>=1.1.6
+matrixprofile>=1.1.7<2.0.0
--- a/tests/units/feature_extraction/test_feature_calculations.py
+++ b/tests/units/feature_extraction/test_feature_calculations.py
@@ -1303,7 +1303,7 @@ class FeatureCalculationTestCase(TestCase):
         self.assertIsNaN(benford_correlation(equal_list))
 
     def test_matrix_profile_window(self):
-        #Test matrix profile output with specified window
+        # Test matrix profile output with specified window
         np.random.seed(9999)
         ts = np.random.uniform(size=2**10)
         w = 2**5
@@ -1311,19 +1311,18 @@ class FeatureCalculationTestCase(TestCase):
         ts[0:w] = subq
         ts[w+100:w+100+w] = subq
         param = [
-        {"threshold": 0.98, "windows": 36, "feature": "min"},
-        {"threshold": 0.98, "windows": 36, "feature": "max"},
-        {"threshold": 0.98, "windows": 36, "feature": "mean"},
-        {"threshold": 0.98, "windows": 36, "feature": "median"},
-        {"threshold": 0.98, "windows": 36, "feature": "25"},
-        {"threshold": 0.98, "windows": 36, "feature": "75"}
+            {"threshold": 0.98, "windows": 36, "feature": "min"},
+            {"threshold": 0.98, "windows": 36, "feature": "max"},
+            {"threshold": 0.98, "windows": 36, "feature": "mean"},
+            {"threshold": 0.98, "windows": 36, "feature": "median"},
+            {"threshold": 0.98, "windows": 36, "feature": "25"},
+            {"threshold": 0.98, "windows": 36, "feature": "75"}
         ]
 
-        self.assertAlmostEqual(matrix_profile(ts,param=param)[0][1],2.825786727580335)
-
+        self.assertAlmostEqual(matrix_profile(ts, param=param)[0][1], 2.825786727580335)
 
     def test_matrix_profile_no_window(self):
-        #Test matrix profile output with no window specified
+        # Test matrix profile output with no window specified
         np.random.seed(9999)
         ts = np.random.uniform(size=2**10)
         w = 2**5
@@ -1332,30 +1331,29 @@ class FeatureCalculationTestCase(TestCase):
         ts[w+100:w+100+w] = subq
 
         param = [
-        {"threshold": 0.98, "feature": "min"},
-        {"threshold": 0.98, "feature": "max"},
-        {"threshold": 0.98, "feature": "mean"},
-        {"threshold": 0.98, "feature": "median"},
-        {"threshold": 0.98, "feature": "25"},
-        {"threshold": 0.98, "feature": "75"}
+            {"threshold": 0.98, "feature": "min"},
+            {"threshold": 0.98, "feature": "max"},
+            {"threshold": 0.98, "feature": "mean"},
+            {"threshold": 0.98, "feature": "median"},
+            {"threshold": 0.98, "feature": "25"},
+            {"threshold": 0.98, "feature": "75"}
         ]
 
-        #Test matrix profile output with no window specified
-        self.assertAlmostEqual(matrix_profile(ts, param=param)[0][1],2.825786727580335)
-
+        # Test matrix profile output with no window specified
+        self.assertAlmostEqual(matrix_profile(ts, param=param)[0][1], 2.825786727580335)
 
     def test_matrix_profile_nan(self):
-        #Test matrix profile of NaNs (NaN output)
+        # Test matrix profile of NaNs (NaN output)
         ts = np.random.uniform(size=2**6)
         ts[:] = np.nan
 
         param = [
-        {"threshold": 0.98, "windows": None, "feature": "min"},
-        {"threshold": 0.98, "windows": None, "feature": "max"},
-        {"threshold": 0.98, "windows": None, "feature": "mean"},
-        {"threshold": 0.98, "windows": None, "feature": "median"},
-        {"threshold": 0.98, "windows": None, "feature": "25"},
-        {"threshold": 0.98, "windows": None, "feature": "75"}
+            {"threshold": 0.98, "windows": None, "feature": "min"},
+            {"threshold": 0.98, "windows": None, "feature": "max"},
+            {"threshold": 0.98, "windows": None, "feature": "mean"},
+            {"threshold": 0.98, "windows": None, "feature": "median"},
+            {"threshold": 0.98, "windows": None, "feature": "25"},
+            {"threshold": 0.98, "windows": None, "feature": "75"}
         ]
 
         self.assertTrue(np.isnan(matrix_profile(ts, param=param)[0][1]))
--- a/tests/units/feature_extraction/test_feature_calculations.py
+++ b/tests/units/feature_extraction/test_feature_calculations.py
@@ -1301,6 +1301,31 @@ class FeatureCalculationTestCase(TestCase):
         self.assertAlmostEqual(benford_correlation(list_with_nan), 0.10357511)
         self.assertIsNaN(benford_correlation(equal_list))
 
+    def test_matrix_profile_window(self):
+        #Test matrix profile output with specified window
+        np.random.seed(9999)
+        ts = np.random.uniform(size=2**10)
+        w = 2**5
+        subq = ts[0:w]
+        ts[0:w] = subq
+        ts[w+100:w+100+w] = subq
+        self.assertAlmostEqual(matrix_profile(ts,windows=36)[0],2.826)
+        self.assertAlmostEqual(matrix_profile(ts,windows=36)[1],3.514)
+        self.assertAlmostEqual(matrix_profile(ts,windows=36)[2],3.626)
+
+    def test_matrix_profile_no_window(self):
+        np.random.seed(9999)
+        ts = np.random.uniform(size=2**10)
+        w = 2**5
+        subq = ts[0:w]
+        ts[0:w] = subq
+        ts[w+100:w+100+w] = subq
+
+        #Test matrix profile output with no window specified
+        self.assertAlmostEqual(matrix_profile(ts)[0],2.826)
+        self.assertAlmostEqual(matrix_profile(ts)[1],3.514)
+        self.assertAlmostEqual(matrix_profile(ts)[2],3.626)
+
 
 class FriedrichTestCase(TestCase):
 
--- a/tests/units/feature_extraction/test_feature_calculations.py
+++ b/tests/units/feature_extraction/test_feature_calculations.py
@@ -5,6 +5,7 @@
 from random import shuffle
 from unittest import TestCase
 import warnings
+from matrixprofile.exceptions import NoSolutionPossible
 
 from tsfresh.feature_extraction.feature_calculators import *
 from tsfresh.feature_extraction.feature_calculators import _roll
--- a/tests/units/feature_extraction/test_feature_calculations.py
+++ b/tests/units/feature_extraction/test_feature_calculations.py
@@ -1326,6 +1326,12 @@ class FeatureCalculationTestCase(TestCase):
         self.assertAlmostEqual(matrix_profile(ts)[1],3.514)
         self.assertAlmostEqual(matrix_profile(ts)[2],3.626)
 
+    def test_matrix_profile_nan(self):
+        ts = np.random.uniform(size=2**6)
+        ts[:] = np.nan
+
+        self.assertRaises(NoSolutionPossible,matrix_profile(ts))
+
 
 class FriedrichTestCase(TestCase):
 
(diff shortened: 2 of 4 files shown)


tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -2212,3 +2214,69 @@ def benford_correlation(x):
...
+    x = np.asarray(x)
+
+    def _calculate_mp(**kwargs):
+                """Calculate the matrix profile using the specified window, or the maximum subsequence if no window is specified"""
+        try:
+            if 'windows' in kwargs:
+                m_p = mp.compute(x,**kwargs)
+
+            else:
+                m_p = mp.algorithms.maximum_subsequence(x, include_pmp=True)['pmp'][-1]
```
nils-braun, line 2243: In this case the `kwargs` are not used in the function call. Is this on purpose?

tsfresh/feature_extraction/settings.py
```diff
@@ -152,6 +152,8 @@ def __init__(self):
             "lempel_ziv_complexity": [{"bins": x} for x in [2, 3, 5, 10, 100]],
             "fourier_entropy":  [{"bins": x} for x in [2, 3, 5, 10, 100]],
             "permutation_entropy":  [{"tau": 1, "dimension": x} for x in [3, 4, 5, 6, 7]],
+                        "matrix_profile": [{"sample_pct": 1, "threshold": 0.98, "feature": f}
```
nils-braun, line 155: Currently, as `window` is not included, those kwargs are never used. Should we remove them?

nils-braun (commented):
Thanks!
Do you think you can also add some small tests for this feature?
And before we can merge, we would need the conda package (do not know how far you are already with that, just want to mention it)

requirements.txt
```diff
@@ -8,3 +8,4 @@ scikit-learn>=0.19.2
 tqdm>=4.10.0
 dask[dataframe]>=2.9.0
 distributed>=2.11.0
+matrixprofile>=1.1.6
```
tylerwmarrs, line 11: Based on our versioning and the requirement of 1.1.7, this should be:

`matrixprofile>=1.1.7<2.0.0`

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -2212,3 +2214,69 @@ def benford_correlation(x):
...
+            return m_p[(~np.isnan(m_p)) & (~np.isinf(m_p))]
```
tylerwmarrs, line 2244: I don't think you want to return a modified version of the matrix profile at this stage. What if additional features in the future handle imputation or something?

@nils-braun based on our conversation in the matrixprofile github issue, how do you want to handle any exception vs the specific "NoSolutionPossible" case?
```
    return m_p
except mp.exceptions.NoSolutionPossible as e:
    warnings.warn(str(e))
    return None
except Exception as e:
    # ?????????
    return None
```

tsfresh/feature_extraction/feature_calculators.py
tylerwmarrs: He should be using the threshold parameter here.

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -2212,3 +2214,69 @@ def benford_correlation(x):
...
+        feature = kwargs.pop('feature')
+
+        # Only calculate the pmp if we have not already done so
+        # The feature calculation can happen afterwards
+        featureless_key = convert_to_output_format(kwargs)
+        if featureless_key not in matrix_profiles:
+            matrix_profiles[featureless_key] = _calculate_mp(**kwargs)
+
+        m_p = matrix_profiles[featureless_key]
+
```
tylerwmarrs, line 2265: Here you can find the finite indices and store them for functions that do not work on non-finite data.
```
finite_indices = np.finite(m_p)
```

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -2212,3 +2214,69 @@ def benford_correlation(x):
...
+
+        if feature == "min":
+            res[key] = np.min(m_p)
```
tylerwmarrs, line 2268: Here you would use the finite indices and in additional places where it makes sense. This is not really a performance hit because numpy is highly optimized when working with memory views.

```
res[key] = np.min(m_p[finite_indices])
```

tsfresh/feature_extraction/feature_calculators.py
vanbenschoten: @tylerwmarrs I don't think threshold needs to be set here, as we planned to go with the default value of 0.95, and that's already set in the maximum_sequence function. That said, I'll re-insert kwargs for maximum flexibility.

tsfresh/feature_extraction/settings.py
vanbenschoten: @nils-braun per the earlier comment, I'll remove "sample_pct" and leave "threshold" for the time being.

requirements.txt
vanbenschoten: Good catch. Is there a reason we're specifying version < 2.0.0?

tsfresh/feature_extraction/feature_calculators.py
vanbenschoten: @tylerwmarrs this is a good callout. I like the idea of pulling out finite data later and leaving the full MP in case other features need it later.

requirements.txt
tylerwmarrs: We use semantic versioning like most packages. A version less than 2 should guarantee compatibility. See

 https://link.medium.com/rIKVIvX34cb

vanbenschoten:
@nils-braun tests are in there! Let me know if I should approach them differently.

tsfresh/feature_extraction/feature_calculators.py
nils-braun: Yes, I think a warning would make sense here! The question is, do we expect any other exception? If not, let's do not catch it and let it actually fail for the user

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -2212,3 +2214,70 @@ def benford_correlation(x):
     # np.corrcoef outputs the normalized covariance (correlation) between benford_distribution and data_distribution.
     # In this case returns a 2x2 matrix, the  [0, 1] and [1, 1] are the values between the two arrays
     return np.corrcoef(benford_distribution, data_distribution)[0, 1]
+
+
+@set_property("fctype", "combiner")
+def matrix_profile(x, param):
+    """
+    TODO: Documentation
```
nils-braun, line 2222: Just to mention, documentation is still missing :-)

tests/units/feature_extraction/test_feature_calculations.py
```diff
@@ -1301,6 +1301,31 @@ def test_benford_correlation(self):
...
+        w = 2**5
+        subq = ts[0:w]
+        ts[0:w] = subq
+        ts[w+100:w+100+w] = subq
+
+        #Test matrix profile output with no window specified
+        self.assertAlmostEqual(matrix_profile(ts)[0],2.826)
+        self.assertAlmostEqual(matrix_profile(ts)[1],3.514)
+        self.assertAlmostEqual(matrix_profile(ts)[2],3.626)
+
```
nils-braun, line 1328: Would you also like to add a test for the NaN (= failure) case?

vanbenschoten:
@nils-braun thanks for the feedback! I'll update the documentation and try adding the NaN test.

Would you happen to know why the checks above are failing? The error message isn't making sense to me (I checked the Matrix Profile code, and it runs just fine).

vanbenschoten:
Ok, all corrections have been made.

tylerwmarrs:
> @nils-braun thanks for the feedback! I'll update the documentation and try adding the NaN test.
> 
> Would you happen to know why the checks above are failing? The error message isn't making sense to me (I checked the Matrix Profile code, and it runs just fine).

I'll take a look later on. The error is basically saying that there is no valid indices from finite_indices variable. This could be that you are always returning [np.nan]. So something your exception is getting swallowed making it less obvious of what is really going on.

nils-braun:
Just to be sure (as @vanbenschoten said it is running fine): can you also reproduce the error locally? You can run the tests with `pytest` if you like

vanbenschoten:
@nils-braun to confirm, are you saying the code works when you run it?

nils-braun:
Ah sorry - no, it does also fail. I just thought it is working for you. 
Concerning the error, there are actually a few ones:
* `TypeError: only integer scalar arrays can be conver...`: this happens when you return the list `[np.NaN]`. You need to turn it into a `np.array` to use the `np.isfinite` function properly
* `NameError: name 'NoSolutionPossible' is not defined` your test misses an import :-)
* `TypeError: matrix_profile() missing 1 required positional argument: 'param'` I do not know what is going on here - that is probably related to your package. This is probably also true for `TypeError: matrix_profile() got an unexpected keyword argument 'windows'`

vanbenschoten:
Got it - thanks for the clarification. I had meant that the Matrix Profile
code worked locally, so these errors makes sense. Unless @tylerwmarrs can
push a quick fix I'll take a look in a bit.

tylerwmarrs:
> Ah sorry - no, it does also fail. I just thought it is working for you.
> Concerning the error, there are actually a few ones:
> 
> * `TypeError: only integer scalar arrays can be conver...`: this happens when you return the list `[np.NaN]`. You need to turn it into a `np.array` to use the `np.isfinite` function properly

@vanbenschoten I cannot work on this. You need to just wrap the return as Nils mentions
```
np.array([np.nan])
```

vanbenschoten:
@nils-braun Ok, all three Matrix Profile tests are passing (the failures are stemming from unrelated parts of the codebase)! What are next steps here?

nils-braun:
> @nils-braun Ok, all three Matrix Profile tests are passing (the failures are stemming from unrelated parts of the codebase)! What are next steps here?

These other failures are not unrelated :-)
They all boil down to the same problem:
You did just wrap everything with a big try catch block and return NaN on exception. While I would generally not recommend doing so (catching known and expected exceptions like the No Solution one is fine, but in the rest of the code an exception is unexpected and should ready be seen), you are also breaking the return type convention: the non-exception case returns a list of tuples - feature name to float. Now you are just returning a float. 
What I would recommend is what I have implemented at the very beginning: only do this while calculating the actual matrix profile and only catch those exception you expect.

vanbenschoten:
Ah, got it. Ok, I'll push some updated code later today.

vanbenschoten:
Just to make sure I'm understanding correctly, the expected feature return for the No Solution case should be:

[('feature_"min"__threshold_0.98', NaN),
 ('feature_"max"__threshold_0.98', NaN),
 ('feature_"mean"__threshold_0.98', NaN),
 ('feature_"median"__threshold_0.98', NaN),
 ('feature_"25"__threshold_0.98', NaN),
 ('feature_"75"__threshold_0.98', NaN)]

Sorry for missing this the first time!

vanbenschoten:
@nils-braun I've updated the code to return the "expected feature" listed above, but I'm not sure what's going on with the other errors listed in the pytest runs. It seems as though the "feature" key in the dictionary isn't being passed through - as an example, line 1460 in the Python 3.6 (lowest) shows:

param = [{'threshold': 0.98}, {'threshold': 0.98}, {'threshold': 0.98}, {'threshold': 0.98}, {'threshold': 0.98}, {'threshold': 0.98}]

This is odd, because "feature" is being explicitly set in settings.py. Any idea what might be taking place? If it's still my NaN return case let me know and I'll adjust :)

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -2212,3 +2214,84 @@ def benford_correlation(x):
...
+
+    # The already calculated matrix profiles
+    matrix_profiles = {}
+
+    # The results
+    res = {}
+
+    for kwargs in param:
+        key = convert_to_output_format(kwargs)
+        feature = kwargs.pop('feature')
```
nils-braun, line 2264: Sorry, that took some time for me to debug! And unfortunately I think it was me introducing the bug in one of my previous commits :-/
The problem is a bit complicated to describe, so here is the short version:
the parameters you are using here come from the settings object given to the `extract_features` function. Due to reference/pointer magic happening in python, the `kwargs` you are using here is actually the *exact* one stored in the settings object. If you now use these setings twice in the same test (which all of those failed tests do), you actually *remove* `features` from the original settings object and will not be present the next time :-)
So, simple fix: `add kwargs = kwargs.copy()` before that.

nils-braun (commented):
Sorry for the late reply, I am currently involved in a lot of different things :-/ But with the one additional line I propose below all your tests should work!

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -2212,3 +2214,84 @@ def benford_correlation(x):
...
+
+        featureless_key = convert_to_output_format(kwargs)
+        if featureless_key not in matrix_profiles:
+            matrix_profiles[featureless_key] = _calculate_mp(**kwargs)
+
+        m_p = matrix_profiles[featureless_key]
+
+        #Set all features to nan if Matrix Profile is nan (cannot be computed)
+        if len(m_p) == 1:
```
nils-braun, line 2273: Can the len be 1 also for "normal" cases? If that can ever happen, I would propose to make this one simpler:
do not return `[np.nan]` on errors, but actually `None` and here only check for `if m_p is None`- I also think that is more pythonic, but that might be a matter of taste

vanbenschoten: The good news is that the length cannot be 1 for normal cases, otherwise I'd definitely go with your approach.

tsfresh/feature_extraction/feature_calculators.py
vanbenschoten: No worries! I've updated the code to reflect this :)

vanbenschoten:
@nils-braun tests are passing now! Thanks much for your help. What are next steps here?

nils-braun:
There is just a minor style error - but once you did also resolve these, I will merge:
```
./tsfresh/feature_extraction/feature_calculators.py:2226:121: E501 line too long (151 > 120 characters) ./tsfresh/feature_extraction/feature_calculators.py:2243:121: E501 line too long (123 > 120 characters) ./tsfresh/feature_extraction/feature_calculators.py:2246:35: E231 missing whitespace after ',' ./tsfresh/feature_extraction/feature_calculators.py:2249:76: E231 missing whitespace after ',' ./tsfresh/feature_extraction/feature_calculators.py:2253:9: E722 do not use bare 'except' ./tsfresh/feature_extraction/feature_calculators.py:2263:5: E303 too many blank lines (2) ./tsfresh/feature_extraction/feature_calculators.py:2274:9: E265 block comment should start with '# ' ./tsfresh/feature_extraction/feature_calculators.py:2278:9: E265 block comment should start with '# ' ./tsfresh/feature_extraction/feature_calculators.py:2284:13: E303 too many blank lines (2) ./tests/units/feature_extraction/test_feature_calculations.py:1306:9: E265 block comment should start with '# ' ./tests/units/feature_extraction/test_feature_calculations.py:1314:9: E122 continuation line missing indentation or outdented ./tests/units/feature_extraction/test_feature_calculations.py:1315:9: E122 continuation line missing indentation or outdented ./tests/units/feature_extraction/test_feature_calculations.py:1316:9: E122 continuation line missing indentation or outdented ./tests/units/feature_extraction/test_feature_calculations.py:1317:9: E122 continuation line missing indentation or outdented ./tests/units/feature_extraction/test_feature_calculations.py:1318:9: E122 continuation line missing indentation or outdented ./tests/units/feature_extraction/test_feature_calculations.py:1319:9: E122 continuation line missing indentation or outdented ./tests/units/feature_extraction/test_feature_calculations.py:1322:49: E231 missing whitespace after ',' ./tests/units/feature_extraction/test_feature_calculations.py:1322:68: E231 missing whitespace after ',' ./tests/units/feature_extraction/test_feature_calculations.py:1325:5: E303 too many blank lines (2) ./tests/units/feature_extraction/test_feature_calculations.py:1326:9: E265 block comment should start with '# ' ./tests/units/feature_extraction/test_feature_calculations.py:1335:9: E122 continuation line missing indentation or outdented ./tests/units/feature_extraction/test_feature_calculations.py:1336:9: E122 continuation line missing indentation or outdented ./tests/units/feature_extraction/test_feature_calculations.py:1337:9: E122 continuation line missing indentation or outdented ./tests/units/feature_extraction/test_feature_calculations.py:1338:9: E122 continuation line missing indentation or outdented ./tests/units/feature_extraction/test_feature_calculations.py:1339:9: E122 continuation line missing indentation or outdented ./tests/units/feature_extraction/test_feature_calculations.py:1340:9: E122 continuation line missing indentation or outdented ./tests/units/feature_extraction/test_feature_calculations.py:1343:9: E265 block comment should start with '# ' ./tests/units/feature_extraction/test_feature_calculations.py:1344:69: E231 missing whitespace after ',' ./tests/units/feature_extraction/test_feature_calculations.py:1347:5: E303 too many blank lines (2) ./tests/units/feature_extraction/test_feature_calculations.py:1348:9: E265 block comment should start with '# ' ./tests/units/feature_extraction/test_feature_calculations.py:1353:9: E122 continuation line missing indentation or outdented ./tests/units/feature_extraction/test_feature_calculations.py:1354:9: E122 continuation line missing indentation or outdented ./tests/units/feature_extraction/test_feature_calculations.py:1355:9: E122 continuation line missing indentation or outdented ./tests/units/feature_extraction/test_feature_calculations.py:1356:9: E122 continuation line missing indentation or outdented ./tests/units/feature_extraction/test_feature_calculations.py:1357:9: E122 continuation line missing indentation or outdented ./tests/units/feature_extraction/test_feature_calculations.py:1358:9: E122 continuation line missing indentation or outdented
```

vanbenschoten:
Sorry, just noticed I didn't push my style updates :/ Pushing now.

vanbenschoten:
All set - @nils-braun over to you!

nils-braun:
Nice, its in!

vanbenschoten:
Thanks! I greatly enjoyed working together - excited to keep partnering in
the future!

PR #798, opened 2021-01-23
Added new query similarity counter feature

seanlaw:
@nils-braun I've implemented all of the things that we had discussed. Would you mind taking a look and see if we're good to go?

### DIFF
--- a/tsfresh/feature_extraction/feature_calculators.py
+++ b/tsfresh/feature_extraction/feature_calculators.py
@@ -2248,12 +2248,12 @@ def query_similarity_count(x, param):
             and isinstance(Q, (np.ndarray, pd.core.series.Series))
             and len(Q) >= 3
             and len(x) >= 3
-            ):
+        ):
             distance_profile = stumpy.core.mass(np.asarray(Q).astype(float),
                                                 np.asarray(x).astype(float),
-                                               )
+            )
             count = np.sum(distance_profile <= threshold)
 
         res[i] = (f"query_{i+1}", count)
-    
+
     return res
--- a/tsfresh/feature_extraction/feature_calculators.py
+++ b/tsfresh/feature_extraction/feature_calculators.py
@@ -20,6 +20,7 @@ alphabetically ascending.
 import itertools
 import functools
 import warnings
+from tsfresh.utilities.string_manipulation import convert_to_output_format
 from builtins import range
 from collections import defaultdict
 
@@ -2221,8 +2222,10 @@ def query_similarity_count(x, param):
     This feature calculator accepts an input query subsequence parameter,
     compares the query (under z-normalized Euclidean distance) to all
     subsequences within the time series, and returns a count of the number
-    times the query was found in the time series (within some predefined
-    maximum distance threshold).
+    of times the query was found in the time series (within some predefined
+    maximum distance threshold). Note that this feature will always return
+    `np.nan` when no query subsequence is provided and so users will need
+    to enable this feature themselves.
 
     :param x: the time series to calculate the feature of
     :type x: numpy.ndarray
@@ -2241,31 +2244,23 @@ def query_similarity_count(x, param):
     :return x: the different feature values
     :return type: int
     """
-    res = [0] * len(param)
+    res = {}
+    T = np.asarray(x).astype(float)
 
-    for i, d in enumerate(param):
-        normalize = d.get("normalize", True)
-        if not isinstance(normalize, bool):
-            normalize = True
-        threshold = d.get('threshold', 0.0)
-        Q = d.get('query', None)
-        count = 0
-        if Q is not None and \
-                isinstance(Q, (np.ndarray, pd.core.series.Series)) and \
-                len(Q) >= 3 and \
-                len(x) >= 3:
-
+    for i, kwargs in enumerate(param):
+        key = convert_to_output_format(kwargs)
+        normalize = kwargs.get("normalize", True)
+        threshold = kwargs.get('threshold', 0.0)
+        Q = kwargs.get('query', None)
+        Q = np.asarray(Q).astype(float)
+        count = np.nan
+        if Q is not None and Q.size >= 3:
             if normalize:
-                distance_profile = stumpy.core.mass(np.asarray(Q).astype(float),
-                                                    np.asarray(x).astype(float),
-                                                    )
+                distance_profile = stumpy.core.mass(Q, T)
             else:
-                distance_profile = stumpy.core.mass_absolute(np.asarray(Q).astype(float),
-                                                             np.asarray(x).astype(float),
-                                                             )
-
+                distance_profile = stumpy.core.mass_absolute(Q, T)
             count = np.sum(distance_profile <= threshold)
 
-        res[i] = (f"query_{i+1}", count)
+        res[key] = count
 
-    return res
+    return [(key, value) for key, value in res.items()]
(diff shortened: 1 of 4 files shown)


tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -2212,3 +2213,59 @@ def benford_correlation(x):
...
     # In this case returns a 2x2 matrix, the  [0, 1] and [1, 1] are the values between the two arrays
     return np.corrcoef(benford_distribution, data_distribution)[0, 1]
+
+
+@set_property("fctype", "combiner")
+def query_similarity_count(x, param):
+    """
+    This feature calculator accepts an input query subsequence parameter,
+    compares the query (under z-normalized Euclidean distance) to all
+    subsequences within the time series, and returns a count of the number
```
nils-braun, line 2223: Just a small typo
```suggestion
    subsequences within the time series, and returns a count of the number of
```

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -2212,3 +2213,59 @@ def benford_correlation(x):
...
+                  `norm` (bool) to `False.
+    :type param: list
+    :return x: the different feature values
+    :return type: int
+    """
+    res = [0] * len(param)
+
+    for i, d in enumerate(param):
+        normalize = d.get("normalize", True)
+        if not isinstance(normalize, bool):
```
nils-braun, line 2248: I would be very careful with that. Python users are used to use "untypical types", as Python has a very sophisticated bool-cast system. Example: 0 and 1 are very common alternatives to False and True and work exactly the same way as the boolean constants. If you want to check for `None`, just use `if normalize is None` (or maybe I misunderstood your intention).

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -2212,3 +2213,59 @@ def benford_correlation(x):
...
+            normalize = True
+        threshold = d.get('threshold', 0.0)
+        Q = d.get('query', None)
+        count = 0
+        if Q is not None and \
+                isinstance(Q, (np.ndarray, pd.core.series.Series)) and \
```
nils-braun, line 2254: Again here, I would not check for the explicit type. You basically do not care if it is a numpy array or a series or anything else which has the same interface. Example: there are extension types to both numpy and pandas. They might or might not pass your type check, but have definitely the same interface and behavour. I would not check at all - there will be a somewhat meaningful error message in case something breaks.

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -2212,3 +2213,59 @@ def benford_correlation(x):
...
+                len(Q) >= 3 and \
+                len(x) >= 3:
```
nils-braun, line 2256: This part (`len(x) >= 3`) is either true for all params or for none, so you could check that already in the beginning

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -2212,3 +2213,59 @@ def benford_correlation(x):
...
+
+            if normalize:
+                distance_profile = stumpy.core.mass(np.asarray(Q).astype(float),
+                                                    np.asarray(x).astype(float),
```
nils-braun, line 2260: This conversation can also be done just once in the beginning.

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -2212,3 +2213,59 @@ def benford_correlation(x):
...
+                                                    )
+            else:
+                distance_profile = stumpy.core.mass_absolute(np.asarray(Q).astype(float),
+                                                             np.asarray(x).astype(float),
+                                                             )
+
+            count = np.sum(distance_profile <= threshold)
+
+        res[i] = (f"query_{i+1}", count)
```
nils-braun, line 2269: Concerning the output format:
`tsfresh` has a defined output format and column naming, which should depend on the parameters and names. 
The output of a combiner should be a list of tuples `key, result` where key is the result of `key = convert_to_output_format(param)`. (import from `from tsfresh.utilities.string_manipulation import convert_to_output_format`). 

See also our [documentation](https://tsfresh.readthedocs.io/en/latest/text/feature_calculation.html) on what this function basically does.

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -2212,3 +2213,59 @@ def benford_correlation(x):
...
+    """
+    res = [0] * len(param)
+
+    for i, d in enumerate(param):
+        normalize = d.get("normalize", True)
+        if not isinstance(normalize, bool):
+            normalize = True
+        threshold = d.get('threshold', 0.0)
+        Q = d.get('query', None)
+        count = 0
```
nils-braun, line 2252: for tsfresh, the result of NaN means "can not calculate" or "something is wrong". I think this is also the case for you, so I would replace this with 
```suggestion
        count = np.NaN
```

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -2212,3 +2213,59 @@ def benford_correlation(x):
...
+
+
+@set_property("fctype", "combiner")
+def query_similarity_count(x, param):
+    """
+    This feature calculator accepts an input query subsequence parameter,
+    compares the query (under z-normalized Euclidean distance) to all
+    subsequences within the time series, and returns a count of the number
+    times the query was found in the time series (within some predefined
+    maximum distance threshold).
```
nils-braun, line 2225: Maybe mention, that users need to enable this feature by themselves as the default implementation will always return NaN.

nils-braun (commented):
Cool! Thanks for the addition. I have some comments on the implementation. After those are fixed I will have another look.

tsfresh/feature_extraction/feature_calculators.py
seanlaw: Well, I'm allowing the user to omit the `normalize` parameter (i.e., `normalize` is optional and defaults to `True` when not provided). Currently, we allow three optional parameters: `query`, `threshold`, and `normalize` and they are all optional. It isn't clear to me if defaults are automatically taken from `settings.py` if some or all are missing or when a user provides a mixture of parameters in a list.

tsfresh/feature_extraction/feature_calculators.py
seanlaw: I will remove the `isinstance` check

tsfresh/feature_extraction/feature_calculators.py
seanlaw: I see. I will take a look

tsfresh/feature_extraction/feature_calculators.py
seanlaw: Good idea. Thank you

tsfresh/feature_extraction/feature_calculators.py
seanlaw: I think this is fixed now.

seanlaw:
@nils-braun I think I've addressed all of the comments. They were very helpful so thank you!

nils-braun:
I just fixed some merge conflicts, but should be ready to go immediately!

seanlaw:
Very exciting!

seanlaw:
@nils-braun Thank you for all of your help, support, and collaboration!

PR #805, opened 2021-02-07
Issue# 786:  Update ar_coefficient 

hassan-jvd90:
Updated ar_coefficient to use AutoReg for AR model fitting


hassan-jvd90:
@nils-braun  I think the test cases are failing due to an older version of statsmodels as a dependency. 

image

image

nils-braun:
First of all, thanks for the PR :-)
Feel free to bump the dependency of the statsmodels package! Both in `requirements.txt` and in `.github/workflows/test.yml` to the version you would like to have.

nils-braun:
Just as additional comment, also the tests with higher dependencies (which install statsmodels==0.12.2) are failing - probably due to a different result of the feature. Is this expected?

hassan-jvd90:
> 
> 
> First of all, thanks for the PR :-)
> Feel free to bump the dependency of the statsmodels package! Both in `requirements.txt` and in `.github/workflows/test.yml` to the version you would like to have.

Thanks a lot!



> 
> 
> Just as additional comment, also the tests with higher dependencies (which install statsmodels==0.12.2) are failing - probably due to a different result of the feature. Is this expected?

I found this error earlier

image


It looked odd to me that the extracted_features.columns set wasn't greater than the some_expected_features. If I am not wrong it probably needs numbers to compare and doesn't resolve the objects __cmp__ or __gt__. So, I compared the lengths instead and some test cases are passing.

Not sure, about the others.

nils-braun:
The two things that are compared here are sets. So the greater than actually checks for set inclusion, which means: the selected features should at least include the mentioned features. I guess that with your update, the feature now has a different value is therefore not selected anymore in the feature selection.
If you expect the feature result to change I have no problem in changing the tests. If you however expect the feature calculator to give exactly the same result, then we need to look into this :-)

nils-braun:
@hassan-jvd90 - I did some test fixes on the main branch. Could you merge them in and see if the tests are still failing? Is there anything I can do to help you with those tests?

I am planning to do a new release of tsfresh end of this week. There is no need to hurry (as there will be definitely another release), just for your information!

nils-braun:
@hassan-jvd90 - is there anything I can help you with? Should I also have a look or do you want to debug it?

kempa-liehr:
Hi @hassan-jvd90,
Many thanks for your pull request. It seems that we both have worked on the problem independently :-)
Cheers,
Andreas

PR #825, opened 2021-03-16
Update: psd_aggregated

theinem:
Features based on power spectral density: "mean_frequency", "median_frequency", "peak_frequency", "mean_power". The Scipy's periodogram is used to calculate the base signal for the features. 

This features are useful in frequency domain analisys for EMG signals. The main reference for this features are: [1] Phinyomark, A., Phukpattaranont, P., & Limsakul, C. (2012). Feature reduction and selection for EMG signal classification. Expert systems with applications, 39(8), 7420-7431.

### DIFF
--- a/tsfresh/feature_extraction/feature_calculators.py
+++ b/tsfresh/feature_extraction/feature_calculators.py
@@ -27,7 +27,7 @@ from collections import defaultdict
 import numpy as np
 import pandas as pd
 from numpy.linalg import LinAlgError
-from scipy.signal import cwt, find_peaks_cwt, ricker, welch
+from scipy.signal import cwt, find_peaks_cwt, ricker, welch, periodogram
 from scipy.stats import linregress
 from statsmodels.tools.sm_exceptions import MissingDataError
 from matrixprofile.exceptions import NoSolutionPossible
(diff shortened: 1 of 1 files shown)


tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -1176,6 +1176,107 @@ def get_kurtosis(y):
...
+        Returns the mean power from a given power spectral density. See the feature MNP in reference [1] in psd_aggregated(x, param)
+        
+        :param power_spectrum: the power spectral density
+        :type power_spectrum: np.array
+        :param frequency: the frequency for each value of the power spectrum
+        :type frequency: np.array
+        :return: the value of the feature
+        :return type: float
+        """
+        return np.sum(power_spectrum)/len(power_spectrum)
```
nils-braun, line 1264: There is also a np.mean function - could you use that?

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -1176,6 +1176,107 @@ def get_kurtosis(y):
...
+        for i, num in enumerate(power_spectrum):
+            left_sum += num
+            distance = np.abs(half_sum - left_sum)
+            if best_distance > distance:
+                best_index = i
+                best_distance = distance
+            #the array has only positive values, if there is no better distance, we have found the index
+            if best_distance == previous_distance:
+                return frequency[i]
+            previous_distance = distance
```
nils-braun, line 1238: Is there a reason you implemented the median by yourself and did not just use the median implementation (or quantile) of numpy? Typically, those a very fast and will also cover some edge cases

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -1176,6 +1176,107 @@ def get_kurtosis(y):
...
+        Returns the mean frequency from a given power spectral density. See the feature MNF in reference [1] in psd_aggregated(x, param)
+
+        :param power_spectrum: the power spectral density
+        :type power_spectrum: np.array
+        :param frequency: the frequency for each value of the power spectrum
+        :type frequency: np.array
+        :return: the value of the feature
+        :return type: float
+        """
+        return np.sum(frequency * power_spectrum) / np.sum(power_spectrum)
```
nils-braun, line 1211: Also here it might make sense to use `np.mean`

nils-braun (commented):
Well done, thank you for your first PRs to tsfresh and welcome!
I had a few questions about the overall setting.
Could you also have a look why the unit tests are failing and could you add unittests for your new feature? We try to aim a high test coverage at least for the feature calculators.

MaxBenChrist:
Are you aware of https://docs.scipy.org/doc/scipy/reference/signal.html? There is a wide range of signal processing techniques that is already implemented. I unfortunately do not have much experience with signal processing

nils-braun:
@theinem - is there anything I can help you to finish with this PR?

theinem:
I would like to make the improvements you asked for, but for several weeks I have had serious problems with available time and it will continue like this for several more weeks/months 😓

nils-braun:
@theinem - this is not a problem from our side. We are happy to get your code in also in a few weeks. If you feel you can not make it (at all), just tell use and we will try to take over!

theinem:
Thanks, @nils-braun. I think it would be better if the community uses that code and improves it. It is definitely very complicated for me due to my current responsibilities.

tsfresh/feature_extraction/feature_calculators.py
theinem: On page 7424 in [1] is described the equation for the median frequency and is said: _"MDF is a frequency at which the spectrum is divided into two regions with equal amplitude"_, I'm not sure if that definition is the same as a traditional median (I'm conceptually not that strong to make that conclusion), maybe using the median yields the same results as this algorithm, but I replicated the equation from the paper in my original reference.

PR #826, opened 2021-03-16
Add: slope_sign_change, log_detector and willison_amplitude

theinem:
Features : slope_sign_change, log_detector and willison_amplitude.

This features are popular and useful in time domain analisys for EMG signals. The main reference for this features are: [1] Phinyomark, A., Phukpattaranont, P., & Limsakul, C. (2012). Feature reduction and selection for EMG signal classification. Expert systems with applications, 39(8), 7420-7431.

### DIFF
--- a/tsfresh/feature_extraction/feature_calculators.py
+++ b/tsfresh/feature_extraction/feature_calculators.py
@@ -2361,3 +2361,66 @@ def query_similarity_count(x, param):
         res[key] = count
 
     return [(key, value) for key, value in res.items()]
+
+
+@set_property("fctype", "simple")
+def slope_sign_change(x, t=0):
+    """
+    Returns the number of times that slope of the EMG signal changes sign. This feature meassures frequency. It uses a threshold to avoid noise. See SSC in [1].
+
+    :param x: the time series to calculate the feature of
+    :type x: numpy.ndarray
+    :param t: value used as threshold
+    :type t: float
+    :return: the value of this feature
+    :return type: int
+
+    References
+    |  [1] Phinyomark, A., Phukpattaranont, P., & Limsakul, C. (2012). Feature reduction and selection for EMG signal classification. Expert systems with applications, 39(8), 7420-7431.
+    """
+    # Calculation of feature as int
+    x = np.asarray(x)
+    left = x[1:-1] - x[:-2]
+    right = x[1:-1] - x[2:]
+    product = left * right
+    f = np.sum(product > t)
+    return f
+
+
+@set_property("fctype", "simple")
+def log_detector(x):
+    """
+    Returns a number related to the "force" of a signal. These approaches are based on a functional mathematical model for EMG signal generation. See Log detector in [1] and [2].
+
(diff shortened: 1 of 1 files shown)


tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -2361,3 +2361,66 @@ def query_similarity_count(x, param):
...
+    :return type: float
+    
+    References
+    |  [1] Phinyomark, A., Phukpattaranont, P., & Limsakul, C. (2012). Feature reduction and selection for EMG signal classification. Expert systems with applications, 39(8), 7420-7431.
+    |  [2] Zardoshti-Kermani, M., Wheeler, B. C., Badie, K., & Hashemi, R. M. (1995). EMG feature evaluation for movement control of upper extremity prostheses. IEEE Transactions on Rehabilitation Engineering, 3(4), 324-333.
+    """
+    # Calculation of feature as float
+    x = np.asarray(x)
+    n = len(x)
+    x = x + (x == 0)*0.000000000001
```
MaxBenChrist, line 2407: what is the motivation behind this line?

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -2361,3 +2361,66 @@ def query_similarity_count(x, param):
...
+    :param x: the time series to calculate the feature of
+    :type x: numpy.ndarray
+    :return: the value of this feature
+    :return: type: int
+
+    References
+    |  [1] Phinyomark, A., Phukpattaranont, P., & Limsakul, C. (2012). Feature reduction and selection for EMG signal classification. Expert systems with applications, 39(8), 7420-7431.
+    """
+    # Calculation of feature as int
+    return np.sum(np.abs(x[1:] - x[:-1]) > t)
```
MaxBenChrist, line 2425: instead of `x[1:] - x[:-1]` I think you could use `np.diff`

MaxBenChrist (commented):
can you add some unit tests to this pr?

nils-braun:
Great, @theinem ! Three really interesting features. As @MaxBenChrist mentioned, could you add some unittests for your new feature calculators? We try to test every calculator :-)
Also, you new calculators need settings, so you need to add them to the long list in `settings.py` to the `ComprehensiveFCParameters` (that will also solve your test failures).
Could you also fix the python style issues?

theinem:
As I said in my other PR, I would like to make the improvements you asked for, but for several weeks I have had problems with available time and it will continue like this for several weeks/months.

tsfresh/feature_extraction/feature_calculators.py
theinem: I investigated it better and I think that what I did is not a good strategy. The idea was to avoid errors generated by values equal to zero but we use a low value, and this tends to infinity.

nils-braun:
Thanks for the info, @theinem. As mentioned in https://github.com/blue-yonder/tsfresh/pull/825, this is not a problem for us!

PR #827, opened 2021-03-18
fix range count config

MaxBenChrist:


MaxBenChrist:
I was thinking about somehow defining validators for each feature calculator. (e.g. range_count parameter should have min<max)

However, that would be a costly operation, it should not be done during extraction but somehow before. Maybe we can define some attributes on the feature calculator that are collected before extraction. and then before extraction, we could validate the extraction config

nils-braun:
Why would you think that this is costly? I guess most of the checks will be simple comparison operations, or?
In principal I am fine with your changes - thanks!

PR #833, opened 2021-03-28
Add Features: absolute_maximum; mean_n_absolut_max | Add install_requirements.sh script

OliEfr:
## Added:
**F1**. I added a `install_requirements.sh` script for convenience
**F2**. I added a `absolut_maximum `feature. This feature calculates the absolut maximum.
**F3**. I added a `mean_n_absolut_max ` feature. This calculates the arithmetic mean of the n absolut maximum values of the time series.

## Tests:
**T1:** I added unit-tests for mean_n_absolut_max.
**T2**: I did not ad unit-tests for absolut_maximum, because there is also no unit-test for maximum.

**T3:** Some tests fail. Most are not on my watch, I think. However, I think the following are on me but I dont know how to resolve them: `test_contains_all_non_high_comp_features` and `test_default_calculates_all_features`. In particular, I dont know how I can add my newly created features here.

## Questions:
**Q1:** The feature mean_n_absolut_max takes an additional input parameter (the number of maximums to consider). Where and how should I check, that this input parameter is valid, e.g. not smaller 0 and so on. Should it be checked in the pytest unittests? Or should I raise an error within the feature calculator function itself? For now, I just return np.nan if the input parameter is not valid.

## Others
- [x]  build the docs
- [x]  Sorry for commiting multiple changes at once and sorry for not creating a new branch. I think, I should have done this. Next time I will.


tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -1547,6 +1547,27 @@ def c3(x, lag):
...
+    :param x: the time series to calculate the feature of
+    :type x: numpy.ndarray
+    :param number_of_maxima: the number of maxima, which should be considered
+    :type number_of_maxima: int
+
+    :return: the value of this feature
+    :return type: float
+    """
+
+    if number_of_maxima < 1:
```
MaxBenChrist, line 1565: maybe use an assert statement here?

MaxBenChrist, line 1565: maybe use an assert statement here? see `_estimate_friedrich_coefficients` as an example

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -1547,6 +1547,27 @@ def c3(x, lag):
...
+        return np.nan
```
MaxBenChrist, line 1566:
```suggestion
    assert number_of_maxima > 0, f" number_of_maxima={number_of_maxima} which is not greater than 1"
```

install_requirements.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
MaxBenChrist, line 1: can we test that script somehow?

.gitignore
```diff
@@ -30,7 +30,8 @@ __pycache__/*
 
 # Unittest and coverage
 htmlcov/*
-.coverage
```
MaxBenChrist, line 33: please do not put the configuration file of coverage into the .gitignore (or I don't get why you do that)

MaxBenChrist (commented):
good first draft, btw. what are those lock files in the dask-worker space for?

install_requirements.sh
OliEfr: I don't feel comfortable writing a test for that script. I have not enough experiences and I am not sure what should be tested here and how to implemtent the tests in the environment.

Basically I copied the code from [how_to_contribute.rst](https://github.com/blue-yonder/tsfresh/blob/main/docs/text/how_to_contribute.rst)

However, I have the following ideas:
- Check whether the isntall script and the infos in how_to_contribust.rst match
- Check, whether all "*requirements.txt" files are listed
- Make sure, all listed files meet the regex "*requirements.txt" and that there is only one additional line (`pip install -e .`)

You can point me in the right direction implementation-wise.

OliEfr:
I noticed that you dont test for the AssertionError in `test_estimate_friedrich_coefficients`. Is there a reason for this?

I added `assertRaises` tests for `test_mean_n_absolute_max`.

OliEfr:
@MaxBenChrist Could you please refer to T3 from my PR? What about `test_contains_all_non_high_comp_features` and `test_default_calculates_all_features`?

nils-braun (commented):
Also thanks from my side. I had a few comments in the code.

Some general points:
* as mentioned by Max, please remove the dask-worker-space files from your PR.
* one of your new features (the `mean_n_absolute_max`) requires a parameter. This means you need to add it to the feature calculator settings with a reasonable default. Especially, you need to add an entry in the `ComprehensiveFCParameters` dict (have a look into the other features which are already there). I would assume reasonable values would be small numbers, probably below 10.

To your questions:
* Q1: testing for `< 0` in your code is fine. Users can change the parameters, so testing in the unittests is not enough. 
* no, there is no reason we do not test the assertion error. That could be added, good point.
* to T3: To fix the two remaining tests, you need to add your new feature to the settings (see above) - as mentioned in the exception of the tests.

OliEfr:
Check. All done.

For testing the install_requirements.sh file, I created another issue. Because I am not sure yet how to implement it.
I understand, that it makes sense to include it in the workflow. However, no workflow so far requires this exact install_requirements.sh file.

nils-braun:
I will have a look this evening, but I think I am still seeing the dask worker files :-)
Thanks already for your time!

nils-braun:
Thanks!

Issue #482, opened 2019-01-14
Better support/integration for custom functions


michetonu:
Hi!

First of all congrats for the great package, it's really well made.

I think `extract_features()` would highly benefit from an easier way to add custom feature calculation functions on the fly, without having to modify the `feature_calculators` source file. Having to clone the repository and change the source code is not very handy, especially when the code might need to be used in production environments. At the same time, I've been needing to use very project-specific functions which are not worth adding to the package through a PR.

My suggestion is therefore to add an optional parameter to `extract_features`, which could be a dictionary such as `{'function_name': foo}`, where foo is the actual function to be called. Then each function in `default_fc_parameters` is fetched from `feature_calculators`, if it's not found, it's called from the new dictionary. The flow would look something like this:

```
# Custom function to add
def count(x):
    return len(x)

# Dictionary of custom functions
custom_functions = {
     'count': count
}

# The settings dictionary looks exactly the same
settings = {
    'mean': None,
    'count': None
}

# The extract_features call just has an extra parameter
extract_features(settings, custom_functions=custom_functions ....}
```

In the backend, in `feature_extraction.extraction._do_extraction_on_chunk`, line 286, would be modified to something like:

```
for function_name, parameter_list in fc_parameters.items():
    if hasattr(feature_calculators, function_name):
         func = getattr(feature_calculators, function_name)
    else:
        func = custom_functions.get(function_name)
    # Maybe raise an informative error here if it's not found
```

It doesn't need to be exactly like this, but this way would be pretty straightforward to implement.
If there is no reason why we **shouldn't** do this, I'm happy to make a PR and continue the technical discussion there.

Thanks!

MaxBenChrist:
sorry for the late response. That sounds like a very interesting idea, do you want to submit a pr for that?

michetonu:
@MaxBenChrist Sure! Will come soon.

michetonu:
@MaxBenChrist sadly the approach of passing a dictionary with the functions does not work, because the pickling that occurs during the multiprocessing cannot deal with locally defined functions. The alternatives I came up with are:

1) Pass the path to a separate module containing the custom functions (but it's not very pretty)
2) Pass a class containing the custom functions (which might not very intuitive to make for a "regular" user)

Thoughts?

MaxBenChrist:
Ah yeah, well that is unfortunate. 

1.) will break if tsfresh is run on a cluster, right?
[...]

PR #845, opened 2021-04-17
Make settings pickable

nils-braun:
As discussed in #482, there is the need to allow developers to integrate custom feature definitions into tsfresh, without editing the `feature_calculators.py` file (which e.g. means you need to download and build from source).

In principle, that can be done quite easily with (a) additional arguments to `extract_features` or monkeypatching as shown in #482. The problem however is, that these methods either break on the boundary between processes or machines (because self-defined functions have problems when pickling) or do not work on all OSs.

With this PR, I am proposing another solution. As mentioned in the issue, `cloudpickle` is able to pickle and unpickle functions correctly. As multiprocessing is not using cloudpickle by default, I had to "cheat" a bit and make the settings dictionaries use cloudpickle by under the hood.

With the changes, it is now possible to create custom feature extractors "online" and add them to the settings:

```python
from tsfresh.feature_extraction.settings import MinimalFCParameters

# This is our new feature, taken from @dbarbier in https://github.com/blue-yonder/tsfresh/issues/482
def last_n(x, n):
    return x[-n]

# now create our settings object and add our new feature calculator
# please note that the function is the key and the parameters are the values
# we are using the minimal settings, but will also work with others
settings = MinimalFCParameters()
settings[last_n] = [{"n": 1}, {"n": 2}]

# works with multiprocessing (at least on my linux)
extract_features(df, column_id="id", column_sort="time", default_fc_parameters=settings, n_jobs=4)
```

### DIFF
--- a/docs/text/how_to_add_custom_feature.rst
+++ b/docs/text/how_to_add_custom_feature.rst
@@ -46,8 +46,8 @@ You can write such a simple feature calculator, that returns exactly one feature
         :return type: bool, int or float
         """
         # Calculation of feature as float, int or bool
-        f = f(x)
-        return f
+        result = f(x)
+        return result
 
 or with parameter
 
@@ -77,6 +77,8 @@ or with parameter
 ~~~~~~~~~~~~~~~~~~~~
 
 .. code:: python
+    from tsfresh.utilities.string_manipulation import convert_to_output_format
+
 
     @set_property("fctype", "combiner")
     def your_feature_calculator(x, param):
@@ -96,13 +98,10 @@ or with parameter
                  and f the respective feature value as bool, int or float
         :return type: pandas.Series
         """
-        # s is a function that serializes the config
-        # f is a function that calculates the feature value for the config
-        return [(s(config), f(x, config)) for config in param]
+        # Do some pre-processing if needed for all parameters
+        # f is a function that calculates the feature value for each single parameter combination
+        return [(convert_to_output_format(config), f(x, config)) for config in param]
 
-
-After implementing the feature calculator, please add it to the :mod:`tsfresh.feature_extraction.feature_calculators`
-submodule. tsfresh will only find feature calculators that are in this submodule.
 
 Writing your own time-based feature calculators
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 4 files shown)


MaxBenChrist:
wow, really funky design, 

> With this PR, I am proposing another solution. As mentioned in the issue, cloudpickle is able to pickle and unpickle functions correctly. As multiprocessing is not using cloudpickle by default, I had to "cheat" a bit and make the settings dictionaries use cloudpickle by under the hood.

I wonder, are there some limitations on what kind of operations are allowed inside the function body?
What is if there is a package mismatch between the definition of the function body and inside tsfresh?

nils-braun:
As the code is executed on the same machine with the same interpreter (in multiprocessing), that is not a problem. For multi-node setups such as Dask, one needs to think about that anywho, so this PR does not make it worse.

vutle:
The n_jobs > 1 does not work on windows 10.
extract_features(df, column_id="id", column_sort="time", default_fc_parameters=settings, n_jobs=4)

vutle:
This changes does not work when we run under interactive environment for job > 1, on Windows 10.

It will work running inside the main function.

nils-braun:
Hi @vutle - oh, this is bad. But it does work for non-custom feature extractors?

vutle:
Yes, it works for non custom feature extraction if no_job < 2.

Regards
Vu

vutle:
I found that if you put the custom feature extraction inside a class and you initialise this class before usage,  
it will work correctly in both interactive and non interactive mode for any no_job.   Since the function is registered in the memory address.

nils-braun:
Can you maybe share the error message you see (when it fails in parallel mode)?

enesok:
Hi, 
is there a possibility to pick a subset of e.g. Comprehensivelist and add your custom features? A small example: 
settings = ComprehensiveFCParameters()
settings[diff] = [{"n": 1}]
feature_list_small = [  'minimum', 'maximum',  diff]
def get_features(features_of_interest, tsfresh_list = settings):
        return {key: tsfresh_list[key] for key in features_of_interest }
custom_list=get_features(feature_list_small)

extract_features with a single job stalls with this example, although there are no problems with MinimalFCParameters.

nils-braun:
Hi @enesok - sorry for the massive delay. The documentation on our website (https://tsfresh.readthedocs.io/en/latest/index.html) works for me. The `settings` is actually just a dictionary. If you put in only those feature calculators that you want, `tsfresh` will only calculate those. 

As a hint: if you comment on already closed PRs, there is a high chance people will miss it (because the topic is already closed). Best is to create a new issue in this case!

PR #852, opened 2021-04-24
Enable multiprocessing only for n_jobs > 1

pnb:
After several months away from TSFRESH, I updated to the latest version and encountered some very length error messages. It seemed like the messages were due to multiprocessing, so I set `n_jobs` in `RelevantFeatureAugmentor` to 1 thinking that would fix it by forcing single processing. It didn't fix it, and after some docs reading I discovered `n_jobs=0` is actually the way to force single processing. This is a bit un-intuitive, I think. This pull request changes the behavior so `n_jobs=1` will have the same behavior as `n_jobs=0`.

There might be a good reason why `n_jobs=1` should spawn multiprocessing, but this would have saved me a lot of debugging so I thought I'd suggest it.

On an aside, I should probably also figure out why multiprocessing was causing problems, but that's for later.

### DIFF
--- a/tsfresh/feature_extraction/extraction.py
+++ b/tsfresh/feature_extraction/extraction.py
@@ -233,7 +233,7 @@ def _do_extraction(df, column_id, column_value, column_kind, column_sort,
 
     if distributor is None:
         if isinstance(data, Iterable):
-            if n_jobs <= 1:
+            if n_jobs == 0 or n_jobs == 1:
                 distributor = MapDistributor(disable_progressbar=disable_progressbar,
                                              progressbar_title="Feature Extraction")
             else:
--- a/tsfresh/feature_extraction/extraction.py
+++ b/tsfresh/feature_extraction/extraction.py
@@ -233,7 +233,7 @@ def _do_extraction(df, column_id, column_value, column_kind, column_sort,
 
     if distributor is None:
         if isinstance(data, Iterable):
-            if n_jobs == 0:
+            if n_jobs <= 1:
                 distributor = MapDistributor(disable_progressbar=disable_progressbar,
                                              progressbar_title="Feature Extraction")
             else:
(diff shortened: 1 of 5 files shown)


tsfresh/feature_extraction/extraction.py
```diff
@@ -233,7 +233,7 @@ def _do_extraction(df, column_id, column_value, column_kind, column_sort,
 
     if distributor is None:
         if isinstance(data, Iterable):
-            if n_jobs == 0:
+            if n_jobs <= 1:
```
MaxBenChrist, line 236: I think that if statement is too loose, for `n_jobs=-1` you want to use all cores.

pnb: Excellent point! I fixed it now.

MaxBenChrist:
can you create a unit test for this? and maybe document it in the changes? 

> There might be a good reason why n_jobs=1 should spawn multiprocessing

maybe somebody uses the case `n_jobs=1` as a minimal test for multprocessing?

pnb:
I changed the existing unit test so that it uses `n_jobs=2` for the multiprocessing test. This works well on my computer, and the speedup of parallel processing is evident versus the subsequent serial processing test.

Re: documenting it in *CHANGES.rst*, I'm not quite sure how this works -- should I create a new 0.18.1 version number? Or how are the not-yet-released changes documented?

nils-braun:
Thanks for asking, @pnb (and thanks for the PR!).

Normally, I fill the changelog on a new release manually with all the changes, but it is good to already include it now (so that I do not forget about this). You can just add a section

```
Unreleased
=========
```
in the document and add it there. I will handle the rest on the next release.

Thanks again for your contribution!

pnb:
Great, I added that now! I put it under a "Bugfixes" subsection even though I wouldn't consider it a bug exactly, but that seemed like the semantically closest option. You might have a better idea.

nils-braun:
That's in!

PR #853, opened 2021-04-26
Fix other bugs in documentation page (#851)

tanlin2013:
Add missing space and line break for fixing system message in documentation.


nils-braun:
Thank you so much @tanlin2013 !

PR #868, opened 2021-06-20
Docs regarding irregular time series 

pfornia:
Adds paragraphs to intro and faq in docs to further clarify that tsfresh only uses timestamp for ordering, not time intervals.


nils-braun (approved):
I like! Very clear and adds a lot of value to the user!

PR #870, opened 2021-06-30
Fix denominator in docstring for mean_abs_change

abroekhof:
Mean absolute change is over the differences of consecutive values, so the denominator should be `n-1`, not `n`. Numpy `diff` confirms this, e.g.

```
>>> x = np.array([1, 2, 4, 7, 0])
>>> np.diff(x)
array([ 1,  2,  3, -7])
```


nils-braun:
Thanks @abroekhof! You are completely correct.

PR #874, opened 2021-07-01
fixed missing datetime for timebased features and allowed warnings to be silenced

Apsylem:
dask_feature_extraction_on_chunk did not forwardthe sort_by column as index to the time dependend feastures such al 'linear_trend_timewise. Also The option to forbid warnings has been added 'dask_feature_extraction_on_chunk'

PR #877, opened 2021-07-09
better meta inferrence, to prevent indexing error while parquetting

Apsylem:
a unneccary indexing over cloumns in _feature_extraction_on_chunk_helper has been removed

but mainly, an option "taste_of_pandas_df" has been addet to dask_feature_extraction_on_chunk.
This example of the input is used to infer the correct meta in df.apply. This was neccecery due to  missing and misleading index values while parquetting the data after tsfresh feature extraction. No Bug report has been filed for this issue.


nils-braun:
This PR also includes the changes of #874. I will wait for #874 to be finished (minus the changes in this PR) - after that we can have a look into the changes here individually.

Apsylem:
I am sorry, this is my first pull request in a row and one of my first in general. Is there a better way to build upon my own changes, if the old pull request has not yet been accepted? I would aapreciate any remarks.

nils-braun:
Cool! Then thanks for taking your time and starting with open-source development!

This is a good question and there are some options:
* Easiest solution is to just wait for the other PR to be included before creating a new one. That sounds trivial, but is the solution I normally use.
* If you want to open the PR anyway, open it _against_ the older PR. We can then work on both PR in parallel, the changes of the old PR are automatically included in the new one and are still separate. In this case, you typically merge the PRs in the order "new, old". Just make sure you mention all related persons and explain them why we have two PRs.
* You can also open the PRs as you have done but clearly state the fact that those depend on each other. I would however not recommend to do so, because it is hard to review the changes independently.

The absolut best option is to separate the changes into two independent PRs. That is not always possible though.

Apsylem:
I am sorry, but I don't see the whitespaces, the error report on my other pull request, showed the locations to be modified.

here it says : 
 - hook id: trailing-whitespace
- exit code: 1
- files were modified by this hook

does this mean, the trailing whitespaces have already been removed ?

nils-braun:
Hi @Apsylem !
You can check the [full log](https://github.com/blue-yonder/tsfresh/pull/877/checks?check_run_id=3031620674) out the automatic build result (or by clicking on "Details"). It will show you all problematic lines. The build will not edit your code (fortunately ;-)) and will therefore not fix the style errors for you.

If you have followed our [installation for developers](https://tsfresh.readthedocs.io/en/latest/text/how_to_contribute.html#installation) description on the docu, you should have "pre-commit" installed, which will trigger fixing these style issues locally on your computer before you push to github. If you do not want to do this, you can at least call `black` and `isort` by yourself (but using `pre-commit` is preferred).
If you do not know pre-commit, have a look into its [docu](https://pre-commit.com/).

PR #885, opened 2021-08-19
Multiple Group IDs for rolling_time_series

mmcux:
In some cases you have multiple id columns which define a time series, i.e. factory and machine type (Factory A & Machine A; Factory B & Machine A; ...). You could now create a new single id column beforehand; this would be ideally just a tuple of the combinations. Unfortunately the `rolling_time_series()` function can't handle them right now (at least it didn't work for me).
Therefore I adapted the code a little bit in this function to handle multiple id columns.

Moreover if the column id name is different than "id", i.e. "machine_name", this column is also present in the rolled dataframe after calling `rolling_time_series()`.
```
from tsfresh.utilities.dataframe_functions import roll_time_series
df_rolled = roll_time_series(df, column_id="machine_name", column_sort="time")

df_features = extract_features(df_rolled, column_id="id", column_sort="time")
```
A naive user (like me) would just pass the rolled dataframe directly into the extract_features function. With the above workflow the `extract_features()` function would

- fail if the column "machine_name" contains strings
- would run perfectly and calculate features for "machine_name" if the column contains numerical values -> this doesn't make sense for me as we group indirectly by this column anyway and calculating features for id's...well..

Therefore I dropped the initial id colum(s) in this PR after rolling. The id's are included in the new created "id" and will be easy accessible again after the extracting the features from the rolled dataframe.

I am sure there is a lot of room for improvements in my code changes, but for my use cases it works and I would be happy to get your feedback and thoughts about this.

### DIFF
--- a/tests/units/utilities/test_dataframe_functions.py
+++ b/tests/units/utilities/test_dataframe_functions.py
@@ -145,6 +145,85 @@ class RollingTestCase(TestCase):
             n_jobs=0,
         )
 
+    def test_multiple_index(self):
+        xy_class = pd.DataFrame(
+            {"a": [1, 2], "b": [5, 6], "time": range(2),  "first_id": ["x", "y"]})
+
+        yx_class = pd.DataFrame(
+            {"a": [3, 4], "b": [7, 8], "time": range(2), "first_id": ["y", "x"]})
+
+        df_intermediate = pd.concat([xy_class, yx_class])
+        a_class = df_intermediate.copy()
+        a_class["second_id"] = "a"
+        b_class = df_intermediate
+        b_class["second_id"] = "b"
+        df_full = pd.concat([a_class, b_class], ignore_index=True)
+
+        """ df_full is
+            a	b	 time  first_id	second_id
+        0	1	5  	    0	    x	    a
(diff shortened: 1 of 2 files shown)


nils-braun:
Great idea @mmcux! I like both your ideas (allow for multiple columns in the id as well as remove the created id, as it can be easily recreated).
However, I would propose not to allow for multiple id columns as parameter in the `roll_time_series` function. The reason is as following: all our "main" functions in `tsfresh` allow for the same input parameters (column_id, column_sort, ...) and it would be a bit inconsistent if one of them allows for a list and the others don't. We could of course now change all functions to allow for lists, but that would be a lot of work. Therefore I would propose to go the other way you suggested: make sure the `roll_time_series` function can actually work with a id column, which consists of tuples. There is not much to change in this case, basically [one line](https://github.com/blue-yonder/tsfresh/blob/feature/rolling_multiple_group_ids-proposal/tsfresh/utilities/dataframe_functions.py#L531). What do you think? If you like it, feel free to just cherry-pick the commit this link points to (I have also adjusted the tests) in your PR. We can think of having a dedicated example in the documentation for this use-case.

PR #898, opened 2021-10-03
Fix compatibility issue with statsmodels 0.18.0

### LINKED ISSUE #897: Compatibility issue with statsmodels 0.18.0
**The problem**:

Since v0.18.0, statsmodels stops supporting `unbiased` and uses `adjusted` (https://github.com/statsmodels/statsmodels/commit/3d6c3884b9804842ff039c8c5d63a9ccf519867a#diff-771ed9ae3d316dec0997427f2182d015dbfdaf6935335428bbf177ce1968bc72), and use of `statsmodels.tsa.stattools.acf` in `tsfresh.feature_extraction.ceature_calculators` now causes `TypeError`. We need to use `adjusted` instead of `unbiased` in calls for `acf` in statsmodels later than v0.14.0.

wjsi:
Use `adjusted` instead of `unbiased` when calling `acf`. When `acf` does not have an argument `adjusted`, we creates a compatible version.

This PR fixes #897.

### DIFF
--- a/tsfresh/feature_extraction/feature_calculators.py
+++ b/tsfresh/feature_extraction/feature_calculators.py
@@ -18,6 +18,7 @@ alphabetically ascending.
 """
 
 import functools
+import inspect
 import itertools
 import warnings
 from builtins import range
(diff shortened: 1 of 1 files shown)


kempa-liehr:
Hi @wjsi, 
Many thanks for your pull request. I was really helpful in fixing the recent issues related to statsmodels.
However, due to the fact that a related statsmodels issue required to bump up the statsmodels requirement to v0.13, the suggested solution is not required anymore.
Cheers, 
Andreas

PR #912, opened 2021-11-18
statsmodels AR deprecated in v0.11

benman1:
If you take an up-to-date install of statsmodels this will fail with tsfresh with the following error:

```
settings = ComprehensiveFCParameters()
extract_features(data, column_id='Time', default_fc_parameters=settings)

NotImplementedError: AR has been removed from statsmodels and replaced with statsmodels.tsa.ar_model.AutoReg.
```

This was deprecated. Any version before [0.11.0]((https://www.statsmodels.org/dev/release/version0.11.html) should do. This is a quick fix to avoid the problem.


kempa-liehr:
Hi @benman1, 
Thanks for the workaround. Actually, I stumbled upon the same error today in one of my projects. 
Cheers,
Andreas

PR #917, opened 2021-12-02
typo spark_feature_extraction_on_chunk doc string

shayshimon:
It was written in the doc string that spark_feature_extraction_on_chunk returns a dask dataframe while the function actually returns a spark dataframe.


kempa-liehr:
Hi @shayshimon,

Many thanks for your pull request. 

Actually, this is really a dask dataframe: https://docs.dask.org/en/stable/dataframe.html

Cheers, 
Andreas

PR #994, opened 2022-12-24
Added functionality to test multiple versions of python using tox

Scott-Simmons:
## What functionality changed? ##

This PR adds functionality to be able to locally test multiple versions of python using `tox.`

This PR will run the test suite for python versions 3.7, 3.8, 3.9, 3.10, and 3.11 for whichever platform the user runs the tests on. The specific microversion of python depends on what is installed on the user's machine. 

`tox` will skip over running the test suite for a given version if it cannot find a particular python environment.

## How to run? ##

To test on multiple python versions, edit `envlist` in `setup.cfg` to include the python versions you want to test on, and then run:
```bash
tox -r
``` 
in the top-level directory for tsfresh. Currently the versions being tested span between python 3.7-3.11

## Why? ##

This PR will make it much easier to (locally) test tsfresh against multiple versions of python, for a given platform

## What code changed? ##

`setup.cfg` was changed to include some extra configuration information.

`tox` was added to test-requirements.txt

 
 ## Tips for reviewing/running ##

By default, `tox` will look in binary directories for any relevant python interpreters. For example, if  

```bash 
envlist = (py37, py38)
```
exists in `setup.cfg`, then `tox` will look inside binary directories for python executables named similar to python3.7.X and python 3.8.X. If it cannot find any executables for a given python version, then it will skip over testing that version. 

Note that in this PR:

```bash
envlist = py37, py38, py39, py310, py311
```

But it should also be noted that while included in the envlist, py311 is currently not compatible with tsfresh due to a numba dependency conflict. 

### Using `pyenv` in tandem with `tox` 
A recommended way to handle multiple versions of python is with `pyenv`. `pyenv` will allow you to install multiple versions of python in a well-organised fashion.

If you choose to use `pyenv` to manage the different python versions installed on your machine, then the executables of each python version will be in `~/.pyenv/shims/` which will not be found by `tox` by default. A recommended solution to this is to make a `.python-version` file, with the versions of python that you want `tox` to look for. 

For example, if the output of running

```bash
pyenv versions
``` 

shows that you have installed python 3.7.16 and python 3.8.16, then you should put `3.7.16` and `3.8.16` as separate lines in the `.python-version` file. Running `tox -r` will then run the tests for python3.7.16 and python3.8.16, and `tox` will know where to find the relevant python interpreters. 

##### `pyenv` important note

The python version that the `tox` command is initially invoked from matters! 

Running `pyenv which python` will show the version of python from which the `tox` command will be invoked from. If `tox` is initially invoked from a version of python that is not supported by the package (i.e. package is invoked from python 3.6 and `python_requires` is >=3.7), then `tox` will fail for all environments, _including python versions that would otherwise work if `tox` was originally invoked from a version of python supported with the package._

Note that we can still test tsfresh on unsupported versions of python (such as 3.6), provided that `tox` is initially invoked from a version of python that is in tsfresh `python_requires` (such as 3.8).

### Log files

Log files are stored in the `.tox` directory which is created once `tox` is run.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -140,3 +140,23 @@ max-line-length = 120
 # This will be used when updating. Do not change!
 version = 3.2.3
 package = tsfresh
+
+[tox:tox]
+# Tox global conifguration options
+minversion = 4.0.0
+# Which python versions to test for
+envlist = py37, py38, py39, py310, py311
+# Don't throw an error if a specific version of python is not installed
+skip_missing_interpreters = True
+# Ensure that tox uses a virtual environment to build a source distribution
+isolated_build = True
+
+# Tox configuration options for each version of python
+[testenv]
+commands =
+    # display the python version that tests are being run on
+    python -V
+    # install package requirements before running tests
+    python -m pip install .[testing]
+    # run the tests
+    python -m pytest tests/
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,9 @@ tsfresh/notebooks/data/
 # dask
 dask-worker-space
 dask-worker-space/
+
+# python version files
+.python_version
+
+# tox log files
+.tox
--- a/docs/text/how_to_contribute.rst
+++ b/docs/text/how_to_contribute.rst
@@ -67,6 +67,18 @@ you have to:
     pytest
 
 
+To test changes across multiple versions of Python, run:
+
+
+.. code::
+
+    tox -r
+
+
+`tox -r` will execute tests for the Python versions specified in `setup.cfg <https://github.com/blue-yonder/tsfresh/blob/1297c8ca5bd6f8f23b4de50e3f052fb4ec1307f8/setup.cfg>`_ using the `envlist` variable. For example, if `envlist` is set to `py37, py38`, the test suite will run for Python 3.7 and 3.8 on the local development platform, assuming the binaries for those versions are available locally. The exact Python microversions (e.g. `3.7.1` vs `3.7.2`) depend on what is installed on the local development machine.
+
+A recommended way to manage multiple Python versions when testing locally is with `pyenv`, which enables organized installation and switching between versions.
+
 Documentation
 '''''''''''''
 
--- a/docs/text/how_to_contribute.rst
+++ b/docs/text/how_to_contribute.rst
@@ -90,3 +90,9 @@ We use black and isort for styling. They are automatically triggered on every co
 
 
 We are looking forward to hear from you! =)
+
+
+PR Descriptions
+'''''''''''''''
+
+The PR should have a clear and descriptive title, along with a detailed description of the changes made, the problem being addressed, and any relevant tips for reviewers. An example of what this might look like is `here. <https://github.com/blue-yonder/tsfresh/pull/994#issue-1509962136>`_


Scott-Simmons:
The main changes are in `setup.cfg` and `.gitignore`. The changes to the other files are style changes in order to pass the pre-commit hook.

setup.cfg
```diff
@@ -23,7 +23,7 @@ classifier =
     Topic :: Software Development
 
 [options]
-python_requires = >= 3.7
+python_requires = >= 3.7, <3.11
```
Scott-Simmons, line 26: <3.11 due to a numba dependency requirement

Scott-Simmons:
@nils-braun is there still a need for this PR or should it be closed? Happy to pick it up again and get it over the line (if needed)

Scott-Simmons:
If so, will clean up the commits too...

setup.cfg
nils-braun: Hi @Scott-Simmons - does this still hold? And even if yes, tsfresh itself does not have any constraints, but its dependencies (numba in this case) has. I would therefore not like to constrain tsfresh itself, because the constraint should be transitive.

setup.cfg
```diff
@@ -92,3 +92,27 @@ with-docs = 0
...
+# Tox configuration options for each version of python
+[testenv]
+commands =
+    # display the python version that tests are being run on
+    python -V
+    # install package requirements before running tests
+    python -m pip install -r requirements.txt
+    python -m pip install -r test-requirements.txt
+    python -m pip install -r optional-requirements.txt
+    python -m pip install -r rdocs-requirements.txt
```
nils-braun, line 116: Why do we need to install the docs requirements for the tests?

nils-braun (commented):
:scream_cat: I am so sorry @Scott-Simmons - your PR just slipped through and I never ever had a look into it. Really sorry, this is not a good experience for you. Thanks for taking the time. Please do ping me earlier next time :)

I like the changes. If you update to the newest version of `main`, the pre-commit changes should also not be need anymore. After that, I think the changes are already ready to be merged. Do you think it makes sense to also add parts of your PR description into the `how_to_contribute.rst` document?

setup.cfg
Scott-Simmons: Addressed here https://github.com/blue-yonder/tsfresh/pull/994/commits/f8487373a195698f03eff374be69208f816bad42

Scott-Simmons:
> Really sorry, this is not a good experience for you

No apologies needed, all good with me. Only recently remembered this.

> Do you think it makes sense to also add parts of your PR description into the how_to_contribute.rst document?

If I understand you correctly, I've added something to that effect here: https://github.com/blue-yonder/tsfresh/pull/994/commits/a5907777125544128e040150a74c78b5d24ab658

> Error: This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v1`. Learn more: https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/

Looks like the builds are failing for unrelated reason w.r.t this PR. Can look at updating the actions file in a new PR.

Scott-Simmons:
> Looks like the builds are failing for unrelated reason w.r.t this PR. Can look at updating the actions file in a new PR.

@nils-braun marking as ready for review despite the above

setup.cfg
Scott-Simmons: Does not hold anymore. Have updated (ended up rebasing)

Scott-Simmons (commented):
Unsure this PR is needed anymore since we have https://github.com/blue-yonder/tsfresh/blob/1297c8ca5bd6f8f23b4de50e3f052fb4ec1307f8/.github/workflows/test.yml#

nils-braun:
> Unsure this PR is needed anymore since we have https://github.com/blue-yonder/tsfresh/blob/1297c8ca5bd6f8f23b4de50e3f052fb4ec1307f8/.github/workflows/test.yml#

Yes, definitely your PR ist still needed. The file you point to is for CI/CD, but it does not help the average developer. We have this matrix since the start of the library, but we never had it usable for developers. So your change is still needed!

nils-braun:
> If I understand you correctly, I've added something to that effect [here](https://github.com/blue-yonder/tsfresh/pull/994/commits/b7de7d0607156028e1761a7c6a3d2988d67e9980)

Ah sorry, what you did is also great, but I actually meant adding the description on how to run the tox tests into the document (basically the `tox -r`). But the description on how to do PRs is also good (and your is a good example!)

nils-braun (approved):
Looks good now @Scott-Simmons!
You can or can not add the description on the `tox -r` command - I leave it up to you. I will merge the PR tomorrow :) 
Thanks again for the contribution!

Scott-Simmons:
> You can or can not add the description on the tox -r command 

Added in https://github.com/blue-yonder/tsfresh/pull/994/commits/1e9c3bce4a6bc8c13aa37365e73e54b8afc1d23d

> but we never had it usable for developers

Was thinking about this, maybe we could have a docker container for local testing as well? To get around the developer needing to have python versions 3.7, ...3.12 on their development machine. 

Something like:

`makefile`
```makefile
TEST_DOCKERFILE := Dockerfile-test
TEST_CONTAINER := test-tsfresh
test-all:
    docker build -f $(TEST_DOCKERFILE) -t $(TEST_CONTAINER) .
    docker run --rm $(TEST_CONTAINER)
```

`Dockerfile-test`
```Dockerfile
FROM python:3.10-slim # could be something else too...

# Haven't tested this probably more dependencies required...
RUN apt-get update && \
    apt-get install -y python3.7 python3.8 python3.9 python3.10 python3.11 python3.12

RUN pip install tox

# Just an idea may not need to copy whole thing in....
WORKDIR /app
COPY . .

CMD ["tox -r"]
```


@nils-braun if this is something that looks useful I can take this on - I don't think it will be too much work but could be useful for developer experience?

PR #997, opened 2023-02-03
pdm for project management

HolgerPeters:


nils-braun:
Hi @HolgerPeters !
What made you close this PR again? Having a more modern project and dependency setup seems like a very good idea to me :)

HolgerPeters:
hi Nils, basically I closed it because I was not ready to have discussions about it (I wanted to create a PR within my fork so I have a nice diff :-) ). I actually would like to get some changes into tsfresh first with regards to the supported Python version (i think it was Pandas  that dropped support for python version tsfresh still supports and that causes problems).

And then I am happy to contribute with a more-modern setup for the repo

PR #1002, opened 2023-02-23
Replace hard-coded "id" with column_id in `roll_time_series`

david-waterworth:
This fixes the inconsistency of `roll_time_series` aways naming the resulting tuple index as "id" even if the user passes and alternative `id` via the column_id parameter.

### DIFF
--- a/tsfresh/utilities/dataframe_functions.py
+++ b/tsfresh/utilities/dataframe_functions.py
@@ -343,7 +343,7 @@ def _roll_out_time_series(
         else:
             timeshift_value = timeshift - 1
         # and now create new ones ids out of the old ones
-        df_temp["id"] = df_temp[column_id].apply(lambda row: (row, timeshift_value))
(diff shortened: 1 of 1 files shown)


david-waterworth:
I think now that I've made this change, the return from `roll_time_series` also needs to change - the map_reduce may no longer contain an "id"  column yet the return always sorts by "id"

I'll also add tests before this is ready to merge as well.

    shifted_chunks = distributor.map_reduce(
        _roll_out_time_series,
        data=range_of_shifts,
        chunk_size=chunksize,
        function_kwargs=kwargs,
    )

    distributor.close()

    df_shift = pd.concat(shifted_chunks, ignore_index=True)

    return df_shift.sort_values(by=["id", column_sort or "sort"])

nils-braun:
> I think now that I've made this change, the return from roll_time_series also needs to change - the map_reduce may no longer contain an "id" column yet the return always sorts by "id"

Correct - you would want to also change this to `column_id`.

Thanks for your effort!

PR #1024, opened 2023-05-30
Feature/calculate fft energy blocks

anates:
This feature allows the computation of the frequency energy content of the signal, distributed over n frequency bins. This simplifies the detection of dominant frequencies for different process stages.

This pull request is meant as initial step, to receive feedback (both wrt. code, implementation and intention of the feature) and to improve the code, before it can be merged.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,6 +48,7 @@ docs/text/_generated/
 # Virtual environment
 venv*
 activate
+tsfresh-env
 
 # IPython notebooks
 .ipynb_checkpoints
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,4 @@ tsfresh/notebooks/data/
 # dask
 dask-worker-space
 dask-worker-space/
+*.pyc.*
--- a/tsfresh/feature_extraction/feature_calculators.py
+++ b/tsfresh/feature_extraction/feature_calculators.py
@@ -1237,11 +1237,18 @@ def fft_aggregated(x, param):
 def energy_content_frequency_brackets(
     signal: pd.Series, params: dict = {"number_of_bins": 10, "with_normalization": True}
 ) -> list[tuple[str, float]]:
-    """_summary_
+    """
+
+    Calculates the FFT of the signal and groups the content of the FFT into n bins, \
+        to get an estimate of energy content per frequency bucket
 
     Args:
-        signal (pd.Series): _description_
-        params (_type_, optional): _description_.\
+        signal (pd.Series): The distribution for which the frequency energy buckets \
+            should be calculated
+        params (_type_, optional): Parameters used for calculation of the frequency \
+            energy buckets. Parameters to be set are number_of_bins for the amount \
+                of bins, and if the values should be normalized (divided by the size \
+                    of the frequency bucket) or not.\
             Defaults to {"number_of_bins": 10, "with_normalization": True}.
 
     Returns:
--- a/tsfresh/feature_extraction/feature_calculators.py
+++ b/tsfresh/feature_extraction/feature_calculators.py
@@ -1231,6 +1231,63 @@ def fft_aggregated(x, param):
     return zip(index, res)
 
 
+@set_property("fctype", "combiner")
+@set_property("input", "pd.Series")
+@set_property("index_type", pd.DatetimeIndex)
+def energy_content_frequency_brackets(
+    signal: pd.Series, params: dict = {"number_of_bins": 10, "with_normalization": True}
+) -> list[tuple[str, float]]:
+    """_summary_
+
+    Args:
+        signal (pd.Series): _description_
+        params (_type_, optional): _description_.\
+            Defaults to {"number_of_bins": 10, "with_normalization": True}.
+
+    Returns:
+        list[tuple[str, float]]: _description_
+    """
+    # pylint:disable=dangerous-default-value
+
+    time_vec: np.ndarray = signal.index.to_numpy()
+    number_of_bins: int = 10
+    if "number_of_bins" in params.keys():
+        number_of_bins: int = (
+            params["number_of_bins"]
+            if (
+                params["number_of_bins"] <= len(time_vec)
+                and params["number_of_bins"] >= 0
+            )
+            else len(time_vec) - 1
+        )
+    with_normalization: bool = True
+    if "with_normalization" in params.keys():
+        with_normalization = params["with_normalization"]
+
+    freq_vec: np.ndarray = np.fft.rfftfreq(n=len(signal), d=(time_vec[1] - time_vec[0]))
+    fft_vec: np.ndarray = np.abs(np.fft.rfft(signal.to_numpy()))
+    split_freq_vec: list[np.ndarray] = np.array_split(freq_vec, number_of_bins)
+    frequency_energy_vec: list[float] = []
+    frequency_range_vec: list[tuple[float, float]] = []
+    for _, freq_entry in enumerate(split_freq_vec):
+        lower_frequency_index: int = np.where(freq_vec == freq_entry[0])[0][0]
+        upper_frequency_index: int = np.where(freq_vec == freq_entry[-1])[0][0]
+        frequency_energy_vec.append(
+            np.sum(fft_vec[lower_frequency_index:upper_frequency_index])
+            / (len(freq_entry) if with_normalization else 1.0)
+        )
+        frequency_range_vec.append((freq_entry[0], freq_entry[-1]))
+
+    return_val: list[tuple[str, float]] = [
+        (
+            f"bin_{entry_num}_max_{number_of_bins}_low_{int(frequency_range_vec[entry_num][0])}_high_{int(frequency_range_vec[entry_num][1])}",
+            entry,
+        )
+        for entry_num, entry in enumerate(frequency_energy_vec)
+    ]
+    return return_val
+
+
 @set_property("fctype", "simple")
 def number_peaks(x, n):
     """
@@ -1423,7 +1480,6 @@ def spkt_welch_density(x, param):
     if len(pxx) <= np.max(
         coeff
     ):  # There are fewer data points in the time series than requested coefficients
-
         # filter coefficients that are not contained in pxx
         reduced_coeff = [coefficient for coefficient in coeff if len(pxx) > coefficient]
         not_calculated_coefficients = [
@@ -1525,7 +1581,9 @@ def change_quantiles(x, ql, qh, isabs, f_agg):
     try:
         bin_cat = pd.qcut(x, [ql, qh], labels=False)
         bin_cat_0 = bin_cat == 0
-    except ValueError:  # Occurs when ql are qh effectively equal, e.g. x is not long enough or is too categorical
+    except (
+        ValueError
+    ):  # Occurs when ql are qh effectively equal, e.g. x is not long enough or is too categorical
         return 0
     # We only count changes that start and end inside the corridor
     ind = (bin_cat_0 & _roll(bin_cat_0, 1))[1:]
@@ -2180,7 +2238,6 @@ def agg_linear_trend(x, param):
     res_index = []
 
     for parameter_combination in param:
-
         chunk_len = parameter_combination["chunk_len"]
         f_agg = parameter_combination["f_agg"]
 
@@ -2438,7 +2495,6 @@ def matrix_profile(x, param):
 
         # Handle all other Matrix Profile instances
         else:
-
             finite_indices = np.isfinite(m_p)
 
             if feature == "min":
--- a/tests/units/feature_extraction/test_feature_calculations.py
+++ b/tests/units/feature_extraction/test_feature_calculations.py
@@ -236,7 +236,6 @@ class FeatureCalculationTestCase(TestCase):
         self.assertEqualOnAllArrayTypes(sum_values, [], 0)
 
     def test_agg_autocorrelation_returns_correct_values(self):
-
         param = [{"f_agg": "mean", "maxlag": 10}]
         x = [1, 1, 1, 1, 1, 1, 1]
         expected_res = 0
@@ -266,7 +265,6 @@ class FeatureCalculationTestCase(TestCase):
         self.assertAlmostEqual(res, expected_res, places=4)
 
     def test_agg_autocorrelation_returns_max_lag_does_not_affect_other_results(self):
-
         param = [{"f_agg": "mean", "maxlag": 1}, {"f_agg": "mean", "maxlag": 10}]
         x = range(10)
         res1 = dict(agg_autocorrelation(x, param=param))['f_agg_"mean"__maxlag_1']
@@ -280,7 +278,6 @@ class FeatureCalculationTestCase(TestCase):
         self.assertAlmostEqual(res1, 0.77777777, places=4)
 
     def test_partial_autocorrelation(self):
-
         # Test for altering time series
         # len(x) < max_lag
         param = [{"lag": lag} for lag in range(10)]
@@ -532,7 +529,6 @@ class FeatureCalculationTestCase(TestCase):
         )
 
     def test_ratio_beyond_r_sigma(self):
-
         x = [0, 1] * 10 + [10, 20, -30]  # std of x is 7.21, mean 3.04
         self.assertEqualOnAllArrayTypes(ratio_beyond_r_sigma, x, 3.0 / len(x), r=1)
         self.assertEqualOnAllArrayTypes(ratio_beyond_r_sigma, x, 2.0 / len(x), r=2)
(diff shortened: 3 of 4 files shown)


tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -1231,6 +1231,70 @@ def get_kurtosis(y):
     return zip(index, res)
 
 
+@set_property("fctype", "combiner")
+@set_property("input", "pd.Series")
+@set_property("index_type", pd.DatetimeIndex)
+def energy_content_frequency_brackets(
+    signal: pd.Series, params: dict = {"number_of_bins": 10, "with_normalization": True}
```
nils-braun, line 1238: Please do not set the default parameter to a dictionary (https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python). This is not only related to `tsfresh` but for general python: this can lead to very unexpected behaviour (see the linked stackoverflow if you want to know more).
(you even disabled the "dangerous default parameter" pylint check for this ;) )

The good thing is, you do not even need a default argument. tsfresh will always fill this argument, so you can just remove the default.

nils-braun: Thanks for adding type information. But `params` is actually a list of dictionaries.
As you chose a combiner calculator (https://tsfresh.readthedocs.io/en/latest/text/how_to_add_custom_feature.html#step-1-decide-which-type-of-feature-you-want-to-implement) you will get all parameters for this calculator as a list of dictionaries.

This will also need some changes to your code.

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -1231,6 +1231,70 @@ def get_kurtosis(y):
...
+) -> list[tuple[str, float]]:
+    """
+
+    Calculates the FFT of the signal and groups the content of the FFT into n bins, \
+        to get an estimate of energy content per frequency bucket
+
+    Args:
+        signal (pd.Series): The distribution for which the frequency energy buckets \
+            should be calculated
+        params (_type_, optional): Parameters used for calculation of the frequency \
```
nils-braun, line 1248: See above, the parameter is not optional (and the type information is not `__type__`)

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -1231,6 +1231,70 @@ def get_kurtosis(y):
...
+    number_of_bins: int = 10
+    if "number_of_bins" in params.keys():
+        number_of_bins: int = (
+            params["number_of_bins"]
+            if (
+                params["number_of_bins"] <= len(time_vec)
+                and params["number_of_bins"] >= 0
+            )
+            else len(time_vec) - 1
+        )
```
nils-braun, line 1269: You do not need to use a default value for `number_of_bins` here. You can always assume that the key will be set. The reason is that we define the settings ourself :)
This allows you to simplify the code to (basically)
```
number_of_bins = min(params["number_of_bins"], len(time_vec) - 1)
```

nils-braun: I think if a user sets `number_of_bins` to 0 or below this should be a bug and you can just raise an exception here.

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -1231,6 +1231,70 @@ def get_kurtosis(y):
...
+    with_normalization: bool = True
```
nils-braun, line 1270: Same - you can assume the value is set in the param config

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -1231,6 +1231,70 @@ def get_kurtosis(y):
...
+    if "with_normalization" in params.keys():
+        with_normalization = params["with_normalization"]
+
+    freq_vec: np.ndarray = np.fft.rfftfreq(n=len(signal), d=(time_vec[1] - time_vec[0]))
```
nils-braun, line 1274: What if the time stamps are not equally distributed?

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -1231,6 +1231,70 @@ def get_kurtosis(y):
...
+    fft_vec: np.ndarray = np.abs(np.fft.rfft(signal.to_numpy()))
+    split_freq_vec: list[np.ndarray] = np.array_split(freq_vec, number_of_bins)
+    frequency_energy_vec: list[float] = []
+    frequency_range_vec: list[tuple[float, float]] = []
+    for _, freq_entry in enumerate(split_freq_vec):
```
nils-braun, line 1279: You do not need to use `enumerate` if you do not need the index (and it seems you do not need it as you use `_`)

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -1231,6 +1231,70 @@ def get_kurtosis(y):
...
+        lower_frequency_index: int = np.where(freq_vec == freq_entry[0])[0][0]
+        upper_frequency_index: int = np.where(freq_vec == freq_entry[-1])[0][0]
```
nils-braun, line 1281: What are those two lines doing?

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -1231,6 +1231,70 @@ def get_kurtosis(y):
...
+        frequency_energy_vec.append(
+            np.sum(fft_vec[lower_frequency_index:upper_frequency_index])
+            / (len(freq_entry) if with_normalization else 1.0)
+        )
```
nils-braun, line 1285: I would suggest rewriting this in the form
```
variable = np.sum(fft_vec[lower_frequency_index:upper_frequency_index])
if with_normalization:
   variable /= ...
frequency_energy_vec.append(variable)
```

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -1231,6 +1231,70 @@ def get_kurtosis(y):
...
+        frequency_range_vec.append((freq_entry[0], freq_entry[-1]))
+
+    return_val: list[tuple[str, float]] = [
+        (
+            f"bin_{entry_num}_max_{number_of_bins}_low_{int(frequency_range_vec[entry_num][0])}_high_{int(frequency_range_vec[entry_num][1])}",
+            entry,
+        )
+        for entry_num, entry in enumerate(frequency_energy_vec)
```
nils-braun, line 1293: You can use `zip` to iterate through `frequency_energy_vec` and `frequency_range_vec`at the same time

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -1231,6 +1231,70 @@ def get_kurtosis(y):
...
+        upper_frequency_index: int = np.where(freq_vec == freq_entry[-1])[0][0]
+        frequency_energy_vec.append(
+            np.sum(fft_vec[lower_frequency_index:upper_frequency_index])
+            / (len(freq_entry) if with_normalization else 1.0)
+        )
+        frequency_range_vec.append((freq_entry[0], freq_entry[-1]))
+
+    return_val: list[tuple[str, float]] = [
+        (
+            f"bin_{entry_num}_max_{number_of_bins}_low_{int(frequency_range_vec[entry_num][0])}_high_{int(frequency_range_vec[entry_num][1])}",
```
nils-braun, line 1290: Here, we have a conceptional problem.
Each feature calculator in tsfresh is allowed to only return a single number per parameter input. This is important because our feature selection only works for numbers, not lists of numbers.
Additionally, the feature extractor names (the string you return) need to be always the same (independent on the data) and should only depend on the parameters. 

How can we solve this? We can use the same idea that we e.g. used in the `fft_aggregated` calculator. Just make the `entry_num` and additional parameter :) With this, you can return a single result per parameter and just have a lot more parameters.

nils-braun (commented):
Thank you very much for the contribution! We love seeing more feature calculators!
I have added a few comments on a high level. There is one conceptional problem I see with the current implementation - but this can be fixed (I think), so nothing which prevents the feature in general!

Thanks again for your work :) 
Once you finished implementing the comments, I will have another look.

tsfresh/feature_extraction/feature_calculators.py
anates: Fixed, thanks!

tsfresh/feature_extraction/feature_calculators.py
anates: I see three possible options:

- Raise an error
- Interpolate time stamps and values onto equidistant time steps
- Return NaN-values

Which of those options would be the most appropriate, in the light of other available features?

tsfresh/feature_extraction/feature_calculators.py
anates: The idea behind those two lines was to find the index of the frequency in freq_vec which corresponds to the frequency limits given for each entry in split_freq_vec. Anyway, as this code was updated on my side, I will re-write that part.

tsfresh/feature_extraction/feature_calculators.py
anates: From that point of view, I'm a bit unsure if I should leave the calculator to be a combiner or convert it to a "simple" calculator. Converting it into a "simple" calculator would allow me to simplify the code, and speed up calculation. For the time being I therefore decided to convert it, and would appreciate feedback if that decision is appropriate. 
Thanks!

tsfresh/feature_extraction/feature_calculators.py
anates: I chose to raise a warning to inform the user, and to continue with a default value of 1.

nils-braun:
@anates - the combiner/simple calculator and your bin index issue makes it probably quite complex, so let me try to simplify the problem a bit:
as a start, can you implement a function `energy_content_frequency_brackets` which gets all parmeters as keyword arguments and just produces a single return value (float)? Your function would look like
```python
def energy_content_frequency_brackets(signal, with_normalization, bin_index):
```
Would that be possible? I think the function would look something like
```

def energy_content_frequency_brackets(signal, with_normalization, bin_index):
    freq_vec = np.fft.rfftfreq(n=len(signal), d=(time_vec[1] - time_vec[0]))
    fft_vec = np.abs(np.fft.rfft(signal.to_numpy()))
    split_freq_vec = np.array_split(freq_vec, number_of_bins)

    freq_entry = split_freq_vec[bin_index]

    lower_frequency_index = find_nearest(array=freq_vec, value=freq_entry[0])[0]
    upper_frequency_index: int = find_nearest(array=freq_vec, value=freq_entry[-1])[0]
    fft_sum: float = np.sum(fft_vec[lower_frequency_index:upper_frequency_index])
    if with_normalization:
        fft_sum /= len(fft_vec)

    return fft_sum
```

When I call this function for different `bin_index` (0, ...) I should basically get the list of values you originally had.

If this works, we can go one step further and make the implementation efficient: instead of doing the FFT for every bin independently we do it for all bins once. This is how a "combiner" works (it is just for making the calculation more efficient). But first, lets make sure the implementation with a "simple" type works!

Hope this helps :) Sorry for all the delayed-ping-pong! 

Note: in your current implementation, you also return the "low" and "high" in the feature name. As described above, this is not possible. However we can think of adding yet another feature calculator which just returns the high and one which just returns the low. But this can be a separate PR

.gitignore
```diff
@@ -48,6 +48,7 @@ docs/text/_generated/
 # Virtual environment
 venv*
 activate
+tsfresh-env
```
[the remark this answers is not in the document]
anates: That's the name of my virtual environment (not venv), but I'll remove it.

.gitignore
```diff
@@ -61,3 +62,4 @@ tsfresh/notebooks/data/
 # dask
 dask-worker-space
 dask-worker-space/
+*.pyc.*
```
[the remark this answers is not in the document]
anates: Those are bytecode files, generated when I run the code for the first time. As I don't want to upload them in git, I ignored them.

anates:
Hei,
I addressed the other points, but have some comments regarding the main implementation (dvs. your last comment):

- The "low" and "high"-part of my return string intend to inform the user where the boundaries of that bin are. At least for my understanding those values are quite important, to understand my signal better. Are there other ways of returning those boundaries to the user, or is that on their side to identify them (outside the feature-extractor)?
- I agree with you that having a variable feature name might not be the best approach, but the number of bins will always be the same, independently of the data length, as it is based on a user-defined input. This corresponds to (for example) the coefficient-approach which always returns the first 100 coefficients, independently of the data itself. Would that still be ok?

Thanks!

nils-braun:
Hi!
To your first point:
Would it work to add them as a second feature calculator? One that returns the values and one that returns the high (and one the low)?

To your second point:
Yes indeed, that is the only option in the current way tsfresh is implemented. Except returning a statistics over the values as described above. Not saying this is the best way, but it is just the way tsfresh is currently working...

anates:
Hei,
thanks for the feedback! Based on that, would it be a solution to add an additional input parameter to this extraction function, to either return the bin value (default) for bin `n`, labelled with `_bin_n`, or return the high-/low-values for bin `n`? For typical operation (i.e. feature selection) it would not make a difference, but if the user is interested in the exact boundaries, they can still call this function separately with the same inputs as before, and obtain the bin limits. 

This would also reduce the amount of code duplication, else I would have to re-use the same code in two different extractors.

Concerning the second point: Assumed I understood your point correctly, the current implementation of returning `n` bins is already working as intended? For it being a combiner I then would just have to re-write handling of the input params, to allow it being a list of dicts.

Thanks!

nils-braun:
> Based on that, would it be a solution to add an additional input parameter to this extraction function
Good idea!
Very close to what you described is possible (and preferred as you said!). It would be ok to write a function which gets an additional parameter, but if it is a `simple` function it should only return a single number (the feature name will be set automatically). Something like

```python
def energy_content_frequency_brackets(signal, with_normalization, bin_index, return_mode) -> float:
   if return_mode == "value":
      return fft_sum # single float!
   elif return_mode == "high":
      return 42 # replace with actual number
   elif return_mode == "low":
      return 47 # replace with actual number
```

Feel free to use different names, this was just an example!

> Assumed I understood your point correctly, the current implementation of returning n bins is already working as intended?
Maybe I missed your point, but the current solution as it is implemented right now in the code does not work as expected for tsfresh (again: it is not wrong, it just does not work with the way tsfresh expects feature calculators to work). Let me rephrase what I tried to explain in https://github.com/blue-yonder/tsfresh/pull/1024#issuecomment-1591926985.
If you implement a `simple` calculator, which I would strongly encourage as the first implementation (as a `combiner` is harder to implement and only is for speeding up the implementation which can be the second step), your calculator is only allowed to return a single number. Not a list and not a feature name - but really just a single number.
In the example I gave in the thread above, I only return a single of the bins, while the `bin_index` is an additional parameter.
```
def energy_content_frequency_brackets(signal, with_normalization, bin_index) -> float:
```
(plus the return mode from the question above maybe)

Let me try to give another (maybe simpler) example: assuming you write a simple feature calculator which always returns the three largest number of a time series. You can _not_ write the feature calculator like this
```python
def my_calculator(data) -> List[float]:
   data = list(reversed(sorted(data)))
   return [data[0], data[1], data[2]]
```
and also not like this
```python
def my_calculator(data) -> List[Tuple[str, float]]:
   data = list(reversed(sorted(data)))
   return [("largest_1", data[0]), ...]
```
but you need to write it like this
```python
def my_calculator(data, which_largest) -> float:
   data = list(reversed(sorted(data)))
   return data[which_largest]
```

anates:
Work on this feature is on hold for the next weeks, but I'll take it up again after my break, if that's ok?
Thanks!

anates:
Hei,
I reverted the feature calculator to being a "simple" calculator as an initial proof of concept. I have not adjusted the testing-functions yet, but will do that as soon as the implementation gets closer to being final.
However, is the current implementation closer to being useful for tsfresh?
Thanks!

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -1231,6 +1233,101 @@ def get_kurtosis(y):
...
+
+    time_vec: np.ndarray = signal.index.to_numpy()
+
+    if not np.all(np.isclose(a=np.diff(time_vec), b=np.diff(time_vec)[0])):
+        warnings_message: str = "Uneven time steps detected. This feature can "
+        warnings_message += "only be used for an even distribution of time stamps."
+        warnings.warn(warnings_message, RuntimeWarning)
+        return np.NaN
+
+    number_of_bins: int = 10
```
nils-braun, line 1291: You also have this as an input parameter. Maybe better use that!

tsfresh/feature_extraction/feature_calculators.py
```diff
@@ -1231,6 +1233,101 @@ def get_kurtosis(y):
...
+
+    if bin_index >= number_of_bins:
+        bin_index = number_of_bins - 1
+
+    freq_vec: np.ndarray = np.fft.rfftfreq(n=len(signal), d=(time_vec[1] - time_vec[0]))
+    fft_vec: np.ndarray = np.abs(np.fft.rfft(signal.to_numpy()))
+
+    if (bin_ranges[0] < 0 and bin_ranges[1]) < 0 or (bin_ranges[0] >= bin_ranges[1]):
```
nils-braun, line 1299: I guess you want the bracket after the < 0?
```suggestion
    if (bin_ranges[0] < 0 and bin_ranges[1] < 0) or (bin_ranges[0] >= bin_ranges[1]):
```

nils-braun (commented):
Yes, great! The changes you did are very good - now it is in a shape to fit with the rest of the feature extractors.
Thanks for taking the time and sorry again for all the delays. I really appreciate your contribution.

tsfresh/feature_extraction/feature_calculators.py
anates: Hei,
yes, I'll implement it, but will work on it step-by-step, to ensure that every step still fits within the framework. Thanks!

PR #1043, opened 2023-09-13
added jit decorator to map_reduce method

ItsJustIkenna:


nils-braun:
Hi @ItsJustIkenna - thank you for the PR. Can you add a bit more context on the (speed) impact of this? Does it have any drawbacks? Does it work for any environments?

Issue #90, opened 2016-12-02
Allow NaN or None values to be passed in, and silently ignored


ClimbsRocks:
In my DataFrames, I oftentimes find it very reasonable to have quite a few NaN or None values. 

In order to run tsfresh, I just set all those values to 0, which is... not ideal. Even imputing missing values would not work particularly well for several of my use cases. 

Yet, it seems (from a super naive outsider's perspective), like this is filtering that tsfresh could do relatively easily itself. 

When it grabs each time_series, it can simply remove or categorically ignore NaN/None, and compute features on the values that do exist. This makes my life easier when, say, one customer signs up a month after another customer, and thus has missing values for that month. 

Again, super naive outsider perspective here, I know this might be impossible. But if it is possible, I'd love to add in that bit of filtering!

MaxBenChrist:
You can solve this by calling 'df.dropna(axis=1)' on the data frame that contains the time series in tsfresh format

MaxBenChrist:
This was a design decision. tsfresh will not tinker with the input time series data by for example imputing values or dropping NAs.

Reason behind this: In data science projects, NAs should be handled with special care, often they contain a lot of information. We don't want our packages to silently remove those informations by dropping it. This is way we are not imputing the input data.

tompollard:
> This was a design decision. tsfresh will not tinker with the input time series data by for example imputing values or dropping NAs.

In many cases doesn't it make sense to treat nans as nans, rather than requiring imputation? For example, features like mean, max and min are more informative if the nans are simply ignored when computing the value.

yitao-yu:
If we are not working with a lot of features, we can do this by [implementing our own feature functions](https://tsfresh.readthedocs.io/en/latest/text/how_to_add_custom_feature.html). 

While doing so, I found that one of the official features might actually ignore nan, which is the skewness. It utilizes `pandas.skew` which has a parameter `skipna` with a default `True`. This can contradict the above design principle. 

Will try it on dummy data and update if that's true.

yitao-yu:
Actually, there is a workaround by [customizing features](https://tsfresh.readthedocs.io/en/latest/text/how_to_add_custom_feature.html). 

There is, however, something I discovered that is not consistent with the design principle introduced in this post when doing so. `tsfresh.feature_extraction.feature_calculators.skewness` utilizes `pandas.Series.skew` with a parameter `skipna` default to `True`.

This is confirmed with dummy data: 

```
[...]

PR #1051, opened 2023-10-06
Update tsfresh.feature_extraction.feature_calculators.skewness to make it consistent with the design principle of not ignoring nan

yitao-yu:
The current implementation of `tsfresh.feature_extraction.feature_calculators.skewness` utilizes a function from pandas that ignores `nan` by default, which is not consistent with the performance of other functions and the design principle. 

This update provides a fix by adding a `skipna=False` argument. But since it's discovered simply by chance, I do suggest a review of other parts of the code. 

[More](https://github.com/blue-yonder/tsfresh/issues/90)

### DIFF
--- a/tsfresh/feature_extraction/feature_calculators.py
+++ b/tsfresh/feature_extraction/feature_calculators.py
@@ -757,8 +757,7 @@ def skewness(x):
     """
     if not isinstance(x, pd.Series):
         x = pd.Series(x)
-    return pd.Series.skew(x)
-
+    return pd.Series.skew(x, skipna = False)
 
 @set_property("fctype", "simple")
 @set_property("input", "pd.Series")


nils-braun (approved):
Thanks @yitao-yu - you are completely correct!
It would be nice to also check all other calculators. Do you have an idea on how to do this automatically?

yitao-yu:
> Thanks @yitao-yu - you are completely correct! It would be nice to also check all other calculators. Do you have an idea on how to do this automatically?

So after much closer looks, I do think that the package would throw an error(there is a  `_check_nan` function called in the `WideTsFrameAdapter` class) if you are utilizing `tsfresh.feature_extraction.feature_calculators`. 

For individual features, maybe some sort of testing program can be written to check if they have similar problems. Or I can just write programs to look for features that are calling functions implemented under `pandas.Series` and also has an optional argument `skipna`.

nils-braun:
Sorry for the long pause @yitao-yu - and thanks for looking into it. I think for now, including your change for this function is a good start. Could you fix the code style issue? The  we can merge!

PR #1061, opened 2024-01-26
Make Dask and Distributed optional dependencies

lifepillar:
Addresses https://github.com/blue-yonder/tsfresh/discussions/1060.

Dask was de facto optional already (see the guarded import in `tsfresh/feature_extraction/data.py`), but before this PR it was installed by default.

Distributed is imported only to support Dask, when the classes `ClusterDaskDistribution` or `LocalDaskDistribution` are used, so it is not needed if Dask is not imported.

To install tsfresh with those dependencies, one can do:

    pip install -e ".[dask]"

~This is needed, for example, to run the test suite.~ The documentation has been updated accordingly.

### DIFF
--- a/docs/text/quick_start.rst
+++ b/docs/text/quick_start.rst
@@ -13,6 +13,15 @@ As the compiled tsfresh package is hosted on the Python Package Index (PyPI) you
 
    pip install tsfresh
 
+If you need to work with large time series data that may not fit in memory, install tsfresh with
+`Dask <https://www.dask.org>`_:
+
+.. code:: shell
+
+   pip install tsfresh[dask]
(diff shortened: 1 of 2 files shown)


nils-braun (commented):
Thanks for the PR! Could you make sure the CI pipeline is using the new dask extra dependency as well?
You can also (as an alternative) just reference the dask extra dependency from the testing extra dependency.

nils-braun (approved):
Thanks @lifepillar! I just changed the "tsfresh[dask]" to  "%(dask)s" in the `setup.cfg`, because it is more general - but other than that LGTM!

Issue #1077, opened 2024-06-07
id column in the result of make_forecasting_frame have only (id, ?) as identifier


heib6xinyu:
**The problem**: make_forecasting_frame function provide frame and y with id column in the format of (id,1), (id,2) instead of like the documentation described (For identifying every subsequence, tsfresh uses the time stamp of the point that will be predicted together with the old identifier as “id”.)

```
# Step 1: Create Dummy Data
np.random.seed(42)  # For reproducibility

# Create a date range
date_range = pd.date_range(start='2023-01-01', periods=100, freq='D')

# Create dummy product IDs
product_ids = ['P001', 'P002', 'P003']

# Generate dummy data
data = []
for product_id in product_ids:
    for date in date_range:
        data.append([product_id, date, np.random.uniform(10, 100)])

# Create a DataFrame
df = pd.DataFrame(data, columns=['id', 'timestamp', 'price'])

# Create the forecasting frame
df_forecasting, y = make_forecasting_frame(df['price'], kind="price", max_timeshift=5, rolling_direction=1)

# Display the first few rows of the forecasting frame
print("\nForecasting Frame:")
print(df_forecasting.head())
print("\nTarget Variable (y):")
print(y.head())
```

You should be able to recreate my problem using the above code

```
Rolling: 100%|██████████| 300/300 [00:02<00:00, 132.49it/s]

Forecasting Frame:
        id  time      value   kind
1  (id, 1)     0  43.708611  price
3  (id, 2)     0  43.708611  price
4  (id, 2)     1  95.564288  price
6  (id, 3)     0  43.708611  price
7  (id, 3)     1  95.564288  price

Target Variable (y):
(id, 1)    95.564288
(id, 2)    75.879455
(id, 3)    63.879264
(id, 4)    24.041678
(id, 5)    24.039507
Name: value, dtype: float64
```
As you can see, there are only (id, time), I have no way to see which product the data actually belongs to.

**Anything else we need to know?**: I would hope the result looks like (p001,1),...,(p003, 4), but I don't know how to.

**Environment**:

- Python version: python 3.10
- Operating System: windows 10
- tsfresh version:0.20.1
- Install method (conda, pip, source):pip install tsfresh

```

heib6xinyu:
I look through the source code, there is some hard coding part in it. I copy the function to my local computer and modify it to receive a id field, to replace the 'id' hard coding part, it works for me now.

nils-braun:
[...]

PR #1080, opened 2024-06-14
Added "id" parameter for make_forecaseting_frame so that it will not …

heib6xinyu:
…have hard coded 'id' in the result.

This is my edition of make_forecasting_frame of the issue #1077 .
I added id parameter to the function, so the result will be (a, t1) instead of (id, t1) for item that has the id of a.

### DIFF
--- a/tsfresh/utilities/dataframe_functions.py
+++ b/tsfresh/utilities/dataframe_functions.py
@@ -579,7 +579,7 @@ def roll_time_series(
     return df_shift.sort_values(by=["id", column_sort or "sort"])
 
 
-def make_forecasting_frame(x, kind, max_timeshift, rolling_direction):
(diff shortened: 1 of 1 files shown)


nils-braun:
Hi @heib6xinyu - thanks for providing a PR!
As we discussed on the linked issue, the function `make_forecasting_frame` is only meant to be used for singular time series (as the name "frame" suggests). While having an `id` parameter is technically possible (as you have shown in your PR), it might make people think the function should be used for timeseries with multiple ids (because all other functions which allow for different `id` columns have a parameter to control the `id` column). 
Therefore I would not like to add the `id` parameter. Also, it would be a breaking API change of our public API, so we need to be very careful about it.

Would it be possible for you to do the change from `id` to e.g. `a` as a post-processing step in your code?

Issue #1073, opened 2024-05-19
Feature selection taking magnitutes longer than it should


Sarius2009:
When extracting data from the same dataset and selecting from the extracted features, one set of parameters to extract data, which results in 154 classes, 10400 TSs, longest TS of 10000 datapoints and 1.2 GB of json data works fine with EfficientFCParameters, but another one, which results in <90 classes, 550 sessions and 8MB of data with takes 40seconds to extract the features, as expected, but 3 hours to select the features.. This was the smallest sample I could create which exibited this beahviour, larger samples (~200MB) will take >60GB of RAM an crash the program. Using `n_jobs=0` did not help.

Profiling my example below, almost all of the time seems to be taken up by `_recv_bytes` and `_get_more_data`.

Running W11 on a system with 16GB tsfresh 0.20.2 installed via pip, tested code in a jupyter notebook and normal python 3.12.

Minimal example with provided data:
[problem_features_eff.csv](https://github.com/blue-yonder/tsfresh/files/15396700/problem_features_eff.csv)
[problem_id_to_userid_eff.csv](https://github.com/blue-yonder/tsfresh/files/15447777/problem_id_to_userid_eff.csv)
```
from tsfresh import select_features
import pandas as pd
if __name__ == '__main__':
    X = pd.read_csv('problem_features_eff.csv')
    id_to_userID = pd.read_csv('problem_id_to_userID_eff.csv').T.iloc[1]
    print('Starting')
    select_features(X, id_to_userID, ml_task='classification')
```
Edit: Changed Parameters to efficient, attached problematic data
Edit: Found out it wasn't inherently a memory issue, adjusted title and text accordingly

nils-braun:
Hi @Sarius2009!
Your feature selection is taking so long, because your `id_to_userID` (the series you use as `y` in the `select_features` method) contains more than two distinct values and you selected "classification" as your ml task. This means, tsfresh will perform a 1-vs-all feature selection for all your distinct values and as you have 47 distinct values, this will take quite some time... 

Or maybe I misunderstood: Are you referring to the number of distinct values in `id_to_userID` as your classes?

Sarius2009:
Hi @nils-braun,
Yes, the values in the `id_to_userID` are what I refered to as classes, I seem to have attached the wrong file. It is updated now and should hopefully make sense, eg. the features for id 0 belong to user 324, as do the ones for id 2, etc...
[...]

PR #1081, opened 2024-06-26
Fixes issue #1073: Updated setup.cfg to require newer scipy version

Sarius2009:


nils-braun (approved):
Thanks!

nils-braun:
@Sarius2009 - it seems there is no such version for older python versions

Sarius2009:
> @Sarius2009 - it seems there is no such version for older python versions

Yes, scipy 1.14 requires at least python 3.10

nils-braun:
Can you adjust the PR so that your lower limit only applies in this case? tsfresh is also compatible with older versions of python

Sarius2009:
I think I did, hopefully correctly

nils-braun (approved):
👍

PR #1083, opened 2024-07-09
extract_revant_features now passes chunksize to extract_features

Sarius2009:
`extract_relevant_features` did previously not pass `chunksize` to `extract_features`, but only to `select_features`


nils-braun (approved):
thanks!

PR #1093, opened 2024-11-14
Dockerise testing for dev

Scott-Simmons:
Related: https://github.com/blue-yonder/tsfresh/pull/994. 

Easiest to go commit by commit. Uses 3 "flows"

(1) **Test locally on specific python version** Is useful for quick iteration.
(2) **Test locally on the discoverable python versions** Is a useful sanity check for ensuring new development works on multiple versions of python. 
(3) **Test in dedicated testing environment with guaranteed availability of python versions** Is a good final sanity check. Takes longer, but creates a reproducible environment to test off and is a good 'source of truth'.

### 1 Test locally on specific python version

Testing time for one version is ~1 min. Note this includes the speedup from cached packages

### 2 Test locally on the discoverable python versions

image

Testing time across versions 3.7-3.11 is ~4 mins, but note that this speedup includes the speedup from cached packages.

Also note that python 3.7 fails on my machine due to some idiosyncratic issue related to my setup `py37/lib/python3.7/site-packages/pip/_vendor/typing_extensions.py` ... getting same error as [here](https://github.com/microsoft/vscode/issues/200543) but not using vscode. Not an issue for me personally but shows an example of why testing within an isolated testing environment could be a good way to ensure reproducible version targeting. 

### 3 Test in dedicated testing environment with guaranteed availability of python versions 

image

Build time for testenv ~5 mins but this is a once-off build for the testenv.

Testing time across versions 3.7-->3.11 is ~6 mins.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,53 @@
+WORKDIR := /tsfresh
+TEST_IMAGE := tsfresh-test-image
+TEST_DOCKERFILE := Dockerfile.testing
+TEST_CONTAINER := tsfresh-test-container
+# >= 3.9.2 ---> https://github.com/dask/distributed/issues/7956
+PYTHON_VERSIONS := "3.7.12 3.8.12 3.9.12 3.10.12 3.11.0"
+BLACK_VERSION := 22.12.0
+# Isort 5.12.0 not supported with python 3.7.12
+ISORT_VERSION := 5.12.0
+
+build-testenv:
+	docker build \
+			-f $(TEST_DOCKERFILE) \
+			-t $(TEST_IMAGE) \
+			--build-arg PYTHON_VERSIONS=$(PYTHON_VERSIONS) \
+			.
+
+# Tests `PYTHON_VERSIONS`, provided they are also
+# specified in setup.cfg `envlist`
+test-all-testenv: clean build-testenv
+	docker run --rm \
+			--name $(TEST_CONTAINER) \
+			-v .:$(WORKDIR) \
+			-v build_artifacts:$(WORKDIR)/build \
+			-v tox_artifacts:$(WORKDIR)/.tox \
+			-v egg_artifacts:$(WORKDIR)/tsfresh.egg-info \
+			$(TEST_IMAGE)
+
+# Tests the python binaries installed
+# on local machine, provided they are also
+# specified in setup.cfg `envlist`
+test-all-local: clean
+	tox -r -p auto
+
+# Tests for python version on local machine in
+# current context (e.g. global or local version of
+# python set by pyenv, or the python version in
+# the active virtualenv).
+test-local: clean
+	pip install .[testing]
+	pytest
+
+clean:
+	rm -rf .tox build/ dist/ *.egg-info
+
+install-formatters:
+	pip install black==$(BLACK_VERSION) isort==$(ISORT_VERSION)
+
+format: install-formatters
+	black --extend-exclude \.docs .
+	isort --profile black --skip-glob="docs" .
+
+.PHONY: clean test-all-local test-local test-all-testenv format install-formatters
--- a/Makefile
+++ b/Makefile
@@ -2,22 +2,20 @@ WORKDIR := /tsfresh
 TEST_IMAGE := tsfresh-test-image
 TEST_DOCKERFILE := Dockerfile.testing
 TEST_CONTAINER := tsfresh-test-container
-# >= 3.9.2 ---> https://github.com/dask/distributed/issues/7956
-PYTHON_VERSIONS := "3.7.12 3.8.12 3.9.12 3.10.12 3.11.0"
-BLACK_VERSION := 22.12.0
-# Isort 5.12.0 not supported with python 3.7.12
-ISORT_VERSION := 5.12.0
+PYTHON_VERSIONS := "3.7 3.8 3.9 3.10 3.11"
 
-build-testenv:
+# Tests `PYTHON_VERSIONS`, provided they are also
+# specified in setup.cfg `envlist`
+test-all-testenv: build-docker-testenv run-docker-tests clean
+
+build-docker-testenv:
 	docker build \
 			-f $(TEST_DOCKERFILE) \
 			-t $(TEST_IMAGE) \
 			--build-arg PYTHON_VERSIONS=$(PYTHON_VERSIONS) \
 			.
 
-# Tests `PYTHON_VERSIONS`, provided they are also
-# specified in setup.cfg `envlist`
-test-all-testenv: clean build-testenv
+run-docker-tests:
 	docker run --rm \
 			--name $(TEST_CONTAINER) \
 			-v .:$(WORKDIR) \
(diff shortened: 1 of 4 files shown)


Scott-Simmons:
Dirty commits while in draft will be tidied up before putting into review

Scott-Simmons:
Easiest to review commit-by-commit

Makefile
```diff
@@ -0,0 +1,53 @@
...
+			.
+
+# Tests `PYTHON_VERSIONS`, provided they are also
+# specified in setup.cfg `envlist`
+test-all-testenv: clean build-testenv
+	docker run --rm \
+			--name $(TEST_CONTAINER) \
+			-v .:$(WORKDIR) \
+			-v build_artifacts:$(WORKDIR)/build \
+			-v tox_artifacts:$(WORKDIR)/.tox \
```
Scott-Simmons, line 25: FYI build_artifacts and tox_artifacts are a 'hack'. If no docker volume is specified then they end up in the context of the .:$(WORKDIR) mapping. The issue there is that when .tox and build directories are created in the docker container, they are created by the root user but end up in the source code. This means that have root permissions not user permissions. 


Explicitly using docker volumes build_artifacts and tox_artifacts gets around this as they are removed after the container finishes running.

Scott-Simmons:
@nils-braun let me know If this is worth adding, hoping that it can helpful for the release process going forward

Makefile
```diff
@@ -0,0 +1,53 @@
+WORKDIR := /tsfresh
+TEST_IMAGE := tsfresh-test-image
+TEST_DOCKERFILE := Dockerfile.testing
+TEST_CONTAINER := tsfresh-test-container
+# >= 3.9.2 ---> https://github.com/dask/distributed/issues/7956
+PYTHON_VERSIONS := "3.7.12 3.8.12 3.9.12 3.10.12 3.11.0"
```
nils-braun, line 6: Do we really need to specify also the patch version number? I do not want to create the need to update this file whenever there is a new python version coming out.

Makefile
```diff
@@ -0,0 +1,53 @@
...
+build-testenv:
+	docker build \
+			-f $(TEST_DOCKERFILE) \
+			-t $(TEST_IMAGE) \
+			--build-arg PYTHON_VERSIONS=$(PYTHON_VERSIONS) \
+			.
+
+# Tests `PYTHON_VERSIONS`, provided they are also
+# specified in setup.cfg `envlist`
+test-all-testenv: clean build-testenv
```
nils-braun, line 20: Why do you need to call `clean`?

Makefile
```diff
@@ -0,0 +1,53 @@
...
+	docker run --rm \
+			--name $(TEST_CONTAINER) \
+			-v .:$(WORKDIR) \
```
nils-braun, line 23: Does I maybe make sense to mount the folder read-only?

Makefile
```diff
@@ -0,0 +1,53 @@
...
+# on local machine, provided they are also
+# specified in setup.cfg `envlist`
+test-all-local: clean
+	tox -r -p auto
+
+# Tests for python version on local machine in
+# current context (e.g. global or local version of
+# python set by pyenv, or the python version in
+# the active virtualenv).
+test-local: clean
```
nils-braun, line 39: Same question: why do you need to clean?

docs/text/how_to_contribute.rst
```diff
@@ -64,21 +64,32 @@ you have to:
 
 .. code::
 
-    pytest
+    make test-local
```
nils-braun, line 67: With this, users will need to reinstall the pip packages for testing on very run of the test. I know, it should be quick as all packages are already there, but it does take some extra time nevertheless. 

I would suggest to not add a makefile command for the local testing but to state that the users need to run pytest. The makefile command just replaced two commands and the installation command is explained on the page already anyways

Makefile
```diff
@@ -0,0 +1,53 @@
...
+	pip install .[testing]
+	pytest
+
+clean:
+	rm -rf .tox build/ dist/ *.egg-info
+
+install-formatters:
```
nils-braun, line 46: Why is this command needed and the one below? We have pre-commit for exactly this, or?

docs/text/how_to_contribute.rst
```diff
@@ -64,21 +64,32 @@ you have to:
...
 
 
 .. code::
 
-    tox -r
+    make test-all-local
 
 
-`tox -r` will execute tests for the Python versions specified in `setup.cfg <https://github.com/blue-yonder/tsfresh/blob/1297c8ca5bd6f8f23b4de50e3f052fb4ec1307f8/setup.cfg>`_ using the `envlist` variable. For example, if `envlist` is set to `py37, py38`, the test suite will run for Python 3.7 and 3.8 on the local development platform, assuming the binaries for those versions are available locally. The exact Python microversions (e.g. `3.7.1` vs `3.7.2`) depend on what is installed on the local development machine.
+This will execute tests for the Python versions specified in `setup.cfg <https://github.com/blue-yonder/tsfresh/blob/1297c8ca5bd6f8f23b4de50e3f052fb4ec1307f8/setup.cfg>`_ using the `envlist` variable. For example, if `envlist` is set to `py37, py38`, the test suite will run for Python 3.7 and 3.8 on the local development platform, assuming the binaries for those versions are available locally. The exact Python microversions (e.g. `3.7.1` vs `3.7.2`) depend on what is installed on the local development machine.
```
nils-braun, line 78: Sorry, did not spot this in the last PR: can you make sure the link points to the latest HEAD/main version of the file?

docs/text/how_to_contribute.rst
```diff
@@ -64,21 +64,32 @@ you have to:
...
+    make test-local
 
 
 To test changes across multiple versions of Python, run:
 
 
 .. code::
 
-    tox -r
+    make test-all-local
```
nils-braun, line 76: Similar to above: I am not a big fan of replacing Single-Line commands with a makefile command. This adds another layer of indirection "advanced" users can be annoyed about (they know what tox is, so they want to potentially change options etc) and "new" users copy-paste what is in here anyways, so it does not matter.

nils-braun (commented):
Thanks @Scott-Simmons - similarly to your last PR this is of very high quality. I do have a few suggestions around the single-line makefile commands but I am in general happy with this. I do not think all users will require it, but I do think that it can be a very good UX improvement for some users. Well done!

Makefile
Scott-Simmons: Does not look like read only plays nicely

```bash
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "/var/lib/docker/volumes/egg_artifacts/_data" to rootfs at "/tsfresh/tsfresh.egg-info": mkdir /var/lib/docker/overlay2/f1e44a21b8f2b9407dbce127e0383601fcc372b2dda05cae82e9c2a6de33df42/merged/tsfresh/tsfresh.egg-info: read-only file system: unknown.
```

Makefile
Scott-Simmons: Moved to cleaning within the docker testing target 3ddad5985542de57c11229ccc6a1f47a1e49ec04

Scott-Simmons (commented):
Hey @nils-braun, have responded to the feedback. Let me know if there is anything else

docs/text/how_to_contribute.rst
nils-braun: I do not know for tox, but many tools allow to set default options on the pyproject or setup.cfg or their own config file. This is typically better, because then the tool can do its own "schema" migration. For example, we so this for pytest in the setup.cfg.

nils-braun (approved):
Looks good!

PR #1097, opened 2024-12-31
use PyWavelet for CWT

igor-pechersky:
https://github.com/blue-yonder/tsfresh/discussions/1096

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -37,11 +37,11 @@ install_requires =
     requests>=2.9.1
     numpy >=1.15.1
     pandas>=0.25.0
-    scipy>=1.14.0,<1.15;python_version>='3.10'
-    scipy>=1.2.0,<1.15;python_version<'3.10'
+    scipy>=1.14.0;python_version>='3.10'
+    scipy>=1.2.0;python_version<'3.10'
     statsmodels>=0.13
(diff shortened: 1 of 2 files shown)


nils-braun (approved):
Thank you very much for taking the time to do the fix! I am ok with the changes. Did you have a chance to check the resulting features, are they the same?

Could you also fix the styling issue in the code? Thanks!

nils-braun:
I have seen in the CI/CD that the pywavelets package in the version you were using is not available for python versions 3.9 and older. I can (and should!) drop the python versions 3.7 and 3.8, but 3.9 has not reached end of life yet, so I would like to keep it.
I see two options here:
- the code implements two branches for versions that support this pywavelets and one that does not
- we limit the scipy version (not favoured, but if there is no other way...)

Is this particular version of pywavelets needed?

nils-braun:
@igor-pechersky - are the questions I noted above something you want to take a look at? Is this something you would have time to check in the near future? No worries if not, but I would probably take over the PR in this case to make sure we can release a fix. Thanks again!

nils-braun:
Taking over the PR. The cwt function is also present in older versions of pywavelets, so the high constraint is not needed.

PR #1102, opened 2025-01-09
Add bisect debug helper

Scott-Simmons:
Add Git Bisect Helper to Automate Regression Debugging.

During the process of feature development not part of this PR, I used git bisect` to help identify the commit that introduced a bug (the bug in question was caused by a variable name clashing with a function name).

Adding this git bisect helper to the main line as it has a great impact on automated bug finding (when the commits are atomic).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,11 @@ clean:
 	rm -rf .tox build/ dist/ *.egg-info
 
 bisect:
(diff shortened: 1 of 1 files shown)


nils-braun (approved):
Thanks for the addition! If it helped you, I am ok with the additional entry in the makefile 👍

PR #1120, opened 2026-04-05
Fix #1118: Update stumpy to >=1.11.1 to fix Python 3.11+ uv resolutio…

### LINKED ISSUE #1118: Doesn't work on python 3.11+ in contradiction of what is specified
Error from uv:

> RuntimeError: Cannot install on Python version 3.13.2; only versions >=3.6,<3.10 are supported

> help: `llvmlite` (v0.36.0) was included because `your-repo` (v0.1.0) depends on `tsfresh` (v0.21.1) which depends on `stumpy` (v1.10.2) which depends on `numba` (v0.53.1) which depends on `llvmlite`

divyansh0-ai:
This PR resolves the dependency resolution issues on Python 3.11+ (Issue #1118) and fixes a few initial Pandas 3.0 regression test failures.                                                                                                                                                                                                       
Changes Made:
1.setup.cfg: Bumped stumpy minimum version from >=1.7.2 to >=1.11.1 to fix uv dependency resolution failures on Python 3.11+ (resolves #1118).

2.setup.py & pyproject.toml: Added a standard pyproject.toml exposing pyscaffold and setuptools<70 build requirements, and removed a legacy pkg_resources check in setup.py that caused isolated build managers (like uv or modern pip) to crash with ModuleNotFoundError.

3.tsfresh/utilities/dataframe_functions.py: Fixed a widespread KeyError: 'id' crash in the test suite caused by Pandas 3.0 groupby().apply() no longer including grouping keys in its evaluated chunks. Added fallback logic to dynamically restore missing grouping keys from the DataFrame's name tuple.

4.tests/units/utilities/test_dataframe_functions.py: Updated deprecated Pandas dataframe frequency string aliases (changed freq="H" to freq="h") to fix immediate Pandas 3.0 deprecation failures.

<img width="1298" height="446" alt="image" src="https://github.com/user-attachments/assets/730e8451-f1df-4f93-80aa-971742fe35ef" />

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,3 @@
+[build-system]
+requires = ["setuptools<70", "wheel", "pyscaffold>=3.3a0,<4"]
+build-backend = "setuptools.build_meta"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,6 @@
 [build-system]
+# setuptools is capped below 70 because PyScaffold <4 (used by this project)
+# has known compatibility issues with setuptools>=70.  Once the project
+# migrates to PyScaffold 4+, this upper bound can be removed.
 requires = ["setuptools<70", "wheel", "pyscaffold>=3.3a0,<4"]
 build-backend = "setuptools.build_meta"
--- a/tests/integrations/examples/test_robot_execution_failures.py
+++ b/tests/integrations/examples/test_robot_execution_failures.py
@@ -54,4 +54,4 @@ class RobotExecutionFailuresTestCase(TestCase):
         )
 
         assert len(y.unique()) > 2
-        assert y.dtype == object
+        assert y.dtype == object or isinstance(y.dtype, pd.StringDtype)
--- a/tests/integrations/examples/test_robot_execution_failures.py
+++ b/tests/integrations/examples/test_robot_execution_failures.py
@@ -8,6 +8,7 @@ import tempfile
 from unittest import TestCase
 
 import numpy as np
+import pandas as pd
 from pandas import DataFrame, Series
 
 from tsfresh import extract_features
--- a/tsfresh/utilities/dataframe_functions.py
+++ b/tsfresh/utilities/dataframe_functions.py
@@ -271,6 +271,24 @@ def get_ids(df_or_dict, column_id):
         raise TypeError("df_or_dict should be of type dict or pandas.DataFrame")
 
 
+def _restore_grouping_keys(df_chunk, grouper):
+    """
+    Restore grouping-key columns that Pandas 3.0+ no longer includes in
+    ``groupby().apply()`` chunks.  When there is more than one grouping key,
+    ``df_chunk.name`` is a tuple whose i-th element corresponds to
+    ``grouper[i]``; for a single key it is a scalar.
+
+    :param df_chunk: the DataFrame chunk passed by ``groupby().apply()``
+    :type df_chunk: pandas.DataFrame
+    :param grouper: ordered list of column names used as grouping keys
+    :type grouper: list
+    """
+    multi_key = len(grouper) > 1
+    for i, col in enumerate(grouper):
+        if col not in df_chunk.columns:
+            df_chunk[col] = df_chunk.name[i] if multi_key else df_chunk.name
+
+
 def _roll_out_time_series(
     timeshift,
     grouped_data,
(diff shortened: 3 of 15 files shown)


tsfresh/utilities/dataframe_functions.py
```diff
@@ -334,6 +335,14 @@ def _f(x):
...
 
         df_temp = df_temp.copy()
+        
+        if grouper is not None:
+            for col in grouper:
+                if col not in df_temp.columns:
+                    if len(grouper) > 1:
+                        df_temp[col] = x.name[grouper.index(col)]
+                    else:
+                        df_temp[col] = x.name
```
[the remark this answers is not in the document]
nils-braun: The O(n2) is not so important here, but I think the code is also cleaner. Having a dedicated helper function also makes sense :+1:

pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[build-system]
```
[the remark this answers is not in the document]
nils-braun: I think mentioning that this is a requirement from pyscaffold is indeed a good idea

nils-braun (commented):
Thank you very much @divyansh0-ai for your time. Copilots suggestions were already quite reasonable. Can you implement those as well?

divyansh0-ai:
@nils-braun Thanks for the review and the feedback! I've gone ahead and implemented the suggestions, along with the remaining Pandas 3.0 compatibility fixes. The PR is ready for another look. Could you please approve the workflow runs when you have a moment.

nils-braun (approved):
Well done! Thanks for fixing all this!
I think there are just a few linting issues left.

Update: apart from the missing pd import there are also other two test failures. Do you want to look into them as well @divyansh0-ai or shall I take a look? One looks like a grouping key issue.

tests/integrations/examples/test_robot_execution_failures.py
```diff
@@ -54,4 +54,4 @@ def test_multilabel_target_on_request(self):
         )
 
         assert len(y.unique()) > 2
-        assert y.dtype == object
+        assert y.dtype == object or isinstance(y.dtype, pd.StringDtype)
```
nils-braun, line 58: you probably want to import pd here :)

divyansh0-ai: @nils-braun Thanks for the approval! I will gladly take care of these last few items. I’ll fix the missing pd import, clear up the linting, and dig into those two 3.11/3.12 test failures today. I'll push an update shortly!

tsfresh/utilities/dataframe_functions.py
```diff
@@ -732,13 +759,19 @@ def _add_id_column(df_chunk):
...
+        except TypeError:
+            df = df.groupby(grouper, group_keys=False).apply(_add_id_column)
     else:
         df = _add_id_column(df)
 
     if column_sort:
         df = df.sort_values(column_sort)
 
-    df = df.set_index(df.index.get_level_values(-1))
+    df = df.reset_index(drop=True)
```
nils-braun: Sorry for the late comment @divyansh0-ai - but this change I do not understand. Doesn't this remove some information compared to before? Maybe I am overlooking something

nils-braun (commented):
Sorry I haven't commented in this already @divyansh0-ai - I forgot about this PR!

tsfresh/utilities/dataframe_functions.py
divyansh0-ai: @nils-braun No worries at all about the timing! Thank you for reviewing it so carefully.

You are completely right—it was dropping information. When trying to fix the new Pandas 3.0 `groupby().apply()` MultiIndex errors in the test suite, I used `drop=True` to quickly flatten the dataframe. But as you pointed out, doing that accidentally threw away the original row indexes (which ruins the time-series timestamps!). 

I actually realized my mistake while reviewing the tests and pushed a fix for it yesterday in commit `1d58e96`. It now safely handles the index without dropping the original information. Thank you again for spotting that, and please let me know if the updated version looks good to you!

nils-braun (approved):
Looks good now!

nils-braun:
@divyansh0-ai - can you fix the remaining style check issues? After that, the PR is ready to go in!

divyansh0-ai:
@nils-braun sir Thank you for merging this! Sorry for the delay in responding. I was tied up with my end-semester exams, which finished on May 29th. I appreciate you handling the final style checks to get this merged. Now that I am free, I am looking forward to contributing more to the project please let me know if there are any other issues or areas where I can help.

PR #1126, opened 2026-07-12
Fix #966: accept numpy.ndarray target in calculate_relevance_table

### LINKED ISSUE #966: calculate_relevance_table() does not work with y in form of numpy.ndarray
**The problem**:
The function `feature_selection.relevance.calculate_relevance_table()` is supposed to support numpy.ndarray as y argument. This causes an attribute error though because numpy.ndarray does not have the `sort_index` function that is called in line 164 of https://github.com/blue-yonder/tsfresh/blob/main/tsfresh/feature_selection/relevance.py

**Anything else we need to know?**:

**Environment**:

I think this issue does not depend on the environment.

- Python version:
- Operating System:
- tsfresh version:
- Install method (conda, pip, source):

Leonard013:
Fixes #966.

## Why

`calculate_relevance_table(X, y)` documents its target as

```
:type y: pandas.Series or numpy.ndarray
```

so a `numpy.ndarray` is a **supported, documented input**. However, passing an
ndarray crashes:

```
AttributeError: 'numpy.ndarray' object has no attribute 'sort_index'
```

at `tsfresh/feature_selection/relevance.py` (`y = y.sort_index()`). Two users hit
this in #966 (reporter + @mariokrous "Facing the same error").

The sibling public function `select_features` already accepts an ndarray target and
converts it before delegating to `calculate_relevance_table`
(`tsfresh/feature_selection/selection.py`):

```python
if isinstance(y, np.ndarray):
```

But `calculate_relevance_table` is itself public (exported from
`tsfresh.feature_selection`) and is called directly, so it needs the same handling.

## What

Near the top of `calculate_relevance_table`, before the index-based operations,
convert an ndarray target to a `pandas.Series` aligned to `X`'s rows:

```python
if isinstance(y, np.ndarray):
    assert len(X) == len(y), "X and y must contain the same number of samples."
```

This mirrors `select_features` exactly (same conversion, same length-mismatch
message).

## Index-alignment reasoning (the important part)

A `numpy.ndarray` has no index; the only sensible interpretation is **positional**:
`y[i]` is the target for the `i`-th row of `X`. Assigning `index=X.index` makes
`y[i]` carry the label `X.index[i]`, so the `(row, target)` pairing equals that
positional interpretation.

The function then runs `y = y.sort_index()` and `X = X.sort_index()`. Because the
Series was built with `X.index`, both sort by the **same** labels, so the per-row
`(feature, target)` pairing is preserved and the downstream
`assert list(y.index) == list(X.index)` passes. Passing `y` as an ndarray is thus
**exactly equivalent** to passing `pd.Series(y, index=X.index)` — identical code
path from there on, so the relevance tables are identical (same feature order,
`p_value`, `relevant` flags), not merely both non-crashing.

Verified that correct alignment is load-bearing: with a shuffled, non-default index,
the correctly-aligned separating feature gets `p ≈ 6.8e-8` (relevant) vs a
deliberately mis-aligned target `p ≈ 0.07` (not relevant).

## Tests

Added `TestCalculateRelevanceTable::test_ndarray_target_matches_series_target`
(parametrized over a classification target `[0, 1, 1]` and a regression target
`[0.1, 0.2, 0.1]`, exercising both ml-task code paths). It (a) does not raise when
`y` is a `numpy.ndarray`; and (b) asserts the returned relevance table **equals** the
one from the equivalent `pd.Series(y, index=X.index)` via
`pd.testing.assert_frame_equal`.

The test uses a **non-default, unsorted** `X` index (`[10, 5, 3]`) so it also guards
the alignment: a naive `pd.Series(y)` (fresh `RangeIndex`) fix would trip the
`assert list(y.index) == list(X.index)` check and fail the test.

Before the fix both parametrized cases fail with the `AttributeError`; after, they
pass. `pytest tests/units/feature_selection/test_relevance.py` → `21 passed`; full
`tests/units/feature_selection/` → `74 passed`. `black`/`isort` clean.

## Notes for reviewers

- Minimal change (8 lines in `relevance.py`), mirroring the established
  `select_features` precedent; no behaviour change for existing `pandas.Series`
  inputs.
- `CHANGES.rst` intentionally not touched (maintainer-managed in the "Prepare X.Y.Z"
  release commits; recent feature/bugfix PRs don't edit it). Suggested next-release
  entry: `- Fix #966: calculate_relevance_table now accepts a numpy.ndarray target`.


nils-braun (approved):
Looks good! Thanks for the fix

